Issue #271, opened 2014-01-31
Argument name harmonization base-command, base-console


Bladrak:
I noticed there are two ways to specify a console/command path to the `update-core-routes` & `create-snapshots commands` (maybe other commands as well, didn't look for it). It can be either `--base-command` or `--base-console` for the argument. It would be nice to harmonize that.

core23:
Can you provide a PR for that? @Bladrak

PR #273, opened 2014-02-01
Hamonize commands argument names base-console to base-command

eko:
I've harmonized commands using both `base-console` and `base-command` option to `base-command`.

This is related to issue https://github.com/sonata-project/SonataPageBundle/issues/271


eko:
@Bladrak @rande,

I've rebased this PR in case you would merge it.

Indeed, some commands are still using a `base-command` option (like `sonata:page:create-site` or `sonata:page:update-core-routes`) and these are using `base-console`.

The goal of this PR was to use the `base-command` option for all commands but keep the backward compatibility with `base-console`.

Bladrak:
Still seems important to me, @rande thoughts?

eko:
I close this ticket because I don't have any news for 1 year and want to clean my pull request history.

PR #331, opened 2014-05-12
Restrict core bundle version to dev-master

greg0ire:
This bundle relies on PageableManagerInterface, which is only available
on the master branch of the core bundle. Specifying @dev is not enough
to ensure this. It allows me to install dev-master, but does not force
me to do it, and since I have prefer-stable set to true, Composer
installs 2.2.5, which does not have the interface.


OskarStark:
any news on this @greg0ire

greg0ire:
This can be closed, the bundle now relies on `^2.3.1`

OskarStark:
:+1:  thank you @greg0ire

PR #339, opened 2014-05-14
Solve requirement problem

greg0ire:
The parent needs to be required when there is already a root the site. Also, this PR contains a small code simplification.


OskarStark:
Any news on this PR @greg0ire

greg0ire:
This was done way too long ago, I cannot be sure it is a good thing to merge now, so let's close it too.

PR #384, opened 2014-08-28
Update installation.rst

Alekc:
Uniforming instruction to those in SonataUserBundle (also files created under app directory would not work anyway, it can confuse beginners)


OskarStark:
Done by 534b9848a8b39f03bb4184b5b7d910884fe3ec86

PR #387, opened 2014-09-08
order found snapshots by publicationDateStart

koyaan:
It is counterintuitive that if more then one snapshot matches the date criteria, the one that was created first is selected, it should be the one with the newest startDate.

### DIFF
--- a/Entity/SnapshotManager.php
+++ b/Entity/SnapshotManager.php
@@ -131,6 +131,7 @@ class SnapshotManager extends BaseEntityManager implements SnapshotManagerInterf
             throw new \RuntimeException('please provide a `pageId`, `url`, `routeName` or `name` as criteria key');
         }
 
+        $query->orderBy("s.publicationDateStart","desc");
         $query->setMaxResults(1);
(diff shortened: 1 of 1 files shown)


OskarStark:
This is BC because you set ``$query->setMaxResults(1);``

Soullivaneuh:
> This is BC because you set `$query->setMaxResults(1);`

I don't think so. This code part is already here.

OskarStark:
yes, but it changed the behavior/result by sorting it another why, doesn't it?

Soullivaneuh:
According to the new [Sonata version management](https://github.com/sonata-project/SonataAdminBundle/issues/3053) and next major [release plan](https://github.com/sonata-project/SonataAdminBundle/issues/3731), this project has been refactored regarding branching and versioning.

If you see this message, your PR concerns a patch or a minor release and is not targeting the right branch.

So I'm closing this one, but don't see it as a refusal. If you think your work is still relevant and want to continue, feel free to reopen it on the right branch (e.g. the default one).

Regards.

PR #402, opened 2014-10-05
Fixed paths for bootstrap and jquery

Alekc:
Bootstrap and Jquery has been moved avay from sonataadmin, so current situation breaks the sites as it is. 

Fixed wrong paths.


OskarStark:
This is already fixed here: https://github.com/sonata-project/SonataPageBundle/blob/master/DependencyInjection/Configuration.php#L140

If i missed something feel free to reopen this PR.

Thanks for your contribution! :+1: 

Oskar

PR #442, opened 2015-01-08
fix(composer): add composer parmeter

amine26tz:


Resources/public/sonata-page.back.js
```diff
@@ -498,7 +498,7 @@
                 $loader.show();
 
                 $.ajax({
-                    url:     url,
+                    url:     url + '&' + $.param({'composer': 1}),
```
rande: why the url does not contains the composer query string in the twig template ?

rande: does not look to be a valid solution

amine26tz: @rande the from is genrated into CRUDController (createAction), i think its more complicated to inject other parameters from the twig.

PR #456, opened 2015-02-16
Fix SnapshotManager enableSnapshots method to correctly end all previous snapshots

eko:
Currently, `SnapshotManager::enableSnapshots()` method is setting the actual date to all page snapshots that are not the current one (we just created) AND where the end publication date is null which is not correct as there are snapshots with an end date already set.

This can result in multiple active snapshots for a single page, which is not correct.

See actual query:

```php
UPDATE %s SET publication_date_end = '%s' WHERE id NOT IN(%s) AND page_id IN (%s) AND publication_date_end IS NULL
```

Main goal of this PR is to remove this part: `AND publication_date_end IS NULL`.

I've also added a second optional parameter to `enableSnapshots()` method with a date (mostly to unit test the date passed to the executed SQL query), but it also could be helpful if someone needs to enable snapshots after a given date.

### DIFF
--- a/Entity/SnapshotManager.php
+++ b/Entity/SnapshotManager.php
@@ -63,19 +63,20 @@ class SnapshotManager extends BaseEntityManager implements SnapshotManagerInterf
     /**
      * {@inheritdoc}
      */
-    public function enableSnapshots(array $snapshots)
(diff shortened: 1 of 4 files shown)


eko:
Ping @rande

PR #465, opened 2015-03-10
Avoid starting session when checking for session variable

artursvonda:


CmsManager/CmsManagerSelector.php
```diff
@@ -67,7 +67,10 @@ public function isEditor()
          * but do the job for now.
          */
 
-        return $this->getSession()->get('sonata/page/isEditor', false);
+        $request = $this->getRequest();
```
rande: You will a fatal error if ``getRequest`` return null...

artursvonda: Added check for request existence.

artursvonda:
ping @rande

PR #478, opened 2015-03-30
Adding translations in the tree and some spanish translations

orlandordzch:


orlandordzch:
sorry for all the commits, it's not going to happen again :(

rande:
Can you squash your commit ? Also you need to keep the key as target for non translated message.

PR #494, opened 2015-06-01
Updating oracle deletion query

Deamon:
In Oracle, there should not be the AS keyword in the where sub-select otherwise an '00907. 00000 -  "missing right parenthesis"' will occurs.

If there is any question about this pr, feel free to ask!

Bastien


Deamon:
Thank you @rande, do you plan to create a new tag that includes this PR soon ?

PR #509, opened 2015-07-27
Fix assets paths for bootstrap JS and CSS files

borNfreee:
SonataAdminBundle does not contain bootstrap assets anymore.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -96,15 +96,15 @@ class Configuration implements ConfigurationInterface
                 ->children()
                     ->arrayNode('stylesheets')
                         ->defaultValue(array(
-                                'bundles/sonataadmin/vendor/bootstrap/dist/css/bootstrap.min.css',
+                                'bundles/sonatacore/vendor/bootstrap/dist/css/bootstrap.min.css',
                                 'bundles/sonatapage/sonata-page.front.css',
                             ))
                         ->prototype('scalar')->end()
                     ->end()
                     ->arrayNode('javascripts')
                         ->defaultValue(array(
-                                'bundles/sonataadmin/vendor/jquery/dist/jquery.min.js',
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -104,7 +104,7 @@ public function getConfigTreeBuilder()
                     ->arrayNode('javascripts')
                         ->defaultValue(array(
                                 'bundles/sonataadmin/vendor/jquery/dist/jquery.min.js',
```
Soullivaneuh: I think this could be changed too, isn't it?

borNfreee: Fixed

DependencyInjection/Configuration.php
```diff
@@ -96,15 +96,15 @@ public function getConfigTreeBuilder()
                 ->children()
                     ->arrayNode('stylesheets')
                         ->defaultValue(array(
-                                'bundles/sonataadmin/vendor/bootstrap/dist/css/bootstrap.min.css',
+                                'bundles/sonatacore/vendor/bootstrap/dist/css/bootstrap.min.css',
                                 'bundles/sonatapage/sonata-page.front.css',
                             ))
                         ->prototype('scalar')->end()
-                    ->end()
+ so                   ->end()
```
Soullivaneuh: Hmm... I don't think this should be here. :-)

borNfreee: So stupid mistake :) Fixed

Soullivaneuh:
Thanks. Could you please now squash your commit?

Soullivaneuh:
Before squashing, please fix the tests.

borNfreee:
@Soullivaneuh fixed tests related to my update and squashed.

There is only one failed test, but this is related to Page::slugify() method.

Soullivaneuh:
Thank you @borNfreee.

PR #517, opened 2015-08-05
Missing semicolon in sonata-page.front.js

core23:
This can be a problem when concatenating with other JS files.


core23:
ping @rande

PR #518, opened 2015-08-13
Fix issue with template containers authorized blocks and composer dnd

plouc:


rande:
:+1: thanks @plouc it was the missing feature to remove the beta flag of compose.

PR #526, opened 2015-09-10
[MOD] Extract javascript of compose page to appropriate block

Th3Mouk:
for 2.3.10

PR #527, opened 2015-09-10
[MOD] Extract javascript of compose page to appropriate block

Th3Mouk:
for 2.4.x-dev

### DIFF
--- a/Resources/views/PageAdmin/compose.html.twig
+++ b/Resources/views/PageAdmin/compose.html.twig
@@ -83,30 +83,32 @@ file that was distributed with this source code.
                     </div>
                 </div>
             </div>
-            <script>
-                $(document).ready(function () {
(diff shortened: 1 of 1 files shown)


Resources/views/PageAdmin/compose.html.twig
```diff
@@ -83,30 +83,32 @@ file that was distributed with this source code.
...
-                    window.SonataPageComposer = composer;
-                });
-            </script>
         {% endif %}
     </div>
+{% endblock %}
 
+{% block javascripts %}
+    {{ parent() }}
+    <script>
```
Soullivaneuh: Missing `type="text/javascript"`.

Soullivaneuh:
We merge branches. #526 will be the base.

Th3Mouk:
ok i will see because the script is different so i was thinking merge isn't possible

PR #529, opened 2015-09-13
Missing translation key

core23:
This key comes from `Page:create.html.twig`


Soullivaneuh:
As a bugfux, could you please do this PR on last stable branch (`2.3`)? Thanks.

PR #531, opened 2015-09-17
[FIX] PHPDoc & clean code with scrutinizer

Th3Mouk:


Command/RenderBlockCommand.php
```diff
@@ -53,8 +53,6 @@ public function execute(InputInterface $input, OutputInterface $output)
             throw new \RuntimeException('The service does not implement the CmsManagerInterface');
         }
 
-        $page = $manager->getPageById($input->getArgument('page_id'));
```
Soullivaneuh: Why removing this?

rande: This code is required as it load some data

Th3Mouk: IMO this variable is unused, i missed some reasoning ?

rande: The variablie is not used, but the getPageById load internal data.

Th3Mouk:
@rande there is an interest in correcting the phpdoc for you or not?

rande:
Yes, indeed ;)

Command/RenderBlockCommand.php
Th3Mouk: sorry i feel like a monkey at the end of afternoon =)

Model/PageInterface.php
```diff
@@ -244,7 +244,7 @@ public function addChildren(PageInterface $children);
     /**
      * Get children.
      *
-     * @return array
+     * @return Doctrine\Common\Collections\ArrayCollection|array
```
OskarStark: please use a ``use`` statement here

Model/PageInterface.php
```diff
@@ -258,7 +258,7 @@ public function addBlocks(PageBlockInterface $block);
     /**
      * Get blocs.
      *
-     * @return array
+     * @return Doctrine\Common\Collections\ArrayCollection|array
```
OskarStark: same here

Command/RenderBlockCommand.php
OskarStark: please revert this code change

OskarStark:
Thank you for your contribution @Th3Mouk 

After applying the comments and some final review, we will merge your PR :+1: 

Oskar

OskarStark:
Restarted travis

Th3Mouk:
Should I do something ?

OskarStark:
Thank you @Th3Mouk

PR #532, opened 2015-09-21
typo

OskarStark:


Model/Page.php
```diff
@@ -480,15 +480,15 @@ public function setTarget(PageInterface $target = null)
...
 
     /**
-     * Add blocs.
+     * Add blocks.
      *
-     * @param PageBlockInterface $blocs
+     * @param PageBlockInterface $blocks
      */
-    public function addBlocks(PageBlockInterface $blocs)
+    public function addBlocks(PageBlockInterface $blocks)
```
OskarStark: @rande same here, this should be ```addBlock``` and ```$block``` -> singular!

PR #533, opened 2015-09-21
typo

OskarStark:


Model/PageInterface.php
```diff
@@ -249,14 +249,14 @@ public function addChildren(PageInterface $children);
     public function getChildren();
 
     /**
-     * Add blocs.
```
rande: I guess there is also an extra ``s``

OskarStark: yes, but i can't change the method name, without breaking BC

so lets do this in another PR, because ```addBlockS``` should be ```addBlock```....

this is symfony generating stuff :(

PR #535, opened 2015-09-25
Fix block admin controller

qsomazzi:


OskarStark:
thank you @qsomazzi 

can you please describe the bug and the fix a little bite more?

qsomazzi:
Yeah of course, 

This quick fix goal is to handle the right parameter when you switch a block to another parent.

Before : 
We check if the block exist, then get the parent, and re-check is the block exist
After:
We check if the block exist, then get the parent, then check is the parent exist

OskarStark:
ping @Soullivaneuh @rande

qsomazzi:
@rande can you merge it ? it's a really small fix :)

PR #548, opened 2015-10-19
Fix composer preview

qsomazzi:
By replacing in twig `url` with `path` we allow user to use the protocol https (because path doesn't render the protocol, so it's the one currently used)


rande:
ok to test

qsomazzi:
Vu avec @rande. Ok to merge

PR #549, opened 2015-10-20
set enabledTo required=false in SiteAdmin

OskarStark:
BEFORE:
bildschirmfoto 2015-10-20 um 08 57 17

AFTER:
bildschirmfoto 2015-10-20 um 08 58 29

Otherwise you cannot save the ``Site`` if you don't set this value


OskarStark:
Refs FriendsOfPHP/PHP-CS-Fixer#1450

PR #561, opened 2015-11-05
Added deprecation info to snapshot sources property

core23:
Never used like the same property in the ``Page`` model.

### DIFF
--- a/Model/Snapshot.php
+++ b/Model/Snapshot.php
@@ -103,6 +103,9 @@ abstract class Snapshot implements SnapshotInterface
      */
     protected $parentId;
 
+    /**
+     * @deprecated since version 2.4 and will be removed in 3.0.
+     */
     protected $sources;
 
     protected $target;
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Please add `trigger_error` function inside deprecated methods.

core23:
Done @Soullivaneuh

Model/Snapshot.php
```diff
@@ -387,18 +390,22 @@ public function getParentId()
     }
 
     /**
-     * {@inheritdoc}
+     * @deprecated This method is deprecated since version 2.4 and will be removed in 3.0.
```
Soullivaneuh: Remove "This method is deprecated". `@deprecated` already says that.

Model/Snapshot.php
```diff
@@ -387,18 +390,22 @@ public function getParentId()
...
     public function setSources($sources)
     {
+        trigger_error('The '.__METHOD__.' method is deprecated since version 2.4 and will be removed in 3.0.', E_USER_DEPRECATED);
+
         $this->sources = $sources;
     }
 
     /**
-     * {@inheritdoc}
+     * @deprecated This method is deprecated since version 2.4 and will be removed in 3.0.
```
Soullivaneuh: Remove "This method is deprecated". `@deprecated` already says that.

core23:
Changed depreaction text

OskarStark:
ping @Soullivaneuh

PR #562, opened 2015-11-06
Remove inline scripts

core23:
For security reasons we should remove all inline scripts and instantiate the elements via data listeners.


rande:
Can you provide more insight about the security issue ?

core23:
There is a bundle (https://github.com/nelmio/NelmioSecurityBundle#content-security-policy) which introduces some interesting information about security. You could completly disable inline script via a http header to reduce possible xss attacks.

More information
http://www.w3.org/TR/CSP/

rande:
The code you altered is about page management for the editor and so the admin. The current admin bundle implementation rely a lot on inline javascript. Not sure this commit make sense as if the user disable inline script the admin bundle will not work.

core23:
You're right. Today the admin bundle relies on many inline script and requires a lot of refactoring, but someone should start with it to remove all inline stuff. :smiley:

This code makes a small step to a more secure bundle. If you're okay with it I would refactor the edit/create scripts too: https://github.com/sonata-project/SonataAdminBundle/issues/3173

core23:
@rande Are there any other reasons agains this PR?

Resources/assets_src/src/js/page.js
```diff
@@ -492,4 +492,13 @@ Sonata.Page = {
 
         }
     }
-}
\ No newline at end of file
+};
+
+
+// auto-initialize plugin
+$(function () {
```
rande: This should be call when the dom is loaded ?

core23: Yes, it does as expected. It is called when DOM is ready. It is the shorthand version for it (https://learn.jquery.com/using-jquery-core/document-ready/)

rande: Ok, good to know.

PR #565, opened 2015-11-07
BlockAdmin template chooser

core23:
Added an optional template chooser to the block admin, where the original template is shown as _default_. If only one template is available no template chooser is visible. 

Before:
bildschirmfoto 2015-11-07 um 16 41 24

After:
bildschirmfoto 2015-11-07 um 16 41 01

```yaml
# config.yml
sonata_block:    
    blocks:
        core23.project.block.projects:
            templates:
              - { name: 'Custom', template: 'Core23ProjectBundle:Block:block_projects_custom.html.twig' }
              - { name: 'Foo', template: 'Core23ProjectBundle:Block:block_projects_foo.html.twig' }
```

This PR adds the missing template list to blocks https://github.com/sonata-project/SonataBlockBundle/pull/237

### DIFF
--- a/Admin/BlockAdmin.php
+++ b/Admin/BlockAdmin.php
@@ -14,7 +14,9 @@ namespace Sonata\PageBundle\Admin;
 use Doctrine\ORM\EntityRepository;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Route\RouteCollection;
+use Sonata\BlockBundle\Block\BlockServiceInterface;
 use Sonata\PageBundle\Model\PageInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Admin class for the Block model.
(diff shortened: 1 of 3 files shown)


Admin/BlockAdmin.php
```diff
@@ -65,6 +87,8 @@ protected function configureFormFields(FormMapper $formMapper)
             }
         }
 
+        $type  = $block->getType();
```
OskarStark: IMO ``$blockType`` is a better var name

Admin/BlockAdmin.php
```diff
@@ -144,6 +188,24 @@ protected function configureFormFields(FormMapper $formMapper)
     }
 
     /**
+     * @param BlockServiceInterface $blockService
+     *
+     * @return string|null
+     */
+    private function getDefaultTemplate(BlockServiceInterface $blockService)
+    {
+        $resolver = new OptionsResolver();
```
OskarStark: this should work?

Admin/BlockAdmin.php
```diff
@@ -26,6 +28,26 @@ class BlockAdmin extends BaseBlockAdmin
     protected $parentAssociationMapping = 'page';
 
     /**
+     * @var array
```
OskarStark: can we type hint this a bit better?

core23: It's difficult as it is an array of arrays

OskarStark: :+1: ok

OskarStark:
ping @rande

core23:
@rande ping

core23:
@rande What do you think about a block template chooser?

rande:
@core23 I think it is a good idea!  Can you add documentation too ? Thanks.

core23:
@rande Added a short info to the ``page_composer.rst``. Is this sufficient for this feature?

OskarStark:
can you please check travis or rebase? @core23

core23:
The build error is not related to this PR, the current master has the same problems.

OskarStark:
Thank you @core23, great work :+1:

PR #568, opened 2015-11-11
Composer: Translated block name and icon

core23:
Added block service icon and translated name.

bildschirmfoto 2015-11-11 um 18 18 42


OskarStark:
Do we have a test for this somewhere?

core23:
Travis is building green. So there should be no front end tests for this

OskarStark:
yes, but you introduced the icon, so we have to be sure that this will be rendered correctly, and therefore we need a test IMO

rande:
@OskarStark this test need to be done on the sandbox with an ajax call. There is not functionnal tests inside a bundle.

So this PR is ready to merge

OskarStark:
Thank you

PR #572, opened 2015-11-16
Updated German translation

core23:
@OskarStark What do you think?


OskarStark:
hmm, i think layout is a bit missleading, but i don't like composer, too because its an english word... :smile: 

``Layout-Block-Editor`` ?

core23:
IMO ``Layout-Block-Editor`` seems too much technical. Maybe only ``Editor`` which comes near to the word ``composer``?

OskarStark:
Editor is :+1:

Am Dienstag, 17. November 2015 schrieb Christian Gripp :

core23:
RTM @OskarStark

OskarStark:
Thank you @core23

PR #581, opened 2015-11-26
Fix RedirectResponse bug caused by HostPathSiteSelector::onKernelRequestRedirect

b4zs:
I have been using the RedirectController to create url redirects to external urls, but using the HostPathSiteSelector ruins the absolute urls. In case the path is http://www.google.com the calculated url is the following: 

`/enhttp://www.google.com`

In order to fix this, an additional condition should be added.


OskarStark:
sounds reasonable, could you add a test case for your regex?

and please rebase and execute `make cs` to fix the checkstyle

thank you

PR #585, opened 2015-12-02
cs fix

OskarStark:


Route/CmsPageRouter.php
```diff
@@ -337,6 +337,6 @@ protected function isPageAlias($name)
      */
     protected function isPageSlug($name)
     {
-        return (is_string($name) && $name == PageInterface::PAGE_ROUTE_CMS_NAME);
+        return is_string($name) && $name == PageInterface::PAGE_ROUTE_CMS_NAME;
```
OskarStark: @Soullivaneuh shall we use yoda here?

Soullivaneuh: Technically yes but you can do it on another PR.

OskarStark: is there a check for the php cs fixer?

PR #588, opened 2015-12-06
Corrected indentation and some small typos

cezar77:
The indentation of the list items was not correct.
There were also some minor errors like a HTML -> an HTML, ie -> i.e.
The expression "different types of page" doesn't sound very good to me, so I changed that to "different page types".

### DIFF
--- a/Resources/doc/reference/introduction.rst
+++ b/Resources/doc/reference/introduction.rst
@@ -1,9 +1,9 @@
 Introduction
 ============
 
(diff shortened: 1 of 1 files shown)


OskarStark:
can you please rebase this PR, looks like the travis.yml is not correct

thank you

OskarStark:
ah sorry, this is 2.2, can you please check 2.3, too?

PR #598, opened 2016-01-16
UI: Show modal dialog in page composer

core23:
Replaces the inline confirm request with a modal dialog.

Before:
bildschirmfoto 2016-01-16 um 22 40 27

After:
bildschirmfoto 2016-01-16 um 22 32 08


OskarStark:
@core23, can you please check the travis build?

core23:
There is something wrong with travis... This is not related to this PR, see some recent merged commits.

OskarStark:
Thank you @core23

rande:
Are you sure it is an improvement ?

core23:
The problem with the old implementation is, that the confirm panel is hard to handle. If you hover over an other item the question and button are hidden. 

IMO it also looks crap

PR #608, opened 2016-01-29
Fix broken build

### LINKED ISSUE #607: Documentation is broken
Build is not passing, because something is wrong with sphinx

see sphinx-doc/sphinx#2264

greg0ire:
Fixes #607 
See https://github.com/symfony-cmf/RoutingBundle/issues/317

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ matrix:
 
   allow_failures:
     - php: hhvm
+    - env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
 
 before_script:
   - (phpenv config-rm xdebug.ini || exit 0)
--- a/Resources/doc/reference/api.rst
+++ b/Resources/doc/reference/api.rst
@@ -46,7 +46,7 @@ Here's the configuration we used, you may adapt it to your needs:
             metadata:
                 directories:
                     -
-                        path: %kernel.root_dir%/../vendor/sonata-project/datagrid-bundle/Resources/config/serializer
+                        path: "%kernel.root_dir%/../vendor/sonata-project/datagrid-bundle/Resources/config/serializer"
                         namespace_prefix: 'Sonata\DatagridBundle'
 
 In order to activate the API's, you'll also need to add this to your routing:
--- a/Resources/doc/reference/api.rst
+++ b/Resources/doc/reference/api.rst
@@ -45,7 +45,9 @@ Here's the configuration we used, you may adapt it to your needs:
         jms_serializer:
             metadata:
                 directories:
-                    - { path: %kernel.root_dir%/../vendor/sonata-project/datagrid-bundle/Resources/config/serializer, namespace_prefix: 'Sonata\DatagridBundle' }
+                    -
+                        path: %kernel.root_dir%/../vendor/sonata-project/datagrid-bundle/Resources/config/serializer
+                        namespace_prefix: 'Sonata\DatagridBundle'
 
 In order to activate the API's, you'll also need to add this to your routing:
 
--- a/Resources/doc/reference/page_composer.rst
+++ b/Resources/doc/reference/page_composer.rst
@@ -102,54 +102,54 @@ The **PageComposer** js object trigger various events to allow customization:
 
 .. code-block:: javascript
 
-   event.containerId # loaded container id
+   event.containerId // loaded container id
 
 
 **blockcreated**, event properties:
 
 .. code-block:: javascript
 
-   event.$childBlock # created block jQuery element
-   event.parentId    # created block parent id
-   event.blockId     # created block id
-   event.blockName   # created block name
-   event.blockType   # created block type
+   event.$childBlock // created block jQuery element
+   event.parentId    // created block parent id
+   event.blockId     // created block id
+   event.blockName   // created block name
+   event.blockType   // created block type
 
 
 **blockremoved**, event properties:
 
 .. code-block:: javascript
 
-   event.parentId # removed block parent id
+   event.parentId // removed block parent id
 
 
 **blockcreateformloaded**, event properties:
 
 .. code-block:: javascript
 
-   event.response    # the raw html response (form)
-   event.containerId # current container id
-   event.blockType   # selected block type
+   event.response    // the raw html response (form)
+   event.containerId // current container id
+   event.blockType   // selected block type
 
 
 **blockpositionsupdate**, event properties:
 
 .. code-block:: javascript
 
-   event.disposition # a javascript object containing all child blocks position/ids…
+   event.disposition // a javascript object containing all child blocks position/ids…
 
 
 **blockeditformloaded**, event properties:
 
 .. code-block:: javascript
 
-   event.$block # the block jQuery element
+   event.$block // the block jQuery element
 
 
 **blockparentswitched**, event properties:
 
 .. code-block:: javascript
 
-   event.previousParentId # previous parent block id
-   event.newParentId      # new parent block id
-   event.blockId          # child block id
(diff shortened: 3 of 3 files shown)


OskarStark:
@greg0ire can you please squash your commits and ping me when this is RTM?

composer.json
```diff
@@ -17,6 +17,7 @@
         }
     ],
     "require": {
+        "symfony/security-acl": "~2.4|~3.0",
```
core23: Think we should add this to ``require-dev``

greg0ire:
Again here, I squashed what I could, but all these changes have little to do with each other…

composer.json
```diff
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "symfony/symfony": "~2.3|~3.0",
+        "symfony/symfony": "~2.3",
```
core23: Why did you remove symfony 3? :-1:

greg0ire: Read the commit message ;)

greg0ire: So? What do you think?

greg0ire: I think I know what makes you say that : nothing in the SonataPageBundle code is incompatible with symfony 3.0, only its dependencies are incompatible… but do we really know that, since probably nobody has ever installed SonataPageBundle along with sf 3.0 because of this incompatibility?

core23:
When you say, that your commits have little to do with each other, please create atomic PRs for your changes so we could choose which one we will merge.

greg0ire:
I want to fix several independant problems that all lead to one thing : the build being fixed. How can I know for sure the build will be fixed if I create separate PRs (hint: I can't). I think you guys should relax a bit about the rules you want to blindly apply and see the advantages (plural) that this way of doing things provide as opposed to squash all commits just because it is easier to revert (BTW, is it?). Which are : 

- better commit messages : it is better to have detailed commit message for each commit, explaining each big decision, than just one big message that gives a hazy description of each when it does, or even sums up to "fix build"
- better diffs : you don't want to see my cs commit in your diff, it just pollutes the diff. And no, I'm not creating separate PRs for cs commits. I'd rather not contribute these commits than do a separate PR just for things I find along the way, because there is very little things to discuss about CS commits.
- better revertability : @rande says "@greg0ire just squash your commit please ;) one PR = one commit, easier to revert ;)"… Really?!? The bigger the commit you revert, the more likely you are to get a big conflict for a cs fix you don't really want to revert anyway! If I commit A and B together, and then some dev creates C based on what I did in A, and then someone wants to revert just B, they will get a conflict with C just because of that. So sorry, but no, it's not easier, it might actually be harder.

It's one thing to be about two different concerns, it's another thing to be independent. One commit that fixes CS on a method, and one commit that improves the method are very related. And very dependent. Making separates PR for them would be hard, and both PRs would actually **conflict** with each other. Commit are what should be atomic. Not PRs

To conclude this, if you really want to pick one of my commits and not the others, just say so, and I'll remove them from the PR. It's as simple as that.

OskarStark:
i think in general you @core23 and @greg0ire are right, but in this case to get back to stable i think it is ok. we should decide this from PR to PR and i think that @greg0ire does a good job on the sonata project and i know that he would like to bring this to a success.

Other contributers, which are new or not so experienced should follow the, let's call it "best practices for PR's"

IMO it doesn't matter if this is one commit or on PR in case of reverting, because you would revert the merge-commit of this PR, not each of the commits.

i like small commits and don't be a fan of squashing the whole stuff into one PR-commit... but the decision was done by @rande 

maybe @javiereguiluz can say something about this topic, how other bundles handle this

anyway thank you both for your hard work on each topic! :+1:  :beers:

greg0ire:
Thanks @OskarStark for understanding me. I saw this decision was made, and it appears, that it was also made for symfony, they even have a tool to automatically squash the commits. But no explanation whatsoever… strange.

When in doubt about how to use git, looking at the kernel commit is in general a good idea, and if you look at https://github.com/torvalds/linux/commit/b37a05c083c85c2657dca9bbe1f5d79dccf756d5 it merges a branch with 18 commits.

But anyway… the only thing that really blocks this PR now is [the ACL package problem](https://github.com/sonata-project/SonataAdminBundle/pull/3563).

javiereguiluz:
@OskarStark sadly other bundles ask people to squash their commits too.

In my opinion, bundles should do the effort to make contributions as easy as possible. Personally, there are two phrases that I hate to read when contributing: "Please, rebase" and "Please, squash". I'm always tempted to close my pull request and never contribute again to those projects.

In my bundle I use a proprietary tool (developed by Fabien) that squashes everything automatically before merging, so I never bother my contributors about this. And regarding the rebase, I usually do that myself too to not bother contributors.

greg0ire:
I think this should be merged now, and aab2de5 should be reverted when 2.3.8 is out and has the correct requirement, because people won't be able to get green builds until then (which is bad because they might break other tests without noticing).

composer.json
```diff
@@ -32,6 +32,7 @@
     },
     "require-dev": {
         "sonata-project/doctrine-orm-admin-bundle": "~2.2",
+        "symfony/security-acl": "~2.4",
```
core23: You can remove this as the latest AdminBundle has this dependency

greg0ire: Let's try, but I think it will fail since 2.3.7 does not have it.

rande:
it is too bad that we lost time (that I don't have much those days :( ) on discussing about rebase and squash. 

On the PR, we need to keep sf3.0 in composer.json as it help porting the bundle to this symfony version. however we can move to allow fealures.

greg0ire:
I moved 3.0 to allowed failures.

greg0ire:
About the rebase vs squash debate, sorry but it really is that important to me. I am the first to tell people to do meaningful, small commits, so I really can't do big commits with big messages and unrelated changes on them. I don't know where this trend comes from, but it really is bad practice IMO. I'd like to be at least proven wrong instead of having everyone just tell me I am wrong.

OskarStark:
in my opinion THIS PR with currently 3 commits looks :+1: 

i would merge it

greg0ire:
Thanks @OskarStark ! People working on 3.0 compatibility should revert df15381fca75aeffc1bd99b1517dde781e2e0b05 when they are done.

core23:
I can't figure out why the symfony 3.0 dependency is breaking... But I think we should merge this to become stable.

greg0ire:
Wait a minute… why doesn't 3.0 show up in the build under "allowed failures" ?

greg0ire:
BTW, the 3.0 bundle is failing because the page bundle requires a package that is incompatible with 3.0 : cmf/routing-bundle

core23:
You have to define them under "include" first. @greg0ire

greg0ire:
Oh ok! Doing that rightaway!

greg0ire:
That's better, but… why is 5.6 allowed to fail?

greg0ire:
Looks like a Travis bug to me :/

greg0ire:
I'll try the single line syntax.

greg0ire:
Doesn't work… reverting to the previous commit.

greg0ire:
@core23 : I'm a bit out of ideas : do you have some?

core23:
This should work @greg0ire 
```
  allow_failures:
    - php: hhvm
    - env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
```

greg0ire:
Indeed! Not sure how this works, but thanks!

greg0ire:
Ready To Merge :)

core23:
Thank you @greg0ire

greg0ire:
Thank you for the help and for merging!

PR #609, opened 2016-02-02
Fix issue with block removing

qsomazzi:


qsomazzi:
ping @rande

OskarStark:
@core23 can you verify this PR?

@qsomazzi what is the exact problem?

qsomazzi:
The problem is that sometimes, when I remove a block, the background-layer of the modal stay, and the modal fade properly. So I can't use the website anymore, we need to remove the backdrop to use the app again. My PR is just a fix to prevent this when it's appear.

qsomazzi:
see with @rande

PR #613, opened 2016-02-15
Fix fetching orphaned containers in composer view

StrikeForceZero:
##### Orphaned containers were failing to render. 

---

The ajax call silently failed because it was attempting to get the href attribute on the `li` element for the container list. It would inject the current composer view into the right hand view effectively creating the "window in a window effect".

- moved class and data attributes into anchor tag,
- expanded anchor tag using `display:block` to fill the width of the li parent element
- compiled sass into css

The `PageAdminController` was failing to render the orphaned container as it was attempting to get the container from the current template (which doesn't exist, thus passing an empty array and throwing an undefined index error). 

- added simple check to prevent error in PageAdminController

---

I'm not entirely sure of the side effects for not running:

```php
foreach ($blockServices as $code => $service) {
   if (in_array($code, $container['blocks'])) {
       continue;
   }

   unset($blockServices[$code]);
}
```

but now the user can view and edit orphaned blocks


OskarStark:
@core23 can you please review this PR? thank you :+1:

core23:
Thank you @StrikeForceZero

PR #622, opened 2016-03-08
Fix page creation

rylorin:
When creating a new page: logged as admin, going to non existing page,
asking to create it and validating on page admin. bug can be seen on
Sonata demo.

### DIFF
--- a/Admin/PageAdmin.php
+++ b/Admin/PageAdmin.php
@@ -333,8 +333,13 @@ class PageAdmin extends Admin
             $slugs = explode('/', $this->getRequest()->get('url'));
             $slug  = array_pop($slugs);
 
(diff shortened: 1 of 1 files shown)


Admin/PageAdmin.php
```diff
@@ -333,8 +333,13 @@ public function getNewInstance()
             $slugs = explode('/', $this->getRequest()->get('url'));
             $slug  = array_pop($slugs);
 
+            if (count($slugs) == 1) {
+               	$parent_url = '/';
+            } else {
+               	$parent_url = implode('/', $slugs);
```
OskarStark: wrong indention

rylorin:
Thanks

PR #646, opened 2016-05-28
Use the page name if the page has no SEO titel

core23:
Rebase of #420

### Changelog

```markdown
### Fixed
- The page name is now correctly used as a page title fallback
```

### Subject

The page name fallback could never be reached because of the surrounding if-statement.

### To do

- [X] Update the tests

### DIFF
--- a/Page/Service/DefaultPageService.php
+++ b/Page/Service/DefaultPageService.php
@@ -73,9 +73,7 @@ class DefaultPageService extends BasePageService
             return;
         }
 
-        if ($page->getTitle()) {
-            $this->seoPage->setTitle($page->getTitle() ?: $page->getName());
-        }
+        $this->seoPage->setTitle($page->getTitle() ?: $page->getName());
 
         if ($page->getMetaDescription()) {
             $this->seoPage->addMeta('name', 'description', $page->getMetaDescription());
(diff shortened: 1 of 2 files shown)


OskarStark:
@greg0ire `patch` is correct?

core23:
IMO yes, because there is a logic that will use the page name, but a wrong if-statement makes it unreachable.

greg0ire:
Kein deutsch bitte! "titel" => "title"

greg0ire:
Well I'll fix that with squash merge

greg0ire:
Here you go! Danke schön @core23 !

jaxel87:
This commit breaks seo bundle functionality. Now I can't set  Seo Title values at runtime. Sonata Page Title always replace them.
`$seoPage = $this->container->get('sonata.seo.page');

$seoPage->setTitle($post->getTitle())`

core23:
Can you please show a little bit more of your code?

OskarStark:
@core23 do we have a test for this?

Soullivaneuh:
Maybe adding `is_null($this->seoPage->getTitle())` to the condition will solve the issue? Didn't test.

jaxel87:
$seoPage->setTitle($title) called 3 times. First set titile from config. Second - at runtime. Third - if page have title in admin. SeoPage Title should not be overwritte, if $page->getTitle() is null.

`        if ($page->getTitle()) {
            $this->seoPage->setTitle($page->getTitle());
        }

        if (!$this->seoPage->getTitle()) {
            $this->seoPage->setTitle($page->getName());
        }
`

core23:
IMHO this is another issue.

The title was always overwritten, the only thing that was changed is the fallback to the page name. Maybe you didn't have a page title, so your code just works before the patch.

PR #657, opened 2016-05-29
sort snapshots by publicationDateStart desc

OskarStark:
Replaces #387

### Changelog

```markdown
### Changed
- Sort snapshots by `publicationDateStart`
```

### Subject

It is counterintuitive that if more then one snapshot matches the date criteria, the one that was created first is selected, it should be the one with the newest startDate.

### DIFF
--- a/Entity/SnapshotManager.php
+++ b/Entity/SnapshotManager.php
@@ -133,6 +133,7 @@ class SnapshotManager extends BaseEntityManager implements SnapshotManagerInterf
             throw new \RuntimeException('please provide a `pageId`, `url`, `routeName` or `name` as criteria key');
         }
 
+        $query->orderBy('s.publicationDateStart', 'DESC');
         $query->setMaxResults(1);
         $query->setParameters($parameters);
 


OskarStark:
@greg0ire can you please merge this?

Entity/SnapshotManager.php
```diff
@@ -133,6 +133,7 @@ public function findEnableSnapshot(array $criteria)
             throw new \RuntimeException('please provide a `pageId`, `url`, `routeName` or `name` as criteria key');
         }
 
+        $query->orderBy('s.publicationDateStart', 'DESC');
```
Soullivaneuh: And if the user does not want this order?

Soullivaneuh: BTW for me it's BC break.

By default, it's ordered by ID. So if a user take the first element considering id order, it would not get the same result after that.

OskarStark: I am confused: https://github.com/sonata-project/SonataPageBundle/pull/387#issuecomment-149477956

core23: It's a BC because it changes the expected behavior of a method.

Soullivaneuh: > It's a BC because it changes the expected behavior of a method.

@core23 You mean BC **break**?

core23: Yes, sorry. I meant BC **break**

core23:
Please move this to the `master` branch @OskarStark

core23:
Please rebase this against the `master` branch @OskarStark

PR #662, opened 2016-06-01
use \RuntimeException instead of unavailable \RunTimeException

OskarStark:
### Changelog

```markdown
### Fixed
- Typo on `\RuntimeException` usage
```

### Subject

`\RunTimeException` is not available

### DIFF
--- a/Tests/Entity/SnapshotManagerTest.php
+++ b/Tests/Entity/SnapshotManagerTest.php
@@ -18,6 +18,77 @@ use Sonata\PageBundle\Entity\SnapshotManager;
  */
 class SnapshotManagerTest extends \PHPUnit_Framework_TestCase
 {
+    public function testSetTemplates()
+    {
+        $manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
+            // we need to set at least one method, which does not need to exist!
+            // otherwise all methods will be mocked and could not be used!
+            // we need the real 'setTemplates' method here!
+            ->setMethods(array(
+                'fooBar',
+            ))
+            ->disableOriginalConstructor()
+            ->getMock()
+        ;
+
+        $this->assertEquals(array(), $manager->getTemplates());
+
+        $manager->setTemplates(array('foo' => 'bar'));
+
+        $this->assertEquals(array('foo' => 'bar'), $manager->getTemplates());
+    }
+
+    public function testGetTemplates()
+    {
+        $manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
+            ->setMethods(array(
+                'setTemplates',
+            ))
+            ->disableOriginalConstructor()
+            ->getMock()
+        ;
+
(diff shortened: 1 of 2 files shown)


OskarStark:
ping @greg0ire

greg0ire:
This kind of thing happens when there are no tests. Can you write some ?

Tests/Entity/SnapshotManagerTest.php
```diff
@@ -18,6 +18,78 @@
  */
 class SnapshotManagerTest extends \PHPUnit_Framework_TestCase
 {
+    public function testSetTemplates()
```
OskarStark: @greg0ire can you please help me?

`testSetTemplates` is not working 😞

Tests/Entity/SnapshotManagerTest.php
```diff
@@ -18,6 +18,78 @@
...
+    {
+        $manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
+            // this method is called in SonataPageExtension
+            ->setMethods(array(
+                'setTemplates'
```
greg0ire: What does this do ? If this mocks setTemplates, then the logic in `setTemplates` will never be called and the test cannot work, right ?

OskarStark: I updated my PR

@ahmetakbn will explain the details to you

ahmetakbn: In this case it works:
```php
$manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
    ->setMethods(array(
    'foo',
    ))
    ->disableOriginalConstructor()
    ->getMock()
;
```
In this case does not work,  because we mock the method:
```php
$manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
    ->setMethods(array(
    'setTemplates',
    ))
    ->disableOriginalConstructor()
    ->getMock()
;
```
But why does not work in these two cases?
```php
$manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
    ->disableOriginalConstructor()
    ->getMock()
;


$manager = $this->getMockBuilder('Sonata\PageBundle\Entity\SnapshotManager')
    ->setMethods(array())
    ->disableOriginalConstructor()
    ->getMock()
;
```

Why we need to call the `setMethods` method with any value?

greg0ire: Maybe the default value for methods is everything rather than nothing ?

OskarStark:
This one is `RTM` @greg0ire after travis is green!

Soullivaneuh:
@OskarStark I updated your changelog.

For 99% of cases, a patch changelog should be on the `Fixed` section, telling what is it fixed.

OskarStark:
> @OskarStark I updated your changelog.
>
> For 99% of cases, a patch changelog should be on the Fixed section, telling what is it fixed.

thank you @Soullivaneuh

Issue #97, opened 2012-06-18
bootstrap js


henres:
In the /web/vendor/sonataadmin/bootstrap/js you have an only two file bootstrap.js and bootstrap.min.js. And, sonataPage call a lot of cut files

```
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-tabs.js" type="text/javascript"></script>
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-alerts.js" type="text/javascript"></script>
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-modal.js" type="text/javascript"></script>
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-twipsy.js" type="text/javascript"></script>
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-popover.js" type="text/javascript"></script>
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-scrollspy.js" type="text/javascript"></script>
        <script src="/bundles/sonataadmin/bootstrap/js/bootstrap-dropdown.js" type="text/javascript"></script>
```

the missing blackbar came from this missing js script?

PR #671, opened 2016-06-15
add support for cocur/slugify 2.x

OskarStark:
### Changelog

```markdown
### Added
- Support version 2.x for `cocur/slugify` dependency
```

### Subject

Support `cocur/slugify` 2.x

SonataCoreBundle already supports it: https://github.com/sonata-project/SonataCoreBundle/blob/3.x/composer.json#L27

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "symfony-cmf/routing-bundle": "^1.1",
         "sonata-project/notification-bundle": "^3.0",
         "sonata-project/datagrid-bundle": "^2.2",
-        "cocur/slugify": "^1.0"
+        "cocur/slugify": "^1.0 || ^2.0"
     },
     "require-dev": {
         "sllh/php-cs-fixer-styleci-bridge": "^2.0",


OskarStark:
Lets try it, but i think this isn't working, until we drop Support for PHP 5.3...

From the Changelog:
> #97 Set minimum PHP version to 5.5.9 (by florianeckerstorfer)

OskarStark:
i could resubmit this against `master` and adjust the minimum php version

Soullivaneuh:
> Lets try it, but i think this isn't working, until we drop Support for PHP 5.3...

Then composer will fallback to `cocur/slugify` 1.x. What is the problem if we support both?

Soullivaneuh:
Could you check and confirm that code using `cocur/slugify` is tested? If not, could you please fix this?

OskarStark:
> Then composer will fallback to cocur/slugify 1.x. What is the problem if we support both?

you are right, i am already a bit tired 😄

OskarStark:
> Could you check and confirm that code using cocur/slugify is tested? If not, could you please fix this?

Will do this

OskarStark:
We have tests: https://github.com/sonata-project/SonataPageBundle/blob/3.x/Tests/Model/PageTest.php#L18

Soullivaneuh:
Great, RTM so.

Soullivaneuh:
https://github.com/sonata-project/SonataPageBundle/blob/3.0.1/SonataPageBundle.php#L65

Would be great to have a deprecation warning here.

OskarStark:
> Would be great to have a deprecation warning here.

done in #673

PR #676, opened 2016-06-16
Call `getRelativePath()` on `null`

sergekv:
I am targetting this branch, because…

Closes #
Fixes #

## Changelog

```markdown
### Added
- Added some `Class::newMethod` to do great stuff

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

## Subject

To avoid PHP fatal error I added the check.
```
PHP Fatal error:  Call to a member function getRelativePath() on null in [...]/vendor/sonata-project/page-bundle/Twig/Extension/PageExtension.php on line 281
```

PR #677, opened 2016-06-16
Call `getRelativePath()` on `null`

sergekv:
I am targetting this branch, because I want to propose a patch.

Fixes #676 

## Changelog

```markdown
### Fixed
- Fixed PHP Fatal error:  Call to a member function getRelativePath() on null
```

## Subject

To avoid PHP fatal error I added the check.
`PageExtension` is checking if the `pathInfo` attribute is defined. If not, it's retrieved from the relative path of the site selector. Since my route excluded from `sonata_page`, it couldn't be retrieved.
```
PHP Fatal error:  Call to a member function getRelativePath() on null in [...]/vendor/sonata-project/page-bundle/Twig/Extension/PageExtension.php on line 281
```

### DIFF
--- a/Twig/Extension/PageExtension.php
+++ b/Twig/Extension/PageExtension.php
@@ -275,13 +275,15 @@ class PageExtension extends \Twig_Extension implements \Twig_Extension_InitRunti
      */
     public function controller($controller, $attributes = array(), $query = array())
     {
-        $globals = $this->environment->getGlobals();
-
         if (!isset($attributes['pathInfo'])) {
-            $sitePath = (null !== $site = $this->siteSelector->retrieve()) ? $site->getRelativePath() : '';
-            $currentPathInfo = $globals['app']->getRequest()->getPathInfo();
+            $site = $this->siteSelector->retrieve();
+            if ($site) {
+                $sitePath = $site->getRelativePath();
+                $globals = $this->environment->getGlobals();
+                $currentPathInfo = $globals['app']->getRequest()->getPathInfo();
 
-            $attributes['pathInfo'] = $sitePath.$currentPathInfo;
+                $attributes['pathInfo'] = $sitePath.$currentPathInfo;
+            }
         }
 
         return $this->httpKernelExtension->controller($controller, $attributes, $query);
--- a/Twig/Extension/PageExtension.php
+++ b/Twig/Extension/PageExtension.php
@@ -278,7 +278,7 @@ class PageExtension extends \Twig_Extension implements \Twig_Extension_InitRunti
         $globals = $this->environment->getGlobals();
 
         if (!isset($attributes['pathInfo'])) {
-            $sitePath = $this->siteSelector->retrieve()->getRelativePath();
+            $sitePath = (null !== $site = $this->siteSelector->retrieve()) ? $site->getRelativePath() : '';
             $currentPathInfo = $globals['app']->getRequest()->getPathInfo();
 
             $attributes['pathInfo'] = $sitePath.$currentPathInfo;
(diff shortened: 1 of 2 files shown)


greg0ire:
You could have edited an reopened your other PR, you know ;)

Anyway, this is better :)

sergekv:
OK, thanks :)

greg0ire:
Can you please add a unit test for that ?

sergekv:
Hmm... What do you want me to test?

greg0ire:
If it is easy, try to write a unit test for the line you changed : show that if works with null and with a non-null value.

Soullivaneuh:
> To avoid PHP fatal error I added the check.

Could you please explain in which case this happen and why this should be handled like this (e.g. returning `''`).

sergekv:
Ok.

sergekv:
@Soullivaneuh you mean update the comment?

Soullivaneuh:
> @Soullivaneuh you mean update the comment?

I'm mean start to explain here first.

But yeah this could be described on comment too. :+1:

sergekv:
`PageExtension` is checking if the `pathInfo` attribute is defined. If not, it's retrieved from the relative path of the site selector. Since my route excluded from `sonata_page`, it couldn't be retrieved.

sergekv:
Hmm, I think the behaviour should be different. Will update the PR and will add a test.

Twig/Extension/PageExtension.php
```diff
@@ -278,7 +278,7 @@ public function controller($controller, $attributes = array(), $query = array())
         $globals = $this->environment->getGlobals();
 
         if (!isset($attributes['pathInfo'])) {
-            $sitePath = $this->siteSelector->retrieve()->getRelativePath();
+            $sitePath = (null !== $site = $this->siteSelector->retrieve()) ? $site->getRelativePath() : '';
```
core23: WTF?

Please split this to multiple lines. No one can read this code.

sergekv:
Guys, why coverage is decreased, although I added the tests? Is there any docs that explains how the coverage is calculating?

greg0ire:
Not sure it gives accurate results when travis fails.

greg0ire:
Fix the tests in travis and try again please.

greg0ire:
See? All green.

sergekv:
@greg0ire yeah, you're right

OskarStark:
Good job @sergekv 👍

greg0ire:
This needs a squash merge

greg0ire:
>  PageExtension is checking if the pathInfo attribute is defined. If not, it's retrieved from the relative path of the site selector. Since my route excluded from sonata_page, it couldn't be retrieved.

What route are you talking about ? Explain like we don't know what the bundle does (which is practically).

@sonata-project/contributors : anyone who merges this must fully understand the problem, and explain it in the commit message, to avoid this kind of situation : https://github.com/sonata-project/SonataAdminBundle/pull/3965

sergekv:
@OskarStark thanks :)
@greg0ire OK, I'll try to explain from begin to end briefly.
I rendered the controller inside the template like this `{{ render(controller('FooBundle:Bar:action')) }}`. The route isn't in sonata page bundle routes scope (sites). The `pathInfo` isn't set. So the `PageExtension` tried to fetch the site to get the `pathInfo` but it couldn't and this is caused the fatal error. So basically it breaks default `controller` twig extension which doesn't require `pathInfo` and site. That's it.

greg0ire:
Oh so `controller()` is a twig function defined both in page bundle and symfony, and the symfony one is being overwritten, am I right ?

sergekv:
Right.
https://symfony.com/doc/current/book/templating.html#embedding-controllers

greg0ire:
Proposed commit message

```
check site is found before trying to set pathinfo

When the route used in the controller() twig function defined in the bundle is not managed by the bundle, no site can be found. In that case, don't do anything and forward the call to the native controller() twig function.
```

Did I get that right @sergekv ?

sergekv:
@greg0ire yes, I think this is correct description.

greg0ire:
Thanks @sergekv

sergekv:
You're welcome :)

PR #680, opened 2016-06-20
Configure admin template only if AdminBundle is present

Wharenn:
I am targeting this branch, because this is a non-BC fix

## Changelog

```markdown
### Fixed
- Fixed The service definition "sonata.page.admin.page" does not exist.

```

## Subject

Custom templates for admin are trying to access the ``sonata.page.admin.page`` service in all cases, but this service is only in the container if AdminBundle is loaded in the app Kernel. 

Inn this PR, admin templates configuration is done only if the AdminBundle is present.


composer.json
```diff
@@ -63,7 +63,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.x-dev"
+            "dev-master": "4.x-dev"
```
Soullivaneuh: This must not be changed.

Soullivaneuh:
> I am targeting this branch, because this is a non-BC fix

Yeah but this branch is the branch for BC fixes and features. :-)

BTW, could you make it BC?

DependencyInjection/SonataPageExtension.php
```diff
@@ -54,6 +54,8 @@ public function load(array $configs, ContainerBuilder $container)
             if (!$config['direct_publication']) {
                 $container->removeDefinition('sonata.page.admin.extension.snapshot');
             }
+
+            $this->configureTemplatesAdmin($container, $config);
```
Soullivaneuh: I'm not really sure this is BC break.

If the admin-bundle is not loaded, who the f*ck will use admin templates? :-)

cc @sonata-project/contributors

greg0ire: People doing that deserve to get the BC break in a new minor.

Wharenn:
There was a branch mismatch in my side, closed for #681.
Thank you

Soullivaneuh:
Don't close, this is the right branch!

Soullivaneuh:
Just remove not wanted commit and we will be good. :+1:

UPGRADE-4.0.md
```diff
@@ -0,0 +1,10 @@
+UPGRADE FROM 3.x to 4.0
```
Soullivaneuh: Please remove this file.

Soullivaneuh:
#681

PR #681, opened 2016-06-20
Configure admin template only if AdminBundle is present

Wharenn:
I am targeting this branch, because this is a non-BC fix

## Changelog

```markdown
### Fixed
- Service definition `sonata.page.admin.page` does not exist

```

## Subject

Custom templates for admin are trying to access the ``sonata.page.admin.page`` service in all cases, but this service is only in the container if AdminBundle is loaded in the AppKernel. 

In this PR, admin templates configuration is done only if the AdminBundle is present.


Soullivaneuh:
RTM for me. :+1:

OskarStark:
LGTM 👍 

Thank you @Wharenn 🎉

PR #691, opened 2016-07-13
Condition for enabled field on selecting enabled snapshots

Vesax:
I am targetting this branch, because this is a BC fix

## Changelog

```markdown
### Fixed
- disabled snapshots are available for users
```

## Subject

`Entity\SnapshotManager::findEnableSnapshot` returning a snapshot even if it is disabled.

So, users can see disabled state of a page. I just added obvious condition for it.

### DIFF
--- a/Entity/SnapshotManager.php
+++ b/Entity/SnapshotManager.php
@@ -107,7 +107,9 @@ class SnapshotManager extends BaseEntityManager implements SnapshotManagerInterf
 
         $query = $this->getRepository()
             ->createQueryBuilder('s')
-            ->andWhere('s.publicationDateStart <= :publicationDateStart AND ( s.publicationDateEnd IS NULL OR s.publicationDateEnd >= :publicationDateEnd )');
+            ->andWhere('s.publicationDateStart <= :publicationDateStart AND ( s.publicationDateEnd IS NULL OR s.publicationDateEnd >= :publicationDateEnd )')
+            ->andWhere('s.enabled = 1')
+        ;
 
         if (isset($criteria['site'])) {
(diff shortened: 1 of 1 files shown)


OskarStark:
LGTM 👍

OskarStark:
could enabled be `null`, if yes we need to cover this case too

Vesax:
@OskarStark how should we cover this case? We just select enabled `true` and ignore `false` and `null`

OskarStark:
> @OskarStark how should we cover this case? We just select enabled true and ignore false and null

you are right i mied up things 👍  was thinking from the opposite POV

Entity/SnapshotManager.php
```diff
@@ -107,7 +107,9 @@ public function findEnableSnapshot(array $criteria)
 
         $query = $this->getRepository()
             ->createQueryBuilder('s')
-            ->andWhere('s.publicationDateStart <= :publicationDateStart AND ( s.publicationDateEnd IS NULL OR s.publicationDateEnd >= :publicationDateEnd )');
+            ->andWhere('s.publicationDateStart <= :publicationDateStart AND ( s.publicationDateEnd IS NULL OR s.publicationDateEnd >= :publicationDateEnd )')
+            ->andWhere('s.enabled = 1')
```
core23: Shouldn't this be true instead of 1?

Vesax: @core23 they are equal in this case

core23: IMHO `true` gives you a better understaning when reading the query, because `enabled` is a boolean field and no integer.

core23:
Thank you @Vesax

PR #552, opened 2015-10-23
Remove SnapshotPageProxy hard coded dependancy

FabienD:
According what I've explained in the issue #551, I've built this PR in order to make SnapshotPageProxy customisable in an easy way (I think).

PR #697, opened 2016-07-22
Added SnapshotPageFactory

### LINKED ISSUE #551: Extend Page entity & hard coded SnapshotPageProxy class
I need to add a property called "body css" to my entity "Page".

For this, I started to extend the entity in **"Application\Sonata\PageBundle\Entity"**, far so good. Then, I extended the **"Transformer"** class, so the new property value will be serialized in the snapshot record field ("content").

In the front, if it works when you are logged in admin (the page is retrieve by the **CmsPageManager**), it does not work as an anonymous user (Page is retrieve by the **CmsSnapshotManager**). 

To get everything works, we need to extend **SnapshotPageProxy** class to add the getter and setter for the new property "bodyCss". For that, you need to modify other classes where SnapshotProxy is hard coded. 

Why not use the DI to inject **SnapshotPageProxy** Class rather than hardcode it ? 

Did I miss something ?

Thanks

core23:
I am targetting this branch, because this is BC.

Rebase of #552
Fixes #551

## Changelog

```markdown
### Added
- Added new `SnapshotPageFactory`
```

## Subject

This PR makes `SnapshotPageProxy` customizable in an easy way.


OskarStark:
LGTM 👍

core23:
Ping @OskarStark @greg0ire

greg0ire:
No unit test changed to the `@legacy` group, and I can't see your new deprecation notice in the output, which means there are no unit tests. Can you write some to cover the new code you added?

greg0ire:
Side note, there are many deprecation notices that should be either fixed or hidden by te `@legacy` tag : https://travis-ci.org/sonata-project/SonataPageBundle/jobs/153896793#L1208

core23:
Fixed one deprecation issue: https://github.com/sonata-project/SonataPageBundle/pull/710

core23:
Added some tests

greg0ire:
Thanks @core23 !

Entity/SnapshotManager.php
```diff
@@ -147,6 +168,8 @@ public function findEnableSnapshot(array $criteria)
      * @param string $routeName
      *
      * @return PageInterface|false
+     *
+     * @deprecated since 3.x, to be removed in 4.0
```
Soullivaneuh: Here too, this method is deprecated but nothing on the upgrade doc.

Please @sonata-project/contributors do not merge PR without any note on the upgrade note.

Soullivaneuh: And a `trigger_error` is missing IMO.

CmsManager/CmsSnapshotManager.php
```diff
@@ -152,7 +152,12 @@ protected function getPageBy(SiteInterface $site = null, $fieldName, $value)
                 throw new PageNotFoundException();
             }
 
-            $page = new SnapshotPageProxy($this->snapshotManager, $this->transformer, $snapshot);
+            // NEXT_MJAOR: Remove this check
```
greg0ire: Should be MAJOR

PR #699, opened 2016-08-01
Fixed redeclared interface error

mremi:
I am targetting this branch, because this is BC.

## Changelog

```markdown
### Fixed
- Warmup the cache from the CLI brings an error in HTTP
```

## Subject

When the cache is warmup from the CLI, you get this fatal error in front:
```
PHP message: PHP Fatal error:  Cannot declare interface Sonata\PageBundle\Request\SiteRequestInterface, because the name is already in use in /var/www/my-project/var/cache/prod/classes.php on line 12048
```


greg0ire:
Your email address is not associated with your github account, you might want to fix that.

Tests/DependencyInjection/SonataPageExtensionTest.php
```diff
@@ -0,0 +1,92 @@
...
+            'Sonata\\PageBundle\\Page\\TemplateManager',
+            'Sonata\\PageBundle\\Page\\TemplateManagerInterface',
+            'Sonata\\PageBundle\\Request\\SiteRequestContext',
+            'Sonata\\PageBundle\\Route\\CmsPageRouter',
+            'Sonata\\PageBundle\\Site\\BaseSiteSelector',
+            'Sonata\\PageBundle\\Site\\HostPathSiteSelector',
+            'Sonata\\PageBundle\\Site\\HostSiteSelector',
+            'Sonata\\PageBundle\\Site\\SiteSelectorInterface',
+            'Sonata\\PageBundle\\Twig\\Extension\\PageExtension',
+            'Sonata\\PageBundle\\Twig\\GlobalVariables',
```
greg0ire: Doesn't that feel a bit weird?

mremi:
@greg0ire oops, you're right about my email, fixed ;)

Tests/DependencyInjection/SonataPageExtensionTest.php
mremi: What?

greg0ire: The fact that whenever you add a class / remove a class to compile, you also have to change the test? You should be testing the program, not the data in it…

greg0ire: Plus here you're testing a third-party API call without checking that it actually does what you expect it to do. Food for thought : http://elnur.pro/testing-repositories/

greg0ire: IMO you should be testing that one or two classes are in the argument, but not all of them.

mremi: Still better than no test as before :) what do you expect? Assert that `SiteRequest` and `SiteRequetInterface` are not in the compiled classes?

greg0ire: Think about this test. How will it fail : 

1. If someone adds a new class to compile
2. If someone removes a class from compile
3. If somehow, the code is refactored and the call is made, but with a completely different argument

IMO, the only interesting case when you want it to fail because here is the reaction of the developer when this test fails : 

1. Duh, of course if fails, I added a class to compile
1. Duh, of course if fails, I removed a class from compile
1. Uh oh, this is very interesting, my refactoring is probably not good, the argument does not look like what it was before.

To have the test fail only for 3. , you should assert that the argument looks like a big array of classes (but really, what classes are in there does not matter).

greg0ire: To clarify, if you have this method 

```php
public function getName()
{
   return 'toto';
}
```

It's probably better to test that it returns a string, rather than testing it returns `toto`, do you see why?

greg0ire: Not sure how to implement that though : maybe use [assertArraySubset](https://github.com/sebastianbergmann/phpunit/blob/064c8a53538918260fa0a5693c8d82a09f9b5498/src/Framework/Assert.php#L63) instead of equalTo?

mremi: Yep, no problem with this of course. The fact is these 2 lines break the website, so I really think we need to make sure they are absent from the compiled classes. I can update the test in this way.

greg0ire: Oh ok, I did not catch that. That's indeed an assertion that you want to make clearer.

greg0ire: Maybe you can use this : https://github.com/sebastianbergmann/phpunit/blob/064c8a53538918260fa0a5693c8d82a09f9b5498/src/Framework/Assert.php#L197

mremi: That's it :) Ok, I'll update it

mremi:
@greg0ire done

greg0ire:
Much better! LGTM!

greg0ire:
By the way, do you know why it fails for these class and this interface in particular?

greg0ire:
Similar issue : https://github.com/symfony/symfony/issues/18769

mremi:
I think it was introduced with the usage of the `SiteRequestFactory` in the `app.php` which creates a new `SiteRequest` that implements `SiteRequestInterface` + usage of `$kernel->loadClassCache()` within the `app.php`.

When the cache does not exist, the [ClassCollectionLoader](https://github.com/symfony/class-loader/blob/v2.8.9/ClassCollectionLoader.php#L105)  checks the declared classes and interfaces to add only those not already known by PHP. So from the CLI, `SiteRequest` and `SiteRequestInterface` are added to the cache because not used in the warmup process. And so clash when you load the cache from HTTP.

mremi:
What are the next steps to merge this? :) Need another 👍 from someone else?

greg0ire:
I'm crafting a good commit message right now, but indeed, it would be great to have another review, ping @sonata-project/contributors 

Suggested commit message, to be used by either you or the merger : 

```
Avoid compiling some classes

When using SiteRequestFactory in the front controller,
there might be a cache generation problem : SiteRequest and SiteRequestInterface are loaded,
and then the cache generated in CLI, which may contain the very same classes, is loaded,
leading to a conflict.
```

greg0ire:
Also, can't you load the cache **before** you use the request factory ?

mremi:
The only way I found to do this is to boot the kernel (and so load the cache) before create the request:

```php
<?php

use Sonata\PageBundle\Request\RequestFactory;

/**
 * @var Composer\Autoload\ClassLoader
 */
$loader = require __DIR__.'/../app/autoload.php';
include_once __DIR__.'/../var/bootstrap.php.cache';

$kernel = new AppKernel('prod', false);
$kernel->loadClassCache();
//$kernel = new AppCache($kernel);

$kernel->boot();

// When using the HttpCache, you need to call the method in your front controller instead of relying on the configuration parameter
//Request::enableHttpMethodParameterOverride();
$request = RequestFactory::createFromGlobals('host_with_path_by_locale');
$response = $kernel->handle($request);
$response->send();
$kernel->terminate($request, $response);
```

mremi:
Let me know if you want me to update the commit message

greg0ire:
> Let me know if you want me to update the commit message

Yes please, this will be one less thing to do for the merger.

mremi:
done

greg0ire:
Perfect! Let's wait for the final review and merge now.

OskarStark:
Thank you @mremi 👍

mremi:
Can you tell me when a new release will be created plz?

Soullivaneuh:
@mremi https://github.com/sonata-project/SonataPageBundle/releases/tag/3.1.0

mremi:
Great, thanks guys!

PR #704, opened 2016-08-08
cascade persist for block parent

Vesax:
I am targetting this branch, because this is a non-BC fix

## Changelog

``` markdown
### Fixed
- missing cascade persistence for parent block.
```
## Subject

For now, we can't create block with its new parent, besause cascade persistence not enabled.
But it is enabled for page's parent.
(For example, it is very important feature for moving pages from another cms solution).

### DIFF
--- a/DependencyInjection/SonataPageExtension.php
+++ b/DependencyInjection/SonataPageExtension.php
@@ -261,7 +261,7 @@ class SonataPageExtension extends Extension
             'fieldName' => 'parent',
             'targetEntity' => $config['class']['block'],
             'cascade' => array(
-                'persist'
+                'persist',
             ),
             'mappedBy' => null,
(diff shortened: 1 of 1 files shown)


Vesax:
@OskarStark What do you think about it?

OskarStark:
to be honest, I#m not that deep in this topic

maybe @core23 can help!

core23:
Not sure, if we need this. Can you can name where the cascade persist is needed in this bundle?

When creating a new block, the parent block is already persisted. We need the cascade persist only when creating a cascaded tree, which is not the case in this bundle.

OskarStark:
Ping @Vesax

core23:
Please give an example why we need this fix or close this PR @Vesax

PR #708, opened 2016-08-18
make sure the scope return to the initial state while running the com…

rande:
I am targetting this branch, because it is the stable branch.

## Changelog

```markdown
### Fixed
- make sure the scope in the container is clean to avoid failing test while unit testing the command
```

## Subject

make sure the scope in the container is clean to avoid failing test while unit testing the command


rande:
We can also merge this PR ;)

OskarStark:
Thank you @rande

PR #711, opened 2016-08-25
Fix Unique Url Validator for Dynamic pages

qsomazzi:
## Changelog

```markdown
### Changed
- UniqueUrl validation isn't checked for Dynamic pages anymore
```
## Subject

The goal of this is to allow saving Dynamic pages even if the Url already exists.
On Dynamic pages, if we have parameters in the URL, we can have multiples pages with the same url.

### DIFF
--- a/Validator/UniqueUrlValidator.php
+++ b/Validator/UniqueUrlValidator.php
@@ -49,8 +49,8 @@ class UniqueUrlValidator extends ConstraintValidator
             return;
         }
 
-        // do not validated error page
-        if ($value->isError()) {
+        // do not validate error or dynamic pages
+        if ($value->isError() || $value->isDynamic()) {
             return;
         }
 
--- a/Tests/Validator/UniqueUrlValidatorTest.php
+++ b/Tests/Validator/UniqueUrlValidatorTest.php
@@ -64,4 +64,25 @@ class UniqueUrlValidatorTest extends \PHPUnit_Framework_TestCase
 
         $validator->validate($page, new UniqueUrl());
     }
+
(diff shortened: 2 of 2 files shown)


Validator/UniqueUrlValidator.php
```diff
@@ -49,8 +49,8 @@ public function validate($value, Constraint $constraint)
             return;
         }
 
-        // do not validated error page
-        if ($value->isError()) {
+        // do not validated error and dynamic pages
```
greg0ire: "do not validate error or dynamic pages"

greg0ire:
>  On Dynamic pages, if we have parameters in the URL, we can have multiples pages with the same url.

Can you provide an example?

qsomazzi:
In my current project I have urls like `/{type}/{subtype}/{id}`
I need to have one SonataPage for each existing association (type/subtype) and each type and subtype are translated. 
Finally I got X pages with the same url (but not the same requirements)

rande:
@greg0ire This is a bug, unique path should only be for cms page, not dynamic page which are created by the system not the user.

Validator/UniqueUrlValidator.php
```diff
@@ -49,8 +49,8 @@ public function validate($value, Constraint $constraint)
             return;
         }
 
-        // do not validated error page
-        if ($value->isError()) {
+        // do not validated error or dynamic pages
```
greg0ire: (Maybe) not your fault but "validated" => "validate"

qsomazzi:
@greg0ire @rande all green to merge ?

PR #720, opened 2016-09-15
Save custom options on redirect

Vesax:
I am targetting this branch, because this is a non-BC fix

## Changelog

```markdown

### Fixed
 - losing custom query on redirect
```

## Subject

Currently, If I am passing some custom query option, like here
`{{ admin.generateUrl('create', {parent_id: page.id}) }}`
PageAdminController does redirect with `siteId` but without my `parent_id` option.
In my case I am avoiding this by adding siteId manually 
`{{ admin.generateUrl('create', {siteId: page.site.id, parent_id: page.id}) }}`
And this is good solution, but anyway, I think we should save custom options.

### DIFF
--- a/Controller/PageAdminController.php
+++ b/Controller/PageAdminController.php
@@ -117,12 +117,14 @@ class PageAdminController extends Controller
             $sites = $this->get('sonata.page.manager.site')->findBy(array());
 
             if (count($sites) == 1) {
-                return $this->redirect($this->admin->generateUrl('create', array(
+                $options = array_merge(array(
                     'siteId' => $sites[0]->getId(),
                     'uniqid' => $this->admin->getUniqid(),
-                )));
-            }
+                ), $request->query->all());
 
+                return $this->redirect($this->admin->generateUrl('create', $options));
+            }
+            
             try {
                 $current = $this->get('sonata.page.site.selector')->retrieve();
             } catch (\RuntimeException $e) {
--- a/Controller/PageAdminController.php
+++ b/Controller/PageAdminController.php
@@ -124,7 +124,7 @@ class PageAdminController extends Controller
 
                 return $this->redirect($this->admin->generateUrl('create', $options));
             }
-            
+
             try {
                 $current = $this->get('sonata.page.site.selector')->retrieve();
             } catch (\RuntimeException $e) {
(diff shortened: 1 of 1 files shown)


Controller/PageAdminController.php
```diff
@@ -117,12 +117,14 @@ public function createAction(Request $request = null)
             $sites = $this->get('sonata.page.manager.site')->findBy(array());
 
             if (count($sites) == 1) {
-                return $this->redirect($this->admin->generateUrl('create', array(
+                $options = array_merge(array(
```
greg0ire: Maybe you can use the `+` operator instead of `array_merge` in this case ?

Controller/PageAdminController.php
```diff
@@ -117,12 +117,14 @@ public function createAction(Request $request = null)
...
                     'siteId' => $sites[0]->getId(),
                     'uniqid' => $this->admin->getUniqid(),
-                )));
-            }
+                ), $request->query->all());
 
+                return $this->redirect($this->admin->generateUrl('create', $options));
```
greg0ire: If you do choose to use the `+` operator, maybe revert back to the version without the `$options` one-time variable?

Controller/PageAdminController.php
Vesax: I can, you are right

greg0ire:
Do we have any tests for Controllers? I doubt it, but if we have, please modify them to tests your changes. If it is not the case, make sure you test your changes on a project of yours before we merge.

Vesax:
We don't have tests for this Controller. Changes are tested, I didn't find any problem.

Controller/PageAdminController.php
```diff
@@ -120,7 +120,7 @@ public function createAction(Request $request = null)
                 return $this->redirect($this->admin->generateUrl('create', array(
                     'siteId' => $sites[0]->getId(),
                     'uniqid' => $this->admin->getUniqid(),
-                )));
+                ) + $request->query->all()));
```
core23: @greg0ire why should we use `+` instead of `array_merge`?

greg0ire: Because it is way more concise, that's all. Plus you don't have to go back and forth to add parenthesis.

greg0ire:
Thanks @Vesax !

PR #735, opened 2016-10-18
Add RelativePath in View page link

jmontoyaa:
I am targetting this branch, because there's a bug in "View page" link that doesnt take into account the relative path of the sub site.

## Changelog

``` markdown
### Added
- Add relative path in "View page link"
```
## Subject

When creating a subsite, the relative path is not taking into account.

### DIFF
--- a/Admin/PageAdmin.php
+++ b/Admin/PageAdmin.php
@@ -449,15 +449,15 @@ class PageAdmin extends AbstractAdmin
 
         if (!$this->getSubject()->isHybrid() && !$this->getSubject()->isInternal()) {
             try {
-                $url = $this->getSubject()->getUrl();
-                $relativePath = $page->getSite()->getRelativePath();
-                if (!empty($relativePath)) {
-                    $url = $relativePath.$url;
+                $path = $this->getSubject()->getUrl();
+                $siteRelativePath = $page->getSite()->getRelativePath();
+                if (!empty($siteRelativePath)) {
+                    $path = $siteRelativePath.$path;
                 }
                 
                 $menu->addChild(
                     'view_page',
-                    array('uri' => $this->getRouteGenerator()->generate('page_slug', array('path' => $url)))
+                    array('uri' => $this->getRouteGenerator()->generate('page_slug', array('path' => $path)))
                 );
             } catch (\Exception $e) {
                 // avoid crashing the admin if the route is not setup correctly
--- a/Admin/PageAdmin.php
+++ b/Admin/PageAdmin.php
@@ -449,8 +449,15 @@ class PageAdmin extends AbstractAdmin
 
         if (!$this->getSubject()->isHybrid() && !$this->getSubject()->isInternal()) {
             try {
-                $menu->addChild('view_page',
-                    array('uri' => $this->getRouteGenerator()->generate('page_slug', array('path' => $this->getSubject()->getUrl())))
+                $url = $this->getSubject()->getUrl();
+                $relativePath = $page->getSite()->getRelativePath();
+                if (!empty($relativePath)) {
+                    $url = $relativePath.$url;
+                }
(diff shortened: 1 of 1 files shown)


Admin/PageAdmin.php
```diff
@@ -449,8 +449,15 @@ protected function configureSideMenu(MenuItemInterface $menu, $action, AdminInte
 
         if (!$this->getSubject()->isHybrid() && !$this->getSubject()->isInternal()) {
             try {
-                $menu->addChild('view_page',
-                    array('uri' => $this->getRouteGenerator()->generate('page_slug', array('path' => $this->getSubject()->getUrl())))
+                $url = $this->getSubject()->getUrl();
+                $relativePath = $page->getSite()->getRelativePath();
+                if (!empty($relativePath)) {
+                    $url = $relativePath.$url;
```
greg0ire: That won't work if `getUrl()` returns an absolute url. If it can't, `getUrl()` is badly named, but you can still name your variable correctly (using `$path` instead of `$url`). Likewise `$relativePath` should probably be named `$siteRelativePath`, that way will end up here with `$path = $siteRelativePath . $path`.

jmontoyaa: Thanks. I added those changes.

greg0ire:
Do you need help to fix StyleCI?

jmontoyaa:
Not sure if that will fix it.  :sweat_smile:

greg0ire:
You can use `git commit --amend` in that kind of situation ;)

greg0ire:
> Not sure if that will fix it. :sweat_smile:

It did, congrats!

greg0ire:
Hum there are not test at all for the `PageAdmin` class at the moment :confounded:

Admin/PageAdmin.php
```diff
@@ -449,8 +449,15 @@ protected function configureSideMenu(MenuItemInterface $menu, $action, AdminInte
...
             try {
-                $menu->addChild('view_page',
-                    array('uri' => $this->getRouteGenerator()->generate('page_slug', array('path' => $this->getSubject()->getUrl())))
+                $path = $this->getSubject()->getUrl();
+                $siteRelativePath = $page->getSite()->getRelativePath();
+                if (!empty($siteRelativePath)) {
+                    $path = $siteRelativePath.$path;
+                }
+
+                $menu->addChild(
```
core23: Would prefer this style for better readability:

```               
               $menu->addChild('view_page', array(
                    'uri' => $this->getRouteGenerator()->generate('page_slug', array (
						'path' => $path
					))
                ));
```

haivala:
not merged in 3.3.0 :(

greg0ire:
@haivala : would you like to carry this PR?

haivala:
@greg0ire Just to clear things up I have to say that I don't have time for this..

greg0ire:
Thanks for clearing that up… anyone else?

Aylarius:
Hi! 
I'm having the same issue and just tried the solution. However I had an error "undefined variable page", $page was not defined anywhere in the file PageAdmin.php. 
I used $this->getSubject() instead of $page and it seems to be working just fine.

greg0ire:
@Aylarius would you like to carry this PR?

Aylarius:
@greg0ire I'm not sure I will know how. I'm taking a look at the contribution guide but does it need tests?

greg0ire:
Yes it will, but they won't be that hard to write:

```php
$admin = new PageAdmin();
// call some setters if necessary (there will probably be the need for a router)
$menu = $admin->buildTabMenu();
// write some tests on $menu
```

Aylarius:
Ok I'll try this week-end!

greg0ire:
I answered you on slack @Aylarius

PR #737, opened 2016-10-21
Improved extensibility of controller actions

core23:
I am targetting this branch, because this is a BC break.

## Changelog

``` markdown
### Added
- Added `$request` parameter to all controller actions
```
## Subject

Improved extensibility of controller actions by injection the `$request` parameter by default.


OskarStark:
Imo we should throw a deprecation in 3.x if the first parameter is not a request object and add an UPGRADE note

PR #767, opened 2017-01-15
Revert: Use the page name if the page has no SEO title (#646)

core23:
I am targetting this branch, because this is a bugfix.

Reverts #646 partially

## Changelog

```markdown
### Fixed
 - the page title won't get overwritten anymore
```

## Subject

Reverts my PR from last year, because any existing title was overwritten. This PR fixes this behavior, so the title will be set if there isn't any.

### DIFF
--- a/Page/Service/DefaultPageService.php
+++ b/Page/Service/DefaultPageService.php
@@ -73,7 +73,9 @@ class DefaultPageService extends BasePageService
             return;
         }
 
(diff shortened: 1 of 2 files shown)


Vesax:
@core23 @OskarStark actually there are always a title if default one is defined in the config, so after this patch page titles just don't appear at all. I have just faced this issue when realized that all my pages have the same title - the default one.

PR #777, opened 2017-01-27
Fix wrong admin_sonata_page_page_create page create route

### LINKED ISSUE #493: admin_sonata_page_page_create path not found
locaton: SonataPageBundle:Page:create.html.twig

the path sould be: admin_app_page_page_create

gremo:
I am targetting this branch, because this is a bug filx (very old)

Closes #493.

## Changelog

```markdown
### Changed
- Route name from `admin_sonata_page_page_create` to method `sonata_admin.url()`.
```

### DIFF
--- a/Resources/views/Page/create.html.twig
+++ b/Resources/views/Page/create.html.twig
@@ -18,7 +18,9 @@ file that was distributed with this source code.
 
         {% if creatable %}
             <p>
-                <a href="{{ path('admin_sonata_page_page_create', {'url': pathInfo, 'siteId': site.id}) }}">{{ "create_page"|trans({'pathInfo': pathInfo}, 'SonataPageBundle') }}</a>
+                <a href="{{ sonata_admin.url('sonata.page.admin.page', 'create', {'url': pathInfo, 'siteId': site.id}) }}">
+                    {{ "create_page"|trans({'pathInfo': pathInfo}, 'SonataPageBundle') }}
+                </a>
             </p>
         {% else %}
             <p>


Resources/views/Page/create.html.twig
```diff
@@ -18,7 +18,7 @@ file that was distributed with this source code.
 
         {% if creatable %}
             <p>
-                <a href="{{ path('admin_sonata_page_page_create', {'url': pathInfo, 'siteId': site.id}) }}">{{ "create_page"|trans({'pathInfo': pathInfo}, 'SonataPageBundle') }}</a>
+                <a href="{{ path('admin_app_page_page_create', {'url': pathInfo, 'siteId': site.id}) }}">{{ "create_page"|trans({'pathInfo': pathInfo}, 'SonataPageBundle') }}</a>
```
core23: Looks wrong to me. Why do you need the `*_app_*` route?

gremo: I'll look at this ASAP

gremo: Ok, what about this:

`href="{{ sonata_admin.adminPool.instance('sonata.page.admin.page').generateUrl('create') }}"`

gremo: Even better to use `url` function:

`href="{{ sonata_admin.url('sonata.page.admin.page', 'create', {'url': pathInfo, 'siteId': site.id}) }}"`

Resources/views/Page/create.html.twig
```diff
@@ -18,7 +18,7 @@ file that was distributed with this source code.
 
         {% if creatable %}
             <p>
-                <a href="{{ path('admin_sonata_page_page_create', {'url': pathInfo, 'siteId': site.id}) }}">{{ "create_page"|trans({'pathInfo': pathInfo}, 'SonataPageBundle') }}</a>
+                <a href="{{ sonata_admin.url('sonata.page.admin.page', 'create', {'url': pathInfo, 'siteId': site.id}) }}">{{ "create_page"|trans({'pathInfo': pathInfo}, 'SonataPageBundle') }}</a>
```
greg0ire: Please linebreak this.

gremo:
Are we OK with this PR? Can you merge or need more changes?  Thanks!

greg0ire:
I think you should squash the two first commits together (the first one was a mistake, right?)

gremo:
Ok I'll squash all commits (the line break one is useless imho) and force push tomorrow!

greg0ire:
Can you remind me understand what this `sonata_admin` variable is again? Neither Pool nor AbstractAdmin have an `url` method.

gremo:
It's supposed to be https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Twig/GlobalVariables.php#L74 but honestly can't remember how it gets aliased to sonata_admin. I've found It's used sometimes,  do you think this is wrong?

greg0ire:
The aliasing takes place [here](https://github.com/sonata-project/SonataAdminBundle/blob/025e8e1092ae80d305dcbeb817efb17e4ee7a4d2/DependencyInjection/Compiler/GlobalVariablesCompilerPass.php#L29)

> I've found It's used sometimes, do you think this is wrong?

No, I'm just trying to understand the PR, but I can't say I'm a huge fan of this variable

- the class has a plural name
- the purpose of `url` is not documented, so… I don't really get what it does.

Resources/views/Page/create.html.twig
greg0ire: I don't get what makes `admin_app_page_page_create` the right route either… can you please explain like we're five? Bear in mind that I don't use this bundle.

gremo:
Maybe because of https://github.com/symfony/framework-bundle/blob/master/Templating/GlobalVariables.php btw another option is get the admin instance and generare the URL (same as url global method) but if i remember correctely we need to pass the instance to the template.

Either way this needs to be fixed, is very very confusing for a new developer.

gremo:
ping @greg0ire what should I do to get this merged?

greg0ire:
@core23 please review again.

greg0ire:
Thanks @gremo !

PR #781, opened 2017-02-07
Fixing length of "locale" field for doctrine, it should be 7 instead of 6.

TheCelavi:
"Locale" field of entity "Site" has length set as 6, instead of 7, which is not according to current valid ISO standards.

Example, sr_Latn (which is valid locale code) can not be accommodated into field, see: https://en.wikipedia.org/wiki/Template:ISO_639_name_sr-Latn

There is no BC break for this PR.


TheCelavi:
@OskarStark I can not find a template for PR. What merge conflict is reported here, for wich branch, I have made PR on master branch

OskarStark:
And btw. please submit this against `3.x` branch instead of `master`, thank you!

OskarStark:
Why did you close this PR? @TheCelavi

TheCelavi:
@OskarStark I made a mess with branches, new is opened: https://github.com/sonata-project/SonataPageBundle/pull/782

PR #782, opened 2017-02-07
[Bugfix] Property "locale" of entity "Site" should be length of 7

TheCelavi:
I am targetting this branch, because this is BC compatible.

## Changelog
```markdown
### Fixed
- ISO 639 compatibility  Entity Site, property locale, now has length set as 7 instead of 6.
```

## Subject

This fixes compatibility with ISO standards in regards to locale length, see: https://en.wikipedia.org/wiki/Template:ISO_639_name_sr-Latn 

In general, locale can have length of 7 chars.

### DIFF
--- a/Resources/config/doctrine/BaseSite.orm.xml
+++ b/Resources/config/doctrine/BaseSite.orm.xml
@@ -10,7 +10,7 @@
         <field name="isDefault" type="boolean" column="is_default" default="false"/>
         <field name="createdAt" type="datetime" column="created_at"/>
         <field name="updatedAt" type="datetime" column="updated_at"/>
-        <field name="locale" type="string" column="locale" nullable="true" length="6"/>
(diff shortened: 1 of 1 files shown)


OskarStark:
Do we actually have tests for this? Would be cool to extend it with your example

TheCelavi:
@OskarStark I don't think that there is a one... I am quite new in Sonata, so I am not very familiar to its internal structure and what has been done so far.

If you want me to write a test - please provide me with some input, like, where should I add test which tests if ISO 639 is supported.

OskarStark:
There is no test currently, but I think this one is fine.

@sonata-project/contributors please review

greg0ire:
Thanks @TheCelavi . I included interesting part of your PR body in the commit message, please do so yourself next time.

PR #783, opened 2017-02-09
[IMPROVEMENT] Preserve current site filter when working with pages.

TheCelavi:
I am targetting this branch, because it is BC compatible.

## Changelog
```markdown
### Changed
- `Sonata\PageBundle\Admin\PageAdmin`, added method `getPersistentParameters`
```
## Subject

When dealing with multiple sites, it is real pain in a** during operations which deals with non-default site pages. After each navigation step (save, delete) -> list of pages is displayed (tree view) which belongs to default site -> not current working one.

In general, if there are sites A and B, and if I work with site B, I would like to work with site B during my session until I explicitly change to site A and vice versa.

**Conclusion:** this is not a bugfix, this is usability improvement.

### DIFF
--- a/Admin/PageAdmin.php
+++ b/Admin/PageAdmin.php
@@ -221,6 +221,25 @@ class PageAdmin extends AbstractAdmin
     /**
      * {@inheritdoc}
      */
+    public function getPersistentParameters()
+    {
+        $parameters = parent::getPersistentParameters();
+        $key = sprintf('%s.current_site', $this->getCode());
+
+        if ($site = $this->request->get('site', null)) {
+            $this->request->getSession()->set($key, $site);
+        }
+
+        if ($site = $this->request->getSession()->get($key, null)) {
+            $parameters['site'] = $site;
+        }
+
+        return $parameters;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function configureShowFields(ShowMapper $showMapper)
     {
         $showMapper
--- a/Tests/Entity/BlockManagerTest.php
+++ b/Tests/Entity/BlockManagerTest.php
@@ -36,6 +36,7 @@ class BlockManagerTest extends \PHPUnit_Framework_TestCase
             $this->getMockBuilder('Doctrine\ORM\EntityManager')->disableOriginalConstructor()->getMock(),
         ));
 
+        $qb->expects($this->any())->method('getRootAliases')->will($this->returnValue(array()));
         $qb->expects($this->any())->method('select')->will($this->returnValue($qb));
         $qb->expects($this->any())->method('getQuery')->will($this->returnValue($query));
 
--- a/Tests/Entity/PageManagerTest.php
+++ b/Tests/Entity/PageManagerTest.php
@@ -250,6 +250,7 @@ class PageManagerTest extends \PHPUnit_Framework_TestCase
             $this->getMockBuilder('Doctrine\ORM\EntityManager')->disableOriginalConstructor()->getMock(),
         ));
 
+        $qb->expects($this->any())->method('getRootAliases')->will($this->returnValue(array()));
         $qb->expects($this->any())->method('select')->will($this->returnValue($qb));
         $qb->expects($this->any())->method('getQuery')->will($this->returnValue($query));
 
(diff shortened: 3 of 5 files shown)


Admin/PageAdmin.php
```diff
@@ -457,4 +457,15 @@ protected function configureSideMenu(MenuItemInterface $menu, $action, AdminInte
             }
         }
     }
+
+    public function getPersistentParameters()
```
core23: Please add PHPdoc und fix styleCi

TheCelavi:
@core23 Ok, this is better now, I think...

TheCelavi:
@core23 I guess you wanted for this PR to have one commit, so I squashed them...

TheCelavi:
@core23 This is even better solution - current working site is preserved in session.

Why? I have noticed that I am using left main menu to get to the list of pages, not dropdown menu on right (Actions -> Back to list), left main many is always open, while "actions" menu is closed, so I have to click twice instead of once on main menu.

When I click on "Pages" on main menu, I will get back to the site tree, but, default site will be selected, not the current working one.

So I think that it is better usability to store in session current working site and preserve a point of interest until current working site is not explicitly changed (when it comes to administering pages).

core23:
Please don't forget Travis und StyleCI checks

TheCelavi:
@core23 How can I fix styleCi check when no meaningful info is given, see: https://styleci.io/analyses/qJbbEV

What is wrong with code style, do you know so you can give me some pointers?

TheCelavi:
@core23 And on top of that - I can not error out tests as on Travis, when I run tests on my local, all test passes. Is there any pointer how to replicate travis environment? Is there any referent docker container?

core23:
> How can I fix styleCi check when no meaningful info is given, see: https://styleci.io/analyses/qJbbEV 

Just click "Download raw" and apply this patch file `$ git apply patch.txt `

We use StyleCI to auto indent and format our code for better readability.

> And on top of that - I can not error out tests as on Travis, when I run tests on my local, all test passes. Is there any pointer how to replicate travis environment? Is there any referent docker container?

Sorry, look like we have a general problem with our tests. Just be patient. We'll fix this

TheCelavi:
@core23 Ok, patch applied, I was confused because there were no change in code at all -> only position of the method. Nice.

In regards to tests -> ok, I will wait.

Thanks for your help.

TheCelavi:
@core23 Hi, is there any progress in regards to this PR? Thanks!

core23:
Can't merge this unless there are Travis fails.

@greg0ire Can you help here? Can't find the problem right now

greg0ire:
I think I broke this in https://github.com/sonata-project/SonataDatagridBundle/commit/c3bd0b9f300ef5bef84fdea4ea29466e4c41c859

greg0ire:
`getRootAlias` returns null… not sure why.

TheCelavi:
@greg0ire Isn't that method depreciated? Could you use 

    reset($countQuery->getRootAliases()) 

instead? (or any other method to get first element of array)

greg0ire:
That's what is used : `current(getRootAliases())`. The problem is that `getRepository` in `SiteManager` (for instance) is just a mock, which means it does not do things properly. You probably just have to mock `getRootAliases()` somewhere in there.

greg0ire:
add `$qb->expects($this->any())->method('getRootAliases')->will($this->returnValue([]));` line 121 in SiteManager and you will get less warnings. You probably need to do the same in other places.

TheCelavi:
I dont think that `current` is not valid function to call. It does not guarantee that you will get first element of array.

Someone could move internal pointer of array to some other element of array, and it will fail.

greg0ire:
@TheCelavi still better than `[0]`, b/c 0 is not guaranteed to be the first index in the array. But this is not the problem here, see my comment above with the solution.

TheCelavi:
@greg0ire Instead of `current` U can use `reset`, it is more safe approach. http://php.net/manual/en/function.reset.php

> add $qb->expects($this->any())->method('getRootAliases')->will($this->returnValue(array()); line 121 in SiteManager and you will get less warnings. You probably need to do the same in other places

Sorry, line 121 in which class?

greg0ire:
SiteManagerTest, sorry

greg0ire:
Looking good @TheCelavi :+1:

greg0ire:
Thanks @TheCelavi !

PR #796, opened 2017-03-20
Prevent error rendering blocks when using shared blocks

### LINKED ISSUE #799: Shared blocks cannot be used because and exception is thrown
#### Sonata packages

```bash
$ composer show sonata-project/*
sonata-project/admin-bundle              3.14.0
sonata-project/block-bundle              3.3.1
sonata-project/cache                     1.0.7
sonata-project/cache-bundle              2.3.1
sonata-project/classification-bundle     3.3.0
sonata-project/core-bundle               3.2.0
sonata-project/datagrid-bundle           2.2.1
sonata-project/doctrine-extensions       1.0.2
sonata-project/doctrine-orm-admin-bundle 3.1.3
sonata-project/easy-extends-bundle       2.1.10
sonata-project/exporter                  1.7.1
sonata-project/formatter-bundle          3.2.0
sonata-project/google-authenticator      1.0.2
sonata-project/intl-bundle               2.3.0
sonata-project/media-bundle              3.5.0
sonata-project/notification-bundle       3.1.0
sonata-project/page-bundle               3.4.0
sonata-project/seo-bundle                2.1.0
sonata-project/user-bundle               3.2.3
```

#### Symfony packages

```bash
$ composer show symfony/*
symfony/assetic-bundle     v2.8.1
symfony/monolog-bundle     v3.0.3
symfony/phpunit-bridge     v2.8.18
symfony/polyfill-apcu      v1.3.0
symfony/polyfill-intl-icu  v1.3.0
symfony/polyfill-mbstring  v1.3.0
symfony/polyfill-php54     v1.3.0
symfony/polyfill-php55     v1.3.0
symfony/polyfill-php56     v1.3.0
symfony/polyfill-php70     v1.3.0
symfony/polyfill-util      v1.3.0
symfony/security-acl       v2.8.0
symfony/swiftmailer-bundle v2.5.3
symfony/symfony            v2.8.18
```

#### PHP version

```bash
$ php -v
PHP 5.5.31 (cli) (built: Jan  6 2016 12:57:50)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2015, by Zend Technologies
```

## Subject
Shared block cannot be used because an exception is thrown.

## Steps to reproduce
Create a shared block and add it to a CMS page. Create a snapshot and open the page in your browser.

## Expected results

The shared block is displayed.

## Actual results
An exception is thrown:

> Impossible to access an attribute ("id") on a null variable.
> 500 Internal Server Error - Twig_Error_Runtime

gremo:

Closes #799.

## Changelog

```markdown
### Fixed
- use `is not null` instead of `is defined` in `Block/block_base.html.twig`
```
## Subject

Shared blocks are not linked to pages. Thus, for a shared block, the page variable is `null` **but still defined**.

The default template `Block/block_base.html.twig` does the following check: `{% if block.page is defined %}`.This check is wrong because it returns `true` and the following `{{ block.page.id }}` throws an exception:

> Impossible to access an attribute ("id") on a null variable.
> 500 Internal Server Error - Twig_Error_Runtime

However, when SonataPageBundle is installed and enabled, the template `Block/block_base.html.twig` is used as base template also in the admin. In this case, the page variable is undefined. 

Thus the right check to do is:` {% if block.page|default(false) %}`.

### DIFF
--- a/Resources/views/Block/block_base.html.twig
+++ b/Resources/views/Block/block_base.html.twig
@@ -12,7 +12,7 @@ file that was distributed with this source code.
 {% if sonata_page.isInlineEditionOn %}
     <div id="cms-block-{{ block.id }}"
          class="cms-block{% if block.hasParent() %} cms-block-element{% endif %}{% block block_class %}{% endblock %}"
-        {% if sonata_page.isEditor and block.page is defined %}
+        {% if sonata_page.isEditor and block.page is not null %}
             data-id="{{ block.id }}"
             data-name="{{ block.name }}"
             data-role="{% block block_role %}block{% endblock %}"
@@ -20,7 +20,7 @@ file that was distributed with this source code.
         {% endif %}
         >
 {% elseif sonata_page.isEditor() %}
-    <!-- start rendering, block.id: {{ block.id }} - page.id: {% if block.page is defined %}{{ block.page.id }}{% else %}no related page{% endif %} - name: {{ block.name }} -->
+    <!-- start rendering, block.id: {{ block.id }} - page.id: {% if block.page|default(false) %}{{ block.page.id }}{% else %}no related page{% endif %} - name: {{ block.name }} -->
 {% endif %}
 
 {% block block %}EMPTY CONTENT{% endblock %}
@@ -28,5 +28,5 @@ file that was distributed with this source code.
 {% if sonata_page.isInlineEditionOn %}
     </div>
 {% elseif sonata_page.isEditor() %}
-    <!-- end rendering, block.id: {{ block.id }} - page.id: {% if block.page is defined %}{{ block.page.id }}{% else %}no related page{% endif %}  - name: {{ block.name }} -->
+    <!-- end rendering, block.id: {{ block.id }} - page.id: {% if block.page|default(false) %}{{ block.page.id }}{% else %}no related page{% endif %}  - name: {{ block.name }} -->
 {% endif %}


Resources/views/Block/block_base.html.twig
```diff
@@ -12,21 +12,21 @@ file that was distributed with this source code.
...
 
 {% block block %}EMPTY CONTENT{% endblock %}
 
 {% if sonata_page.isInlineEditionOn %}
     </div>
 {% elseif sonata_page.isEditor() %}
-    <!-- end rendering, block.id: {{ block.id }} - page.id: {% if block.page is defined %}{{ block.page.id }}{% else %}no related page{% endif %}  - name: {{ block.name }} -->
-{% endif %}
\ No newline at end of file
+    <!-- end rendering, block.id: {{ block.id }} - page.id: {% if block.page|default(false) %}{{ block.page.id }}{% else %}no related page{% endif %}  - name: {{ block.name }} -->
```
OskarStark: why not use `is not null` here too, instead of the `|default(false)`?

gremo: Block/block_base.html.twig is used as base template also in the admin. In this case, the page variable is undefined.

OskarStark: thanks for the feedback, this sounds reasonable to me 👍

gremo:
@greg0ire can I change the message using github interface? Because I'm at work now and can't access the repo from here.

greg0ire:
> @greg0ire can I change the message using github interface? Because I'm at work now and can't access the repo from here.

Not that I know of. Out of curiosity, why can't you access the repo from work?

gremo:
@greg0ire Ok thanks. My git configuration (ssh keys and so no) it's not on my workstation

greg0ire:
You can add your ssh keys from work to your GH account if that is what is blocking you.

gremo:
@greg0ire you mean like this? I forced the update, please check! Thanks!

greg0ire:
Doesn't look like the message changed

greg0ire:
Maybe you did change it, but I meant to include all this : 

> Shared blocks are not linked to pages. Thus, for a shared block, the page variable is null but still defined.
>
> The default template Block/block_base.html.twig does the following check: {% if block.page is defined %}.This check is wrong because it returns true and the following {{ block.page.id }} throws an exception:
> 
>     Impossible to access an attribute ("id") on a null variable.
>     500 Internal Server Error - Twig_Error_Runtime
> 
> However, when SonataPageBundle is installed and enabled, the template Block/block_base.html.twig is used as base template also in the admin. In this case, the page variable is undefined.
> 
> Thus the right check to do is:{% if block.page|default(false) %}.

greg0ire:
Looks ok now !

gremo:
@greg0ire I was looking at the title of this PR, not the subject itself 👍 not, after messing up with commits (I'm really good at git) It should be fine.

greg0ire:
Thanks @gremo !

PR #806, opened 2017-04-25
3.x dev

### LINKED ISSUE #744: SharedBlock 'block.page.id' is null and bypass 'is defined'
Using a sharedBlock on my website result in this error 

`Impossible to access an attribute ("id") on a null variable in SonataPageBundle:Block:block_base.html.twig at line 23.`

the line concerned is `<!-- start rendering, block.id: {{ block.id }} - page.id: {% if block.page is defined %}{{ block.page.id }}{% else %}no related page{% endif %} - name: {{ block.name }} -->`

apparently, on a shared block, the 'block.page' variable is defined, but set to null, as twig consider a variable defined, even with a null value, its try to display the id of the page

I would recommend to simply change 
`{% if block.page is defined %}{{ block.page.id }}{% else %} `
to 
`{% if block.page is not null %}{{ block.page.id }}{% else %}`

as well as on the line 31.

jayesbe:
I am targeting this branch, because it has an issue where the page property may not exist on a shared block and fails to render. Can't even login to the Admin dashboard. 

Closes #744 

## Changelog

Check if block.page is defined before checking if the block.page property is null. 

### Fixed

Rendering failure when block.page property does not exist.

### DIFF
--- a/Resources/views/Block/block_base.html.twig
+++ b/Resources/views/Block/block_base.html.twig
@@ -12,7 +12,7 @@ file that was distributed with this source code.
 {% if sonata_page.isInlineEditionOn %}
     <div id="cms-block-{{ block.id }}"
          class="cms-block{% if block.hasParent() %} cms-block-element{% endif %}{% block block_class %}{% endblock %}"
-        {% if sonata_page.isEditor and block.page is not null %}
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @jayesbe !

jayesbe:
@greg0ire Thanks. Sorry could not squash. Will remember for next time.

PR #813, opened 2017-05-06
Remove scope attribute

greg0ire:

## Changelog

```markdown
### Fixed
- deprecation warning about scope attributes
```

## Subject

Request and response listener always have access to the current request
object, by definition, which means there is no point in giving them the
request scope.

I'm really unsure about this. @skoubin @aimeric @Id4v @jmontoyaa @robbert-vdh @pbrzoski @bruno-ds can any of you test this and report back?

### DIFF
--- a/Resources/config/page.xml
+++ b/Resources/config/page.xml
@@ -46,14 +46,14 @@
             <argument type="service" id="sonata.page.decorator_strategy"/>
             <argument type="service" id="sonata.seo.page"/>
         </service>
-        <service id="sonata.page.response_listener" class="%sonata.page.response_listener.class%" scope="request">
+        <service id="sonata.page.response_listener" class="%sonata.page.response_listener.class%">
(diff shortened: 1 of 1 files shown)


bruno-ds:
I have started with symfony 2.8, since this release the scopes are deprecated so  I didn't used them at all.
I see only one potential trap : the service is constructed for each sub-request, so having a shared one could potentially lead to side effect, but it is should be a very rare one, no?

greg0ire:
@bruno-ds yeah I thought of that but both services look stateless to me.

greg0ire:
@sonata-project/contributors , what do you think? I'm thinking at worst, if this breaks, we will be able to revert this AND provide an explanation of why it is important.

greg0ire:
ping @sonata-project/contributors

OskarStark:
sounds reasonable to me

OskarStark:
Thank you @greg0ire 👍

PR #810, opened 2017-04-30
Open Symfony 3 compatibility

greg0ire:
I am targeting this branch, because this is BC.

Closes #668 

## Changelog

```markdown
### Fixed
- Compatibility with Symfony 3 was fixed
```

## To do

See https://github.com/sonata-project/SonataPageBundle/projects/1

PR #814, opened 2017-05-07
Prefer buildViolation over addViolationAt

greg0ire:
addViolationAt is deprecated. I had to remove expections over method
calls otherwise the test would become complicated too. In general, you
should not do that kind of thing in a unit test. See
http://elnur.pro/testing-repositories/

Refs #810 

## Changelog

```markdown
### Fixed
- deprecation error message about `addViolationAt`
```


OskarStark:
Thank you @greg0ire 👍

PR #818, opened 2017-05-23
fix createSiteCommand for symfony3

Dunduro:
I am targeting this branch, because BC

## Changelog

```markdown


### Fixed
fixed the syntax change necessary for question helper as opposed to dialog helper
```

### DIFF
--- a/Command/CreateSiteCommand.php
+++ b/Command/CreateSiteCommand.php
@@ -14,6 +14,8 @@ namespace Sonata\PageBundle\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Question\Question;
 
 /**
  * Create a site.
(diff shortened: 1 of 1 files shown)


Dunduro:
@greg0ire & @core23  :| i made a booboo, syntax of ask changed between the dialog helper and question helper but this should fix it.

greg0ire:
> this should fix it

Can you please make sure of that?

Command/CreateSiteCommand.php
```diff
@@ -80,7 +80,24 @@ public function execute(InputInterface $input, OutputInterface $output)
             $values[$name] = $input->getOption($name);
 
             while ($values[$name] == null) {
-                $values[$name] = $helper->ask($output, sprintf('Please define a value for <info>Site.%s</info> : ', $name));
+                /*
+                 * NEXT_MAJOR: remove if statement and use the question helper only (true value of if) when dropping sf < 2.5
+                 */
+                if ($this->getHelperSet()->has('question')) {
+                    $question = new \Symfony\Component\Console\Question\Question(
```
OskarStark: please use as use statement instead of the FQCN here

Dunduro: but wont that create a problem for backwards compatibility?  before symfony 2.5 it doesn't exist

greg0ire: No it won't. Try adding a use statement referring to some class that does not exist if you want to check.

Dunduro: strange i couldnt reply to your previous comment, but tested both 2.3 and 3.2.8. it slipped through the cracks previously because i was using the arguments in the command line but now i properly tested with and without arguments.

i've used both the question and the confirmation question from previous pull request so should be nice and neat now

Dunduro: ping @OskarStark

greg0ire:
Thanks @Dunduro !

PR #825, opened 2017-06-13
add target redirect type to page

n3o77:
I am targeting this branch, because it adds functionality to the latest version.

## Changelog
```markdown

### Changed

- Allow to specify the type of the redirect (301 or 302) when a target on the page is selected.
- Displays if a page is redirected and if so to which route and its redirect type in the tree view.
- Allows to filter pages which are a redirect or not.
- Displays the status code of the redirect on the redirect page.

```
## Subject

We had to add multiple 301 redirects via the backend for some SEO stuff. This PR adds the functionality for this in configurable manner. I assume this could be useful for others.

### DIFF
--- a/src/Admin/PageAdmin.php
+++ b/src/Admin/PageAdmin.php
@@ -400,6 +400,13 @@ class PageAdmin extends AbstractAdmin
                             'siteId' => $this->getSubject() ? $this->getSubject()->getSite()->getId() : null,
                         ],
                     ])
+                ->add('targetRedirectType', method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
+                    ? 'Symfony\Component\Form\Extension\Core\Type\ChoiceType' : 'choice', [
+                        'choices' => [
+                            '301' => 'Permanent (301)',
+                            '302' => 'Temporary (302)',
+                        ],
+                    ])
                 ->end()
             ;
         }
--- a/src/Page/PageServiceManager.php
+++ b/src/Page/PageServiceManager.php
@@ -126,11 +126,11 @@ class PageServiceManager implements PageServiceManagerInterface
     {
         if ($page->getTarget()) {
             $page->addHeader('Location', $this->router->generate($page->getTarget()));
-            $response = new Response('', 302, $page->getHeaders() ?: []);
-        } else {
-            $response = new Response('', 200, $page->getHeaders() ?: []);
+            $redirectType = $page->getTargetRedirectType() ? $page->getTargetRedirectType() : 302;
+
+            return new Response('', $redirectType, $page->getHeaders() ?: []);
         }
 
-        return $response;
+        return new Response('', 200, $page->getHeaders() ?: []);
     }
 }
--- a/src/Resources/config/doctrine/BasePage.orm.xml
+++ b/src/Resources/config/doctrine/BasePage.orm.xml
@@ -22,6 +22,7 @@
         <field name="templateCode" type="string" column="template" nullable="false"/>
         <field name="createdAt" type="datetime" column="created_at"/>
         <field name="updatedAt" type="datetime" column="updated_at"/>
+        <field name="targetRedirectType" type="integer" column="target_redirect_type" nullable="true" default="302"/>
         <lifecycle-callbacks>
             <lifecycle-callback type="prePersist" method="prePersist"/>
             <lifecycle-callback type="preUpdate" method="preUpdate"/>
--- a/src/Model/Page.php
+++ b/src/Model/Page.php
@@ -174,6 +174,11 @@ abstract class Page implements PageInterface
     protected static $slugifyMethod;
 
     /**
+     * @var int
+     */
+    protected $targetRedirectType = 302;
+
+    /**
      * {@inheritdoc}
      */
     public function __construct()
@@ -563,6 +568,14 @@ abstract class Page implements PageInterface
     /**
      * {@inheritdoc}
      */
+    public function getTargetRedirectType()
+    {
+        return $this->targetRedirectType;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function addSnapshot(SnapshotInterface $snapshot)
     {
         $this->snapshots[] = $snapshot;
@@ -571,13 +584,19 @@ abstract class Page implements PageInterface
     }
 
     /**
-     * Set target.
-     *
-     * @param PageInterface $target
+     * {@inheritdoc}
      */
     public function setTarget(PageInterface $target = null)
     {
         $this->target = $target;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setTargetRedirectType($statusCode = 302)
+    {
+        $this->targetRedirectType = $statusCode;
     }
 
     /**
--- a/src/Model/PageInterface.php
+++ b/src/Model/PageInterface.php
@@ -277,6 +277,20 @@ interface PageInterface
     public function getTarget();
 
     /**
+     * Set HTTP Status Code for Redirect.
+     *
+     * @param int|null $statusCode
+     */
+    public function setTargetRedirectType($statusCode = 302);
+
+    /**
+     * Get HTTP Status Code for the Redirect.
+     *
+     * @return int
+     */
+    public function getTargetRedirectType();
+
+    /**
      * Set parent.
      *
      * @param PageInterface $parent
--- a/src/Model/SnapshotPageProxy.php
+++ b/src/Model/SnapshotPageProxy.php
@@ -695,6 +695,22 @@ class SnapshotPageProxy implements SnapshotPageProxyInterface
     /**
      * {@inheritdoc}
      */
+    public function getTargetRedirectType()
+    {
+        return $this->getPage()->getTargetRedirectType();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setTargetRedirectType($statusCode = 302)
+    {
+        return $this->getPage()->setTargetRedirectType($statusCode);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     private function load()
     {
         if (!$this->page && $this->transformer) {
--- a/src/Resources/translations/SonataPageBundle.de.xliff
+++ b/src/Resources/translations/SonataPageBundle.de.xliff
@@ -603,7 +603,8 @@
                 <target><![CDATA[
              Diese Seite ist als Weiterleitung konfiguriert. Um sie zu konfigurieren
              wurde die Weiterleitung verhindert.<br />
-             Um der Weiterleitung zu folgen klicken Sie bitte hier: <a href="%url%">%url%</a>
+             Um der Weiterleitung zu folgen klicken Sie bitte hier: <a href="%url%">%url%</a><br />
+             Status Code: %statusCode%
           ]]></target>
             </trans-unit>
             <trans-unit id="151" resname="filter.label_parent">
@@ -921,6 +922,14 @@
             <trans-unit id="form.label_icon">
                 <source>form.label_icon</source>
                 <target>Icon</target>
+            </trans-unit>
+            <trans-unit id="228" resname="filter.label_is_redirect">
+                <source>filter.label_is_redirect</source>
+                <target>Ist Weiterleitung</target>
+            </trans-unit>
+            <trans-unit id="229" resname="form.label_target_redirect_type">
+                <source>form.label_target_redirect_type</source>
+                <target>Status-Code für Weiterleitung</target>
             </trans-unit>
         </body>
     </file>
--- a/src/Resources/translations/SonataPageBundle.en.xliff
+++ b/src/Resources/translations/SonataPageBundle.en.xliff
@@ -604,7 +604,8 @@
              This page is configured to redirect the end user to another page. In order
              to configure it, the redirect has been blocked for editor only. <br />
              <br />
-             Please click here to follow the redirection: <a href="%url%">%url%</a>.
+             Please click here to follow the redirection: <a href="%url%">%url%</a>.<br />
+             Status Code: %statusCode%
           ]]></target>
             </trans-unit>
             <trans-unit id="151" resname="filter.label_parent">
@@ -922,6 +923,14 @@
             <trans-unit id="form.label_icon">
                 <source>form.label_icon</source>
                 <target>Icon</target>
+            </trans-unit>
+            <trans-unit id="228" resname="filter.label_is_redirect">
+                <source>filter.label_is_redirect</source>
+                <target>Is Redirect</target>
+            </trans-unit>
+            <trans-unit id="229" resname="form.label_target_redirect_type">
+                <source>form.label_target_redirect_type</source>
+                <target>Redirect Type</target>
             </trans-unit>
         </body>
     </file>
--- a/src/Resources/views/Page/redirect.html.twig
+++ b/src/Resources/views/Page/redirect.html.twig
@@ -17,7 +17,10 @@ file that was distributed with this source code.
         <h2>{{ "title_page_redirected"|trans({}, 'SonataPageBundle') }}</h2>
 
         <p>
-            {{ "message_page_redirected"|trans({'%url%': response.headers.get('Location')}, 'SonataPageBundle')|raw }}
+            {{ "message_page_redirected"|trans({
+                '%url%': response.headers.get('Location'),
+                '%statusCode%': response.statusCode
+            }, 'SonataPageBundle')|raw }}
         </p>
     </div>
 {% endblock%}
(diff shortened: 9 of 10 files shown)


greg0ire:
Nice!

Model/Page.php
```diff
@@ -134,6 +134,11 @@
     protected $target;
 
     /**
+     * @var int
+     */
+    protected $targetRedirectType = 302;
```
greg0ire: Please make this private

Resources/config/doctrine/BasePage.orm.xml
```diff
@@ -22,6 +22,7 @@
         <field name="templateCode" type="string" column="template" nullable="false"/>
         <field name="createdAt" type="datetime" column="created_at"/>
         <field name="updatedAt" type="datetime" column="updated_at"/>
+        <field name="targetRedirectType" type="integer" column="target_redirect_type" nullable="true" default="302"/>
```
greg0ire: You introduce a new field, which means people have to generate and run migrations. Will the software continue to work if they forget to do that? If not, this is a BC-break.

Model/Page.php
n3o77: Could you please explain your reasoning as all the other fields are `protected` too?

greg0ire: My reasoning:
- http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html
- close the API over having it consistent.

n3o77: I agree with this. But I think it's a little weird as this will be the only field which will be private, not that it really matters. I guess changing all to private will introduce more BC-breaks and should be targeted for the next major version.

greg0ire: That's the plan: not introducing new protected fields, and closing the API in the next major if possible.

Resources/config/doctrine/BasePage.orm.xml
n3o77: Database needs to be updated otherwise doctrine will probably throw an error. If the field is null it should work without any problems.

greg0ire: @sonata-project/contributors WDYT? Should this be considered a BC-break or not? Solution exist to make this field optional, like a doctrine loadMetadata event listener, but it's more work.

Model/Page.php
n3o77: Sounds good.

Resources/config/doctrine/BasePage.orm.xml
greg0ire: > probably

@n3o77 please check, I'm really not sure about this. It could make the CI fail though, if like me, you have `doctrine:schema:validate` in CI

n3o77: doctrine:schema:update needs to be executed otherwise doctrine throws an error. If the field is null it will fall back to a 302 redirect. Just tested through it.

Model/PageInterface.php
```diff
@@ -277,6 +277,18 @@ public function setTarget(PageInterface $target = null);
     public function getTarget();
 
     /**
+     * @param int|null $targetRedirectType
+     */
+    public function setTargetRedirectType($targetRedirectType = null);
+
+    /**
+     * Get target.
```
greg0ire: really :trollface: ?

Resources/config/doctrine/BasePage.orm.xml
greg0ire: May I ask what error is thrown?

greg0ire: Here is the event that could help make this optional: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html#load-classmetadata-event

Resources/views/Page/redirect.html.twig
```diff
@@ -17,7 +17,7 @@ file that was distributed with this source code.
         <h2>{{ "title_page_redirected"|trans({}, 'SonataPageBundle') }}</h2>
 
         <p>
-            {{ "message_page_redirected"|trans({'%url%': response.headers.get('Location')}, 'SonataPageBundle')|raw }}
+            {{ "message_page_redirected"|trans({'%url%': response.headers.get('Location'), '%statusCode%': response.statusCode}, 'SonataPageBundle')|raw }}
```
greg0ire: Please linebreak this.

Resources/config/doctrine/BasePage.orm.xml
n3o77: Sure:
InvalidFieldNameException
```
An exception occurred while executing 'SELECT p0_.route_name AS route_name_0, p0_.page_alias AS page_alias_1, p0_.type AS type_2, p0_.position AS position_3, p0_.enabled AS enabled_4, p0_.decorate AS decorate_5, p0_.edited AS edited_6, p0_.name AS name_7, p0_.slug AS slug_8, p0_.url AS url_9, p0_.custom_url AS custom_url_10, p0_.request_method AS request_method_11, p0_.title AS title_12, p0_.meta_keyword AS meta_keyword_13, p0_.meta_description AS meta_description_14, p0_.javascript AS javascript_15, p0_.stylesheet AS stylesheet_16, p0_.raw_headers AS raw_headers_17, p0_.template AS template_18, p0_.created_at AS created_at_19, p0_.updated_at AS updated_at_20, p0_.target_redirect_type AS target_redirect_type_21, p0_.id AS id_22, p0_.site_id AS site_id_23, p0_.parent_id AS parent_id_24, p0_.target_id AS target_id_25 FROM page__page p0_ WHERE p0_.site_id = 1 ORDER BY p0_.position ASC':
```

> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'p0_.target_redirect_type' in 'field list'

greg0ire: I see. Makes sense indeed :/

n3o77: Yes :( Not sure how big of a deal this can be. I know that in our projects we have the schema update in our deployment process integrated and didn't have problems with it but this of course doesn't have to be true for every project out there.

Would be great to have that feature in there so I don't have to run X projects on a custom build.

I haven't worked with the event you mentioned, i'll need some more time to read into that.

greg0ire: The event could be used to remove the field from the mapping based on a configuration node value that would trigger a deprecation notice if not set to true, and would be ignored in next major version.

Model/Page.php
```diff
@@ -174,6 +174,11 @@
     protected static $slugifyMethod;
 
     /**
+     * @var int
+     */
+    private $targetRedirectType = 302;
```
n3o77: @greg0ire i just installed this in another project but this `private` makes some problems which i don't really understand. Exception: `Property Sonate\PageBundle\Entity\BasePage::$targetRedirectType does not exist.`

As soon i change it back to protected everything works just fine. Maybe because the package is extended  (just a guess)? Unfortunately i don't have much time to look into that this week.

screen shot 2017-06-20 at 16 58 44

greg0ire: Could it be some kind of caching issue?

n3o77: That was my first thought too but I cleared all the caches and also deleted the cache folders.

greg0ire: Do you reproduce the issue in CLI, with the following script:

```php
require './vendor/autoload.php';
new \ReflectionProperty(Sonata\PageBundle\Entity\BasePage::class, 'targetRedirectType');
```

Also, please refrain from posting screenshots of text, it's bad for several reasons I contributed to the sf docs. There is a plain text stack trace at the bottom of the page.

Admin/PageAdmin.php
```diff
@@ -311,6 +311,22 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
                 ),
                 'field_type' => 'choice',
             ))
+            ->add('isRedirect', 'doctrine_orm_callback', array(
+                'callback' => function ($queryBuilder, $alias, $field, $data) {
+                    if (in_array($data['value'], array('yes', 'no'))) {
+                        $queryBuilder->andWhere(sprintf('%s.target %s', $alias, $data['value'] == 'yes' ? 'IS NOT NULL' : 'IS NULL'));
```
Soullivaneuh: Line breaks needed.

Admin/PageAdmin.php
```diff
@@ -401,6 +417,10 @@ protected function configureFormFields(FormMapper $formMapper)
                             'siteId' => $this->getSubject() ? $this->getSubject()->getSite()->getId() : null,
                         ),
                     ))
+                    ->add('targetRedirectType', 'Symfony\Component\Form\Extension\Core\Type\ChoiceType', array('choices' => array(
```
Soullivaneuh: Line break needed. Could be done at `array('choices'`

Model/Page.php
Soullivaneuh: Maybe add this constant? https://github.com/symfony/symfony/blob/3bbb657560615b80c8a9b3b6da8602349bd3fbd9/src/Symfony/Component/HttpFoundation/Response.php#L36

Model/Page.php
```diff
@@ -560,6 +565,11 @@ public function getTarget()
         return $this->target;
     }
 
+    public function getTargetRedirectType()
```
Soullivaneuh: Missing PHPDoc.

Model/PageInterface.php
```diff
@@ -277,6 +277,18 @@ public function setTarget(PageInterface $target = null);
     public function getTarget();
 
     /**
+     * @param int|null $targetRedirectType
+     */
+    public function setTargetRedirectType($targetRedirectType = null);
+
+    /**
+     * Get targetRedirectType.
```
Soullivaneuh: Useless comment.

Page/PageServiceManager.php
```diff
@@ -126,7 +126,8 @@ protected function createResponse(PageInterface $page)
     {
         if ($page->getTarget()) {
             $page->addHeader('Location', $this->router->generate($page->getTarget()));
-            $response = new Response('', 302, $page->getHeaders() ?: array());
+            $redirectType = $page->getTargetRedirectType() ? $page->getTargetRedirectType() : 302;
```
Soullivaneuh: Maybe use this constant? https://github.com/symfony/symfony/blob/3bbb657560615b80c8a9b3b6da8602349bd3fbd9/src/Symfony/Component/HttpFoundation/Response.php#L36

Model/Page.php
OskarStark: good idea 👍

Admin/PageAdmin.php
```diff
@@ -401,6 +417,10 @@ protected function configureFormFields(FormMapper $formMapper)
...
+                        '301' => 'Permanent (301)',
+                        '302' => 'Temporary (302)',
```
OskarStark: constants here too?

Model/Page.php
```diff
@@ -581,6 +591,14 @@ public function setTarget(PageInterface $target = null)
     }
 
     /**
+     * @param int|null $targetRedirectType
+     */
+    public function setTargetRedirectType($targetRedirectType = null)
```
OskarStark: maybe rename type to statusCode?

Model/PageInterface.php
```diff
@@ -277,6 +277,18 @@ public function setTarget(PageInterface $target = null);
     public function getTarget();
 
     /**
+     * @param int|null $targetRedirectType
+     */
+    public function setTargetRedirectType($targetRedirectType = null);
```
OskarStark: why null as default? why not using the member variable 302 as default here?

Resources/translations/SonataPageBundle.de.xliff
```diff
@@ -914,6 +915,14 @@
                 <source>pages.label_no_site_selected</source>
                 <target>Keine Seiten ausgewählt</target>
             </trans-unit>
+            <trans-unit id="228" resname="filter.label_is_redirect">
+                <source>filter.label_is_redirect</source>
+                <target>Ist Weiterleitung</target>
+            </trans-unit>
+            <trans-unit id="229" resname="form.label_target_redirect_type">
+                <source>form.label_target_redirect_type</source>
+                <target>Weiterleitungs Art</target>
```
OskarStark: if renaming type to statusCode:
`Status-Code für Weiterleitung`

n3o77:
Thanks everybody for the feedback, i'll have some more time again by the end of the week and will make the changes then.

Model/Page.php
n3o77: Haven't got this working with `private` as the extending class won't have access to the field anymore and so doctrine can't access it too. This will probably affect all bundles which "depend" on the easy-extends bundle.

greg0ire: I never use protected fields with doctrine. If there is a getter and a setter, it should work, shouldn't it?

greg0ire: Hmmm nevermind, the doctrine docs seem to suggest otherwise http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html

Page/PageServiceManager.php
n3o77: Constant not available in symfony 2.3 :(
https://github.com/symfony/http-foundation/blob/2.3/Response.php

https://travis-ci.org/sonata-project/SonataPageBundle/jobs/266601244

Admin/PageAdmin.php
```diff
@@ -432,6 +449,12 @@ protected function configureFormFields(FormMapper $formMapper)
                             ),
                         )
                     )
+                ->add('targetRedirectType', 'Symfony\Component\Form\Extension\Core\Type\ChoiceType', array(
+                        'choices' => array(
+                            '301' => 'Permanent (301)',
```
core23: This won't work with older symfony versions

n3o77: How does it need to be done to work with older versions?

greg0ire: Like that: https://github.com/sonata-project/SonataAdminBundle/blob/f7036764853ca0c5a6994635f114eab5434194e1/Form/Extension/ChoiceTypeExtension.php#L67

Model/PageInterface.php
```diff
@@ -277,6 +277,18 @@ public function setTarget(PageInterface $target = null);
     public function getTarget();
 
     /**
+     * @param int|null $statusCode
```
core23: Please add a description

src/Page/PageServiceManager.php
```diff
@@ -126,7 +126,8 @@ protected function createResponse(PageInterface $page)
     {
         if ($page->getTarget()) {
             $page->addHeader('Location', $this->router->generate($page->getTarget()));
-            $response = new Response('', 302, $page->getHeaders() ?: []);
+            $redirectType = $page->getTargetRedirectType() ? $page->getTargetRedirectType() : 302;
+            $response = new Response('', $redirectType, $page->getHeaders() ?: []);
         } else {
```
kunicmarko20: you can use `return` instead of `else` here.

n3o77: I'll change that. It was like this in the original file so I just made minimal changes.

kunicmarko20: can you also fix flintCI errors

n3o77: Sure

src/Admin/PageAdmin.php
```diff
@@ -400,6 +400,13 @@ protected function configureFormFields(FormMapper $formMapper)
                             'siteId' => $this->getSubject() ? $this->getSubject()->getSite()->getId() : null,
                         ],
                     ])
+                ->add('targetRedirectType', method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
```
jordisala1991: use AbstractType::class

src/Admin/PageAdmin.php
```diff
@@ -400,6 +400,13 @@ protected function configureFormFields(FormMapper $formMapper)
...
+                    ? 'Symfony\Component\Form\Extension\Core\Type\ChoiceType' : 'choice', [
```
jordisala1991: in fact, do not do this if. Add directly Choice::type

jordisala1991: Sorry because it was asked otherwise on a previous comment, but since the start of this PR we changed our deps an we no longer support SF 2.3, the min is 2.8 and you can use FQCN directly on the formTypes

n3o77: Will do, awesome! With the drop of 2.3 I can also use the constants for the status code.

src/Resources/config/doctrine/BasePage.orm.xml
```diff
@@ -22,6 +22,7 @@
         <field name="templateCode" type="string" column="template" nullable="false"/>
         <field name="createdAt" type="datetime" column="created_at"/>
         <field name="updatedAt" type="datetime" column="updated_at"/>
+        <field name="targetRedirectType" type="integer" column="target_redirect_type" nullable="true" default="302"/>
```
core23: This is a major change, because you need to run a database update

greg0ire: Related: https://github.com/sonata-project/SonataPageBundle/pull/825#discussion_r121701047

n3o77: Ok, i'll create a new PR for master.

PR #839, opened 2017-07-14
Page-Composer: removes custom status-code check (relying on Response-Status instead)

mvhirsch:
I am targeting this branch, because of an introduced update of sonata-project/admin-bundle which should be synchronized in page-bundle too.

Fixes bug resulted in change of https://github.com/sonata-project/SonataAdminBundle/pull/4449

## Changelog

```markdown
### Fixed
- Composer (JS): Relying of custom status-property; instead rely on Response Status Code
```

## TODO

- [x] Update sonata-page.back.js
- [x] Generate minified version of sonata-page.back.js

## Subject

After updating to SonataAdminBundle 3.17 (or higher) the composer broke.
Reading this PR https://github.com/sonata-project/SonataAdminBundle/pull/4449#issuecomment-298901450 I saw a non-BC-change which should be updated in SonataPageBundle too.
So I removed the necessity to check the body's status-property. Now only the Response-Status-Code is used to determine success/fail.

I just don't have any clue about how you guys generate the minified version of this JS?


greg0ire:
> I just don't have any clue about how you guys generate the minified version of this JS?

@qsomazzi @core23 @plouc @meyerbaptiste do you know how this is usually done? Also, do you know why minified files have the executable bit set 😕 ?

mvhirsch:
Maybe just being pedantic at this but: why is the header-comment included twice in both (back.js & front.js) files?

greg0ire:
Also: why is the first header saying the file is generated ?

greg0ire:
Or is the header generated?

greg0ire:
Best way to find out which is right might be to remove both and let StyleCI or whatever re add them

greg0ire:
Hmmm, it looks like StyleCI just doesn't care about JS files, which makes sense... sorry for the trouble.

mvhirsch:
> Hmmm, it looks like StyleCI just doesn't care about JS files, which makes sense... sorry for the trouble.

Nevermind :-)
Anyway, should I revert the commit or do we remove the comment anyway?

greg0ire:
I think we should remove one of them. Can you look at other sonata bundles?

mvhirsch:
> I think we should remove one of them. Can you look at other sonata bundles?

I've taken a look at PHP-files and cross-checked with another [JS-file: SonataAdminBundle's Admin.js](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Resources/public/Admin.js). 
So this should be fine now :-)

Left open: how to minify? Should I throw it against [Googles online JS compiler](https://closure-compiler.appspot.com/home)?

greg0ire:
> Left open: how to minify? Should I throw it against Googles online JS compiler?

Yes, please try that, it should be ok, I guess...

greg0ire:

How to do that?

mvhirsch:
@greg0ire I squashed it :-)

greg0ire:
@sonata-project/contributors please review

greg0ire:
Thanks @mvhirsch

PR #844, opened 2017-08-02
Changed fallback translation domain to SonataBlockBundle in page composer

### LINKED ISSUE #842: [RFC] Change translation domain for block names
Today we are using the `SonataPageBundle` as the translation domain for block name translation. This is not a good idea, because we need the [same translation](https://github.com/sonata-project/SonataPageBundle/blob/3.x/Resources/views/BlockAdmin/select_type.html.twig#L30) for the `SonataDashboardBundle` to get a localized block name.

Maybe we should change [this behavior](https://github.com/sonata-project/SonataPageBundle/blob/3.x/Resources/views/BlockAdmin/select_type.html.twig#L30) to use `SonataBlockBundle` as the translation domain, which looks more appropriate.

Example translation entry:
https://github.com/sonata-project/SonataPageBundle/blob/3.x/Resources/translations/SonataPageBundle.en.xliff#L838-L840

core23:
I am targeting this branch, because this is a only a soft BC break. It only affects the fallback behavior.

Closes #842

## Changelog

```markdown
### Changed
- Changed fallback translation domain to `SonataBlockBundle` in page composer
```

## Subject

We should use a common fallback translation domain for all page / block composers across all bundles. Atm we have a page composer in this bundle and a dashboard composer in the [SonataDashboardBundle](https://github.com/sonata-project/SonataDashboardBundle).

### DIFF
--- a/Resources/views/BlockAdmin/select_type.html.twig
+++ b/Resources/views/BlockAdmin/select_type.html.twig
@@ -27,7 +27,7 @@ file that was distributed with this source code.
                         class="btn btn-app btn-block sonata-block-type"
                         data-toggle="tooltip"
                         data-placement="top"
-                        title="{{ service.blockMetadata.description|trans({}, service.blockMetadata.domain|default('SonataPageBundle')) }}"
+                        title="{{ service.blockMetadata.description|trans({}, service.blockMetadata.domain|default('SonataBlockBundle')) }}"
                             >
                         {% if not service.blockMetadata.image %}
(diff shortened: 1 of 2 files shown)


greg0ire:
Restarting the build

OskarStark:
`minor` or `patch` @core23 @greg0ire ?

core23:
Patch, because it's kind of a soft BC break

OskarStark:
Thank you @core23 👍

PR #860, opened 2017-09-04
fix Twig check for block existence

xabbuh:
Silently using an undefined block breaks in Twig 2.x. This behaviour was deprecated in Twig 1.29.

# Changelog

```markdown
## Fixed
- compatibility with Twig 2.0 was improved
```

### DIFF
--- a/Resources/views/base_error.html.twig
+++ b/Resources/views/base_error.html.twig
@@ -8,7 +8,7 @@ For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
 
 #}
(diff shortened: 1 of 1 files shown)


greg0ire:
Green. @sonata-project/contributors please review

greg0ire:
Maybe there is, but I don't remember... @xabbuh ?

jlamur:
The `default` filter returns `false` on empty strings.

xabbuh:
The `default` filter doesn't work with the `block()` function:

bildschirmfoto 2017-09-04 um 20 27 29

greg0ire:
Thanks @xabbuh !

PR #862, opened 2017-09-05
Don't call Extension::addClassesToCompile() on php versions greater than 7

acrobat:
I am targeting this branch, because of a deprecation fix for symfony 3.3.

```markdown
### Fixed
- Don't call Extension::addClassesToCompile() on php versions greater than 7
```

## Subject

This method is deprecated in 3.3. See symfony/symfony#20735


greg0ire:
Thanks @acrobat

PR #863, opened 2017-09-14
Update PageAdmin.php legacy code bug fix 

### LINKED ISSUE #989: PageSelectorType returns Block instead of Page entity
I'm trying to add a PageSelectorType form field to my block form so I can choose a page for the current site.

I'm doing it like this:
```$formMapper
            ->add(
                'settings',
                'sonata_type_immutable_array',
                array(
                    'keys' => array(
                        array(
                            'banner_page',
                            PageSelectorType::class,
                            array(
                                'model_manager' => $formMapper->getAdmin()->getModelManager(),
                                'class' => $formMapper->getAdmin()->getClass(),
                                'site' => $block->getPage()->getSite(),
                                'required' => true,
                                'label' => 'form.label_page',
                            ),
                        ),
                        array($imageBuilder, null, array()),
                        array('title', TextType::class, array('required' => false)),
                        array('cta_button_text', TextType::class, array('required' => false)),
                    ),
                    'translation_domain' => 'SonataPageBundle',
                )
            ); 
```

This renders a correct page selector but when I'm saving the block and I put a breakpoint inside the prepersist or preupdate function I see that the banner_page property inside my block settings is a Block entity instead of a Page entity. This means that the chosen page id in the form gets parsed to a block entity with the same id of the chosen page. This also causes the form to do weird stuff when selecting a page with an id if there's not block with that same id...

tl;dr => pageselector translates page id to a Block entity instead of a Page entity...

rchntrl:
I am targeting this branch, because Calling this function caused a bug because it interferes 
Fixes #989.

### DIFF
--- a/Admin/PageAdmin.php
+++ b/Admin/PageAdmin.php
@@ -336,7 +336,7 @@ class PageAdmin extends AbstractAdmin
         if ($this->hasSubject() && !$this->getSubject()->getId()) {
             $formMapper
                 ->with('form_page.group_main_label')
-                    ->add('site', null, array('required' => true, 'read_only' => true))
+                    ->add('site', null, array('required' => true, 'attr' => array('readonly' => 'readonly'))
                 ->end()
             ;
         }


OskarStark:
Hi, first of all thank you for your contribution 👍 

Can you please write a correct changelog and fill out the PR header?

greg0ire:

Bad example :

```
```

Good example:

```
Remove call to defraculate()

Fixes #989
```

Admin/PageAdmin.php
```diff
@@ -336,7 +336,7 @@ protected function configureFormFields(FormMapper $formMapper)
         if ($this->hasSubject() && !$this->getSubject()->getId()) {
             $formMapper
                 ->with('form_page.group_main_label')
-                    ->add('site', null, array('required' => true, 'read_only' => true))
+                    ->add('site', null, array('required' => true, 'attr' => array('readonly' => 'readonly'))
```
core23: This will work with old symfony versions?

rchntrl: i didn't check that. I thougth "3.x" means version for Symfony 3. I have tried to use Page bundle with Symfony 3.2 and get error when open  page to create new record.

core23: According to the [composer](https://github.com/sonata-project/SonataPageBundle/blob/3.x/composer.json#L33-L45) file, we still support symfony 2.3 on the current stable branch.

greg0ire:
Hey there!

It looks like you did not fill the pull request template properly. Maybe you
were confused, so let me help you go through this step by step.

# Explaining why you target this branch

> I am targeting this branch, because {reason}.

Here, you are supposed to replace `{reason}` with something like:

- "there is a BC-break" if you picked the master branch
- "there is no BC-break" if you picked the stable branch

# Searching and finding related issues

> Closes #{put_issue_number_here}

Please have a look at the list of issues and pull request, and look for issues
that will be fixed, and pull requests that will become irrelevant should we
merge your PR. Each issue / pull request has a number shown right next to the
title. Kindly replace `{put_issue_number_here}` with whatever issue or PR is
relevant. Of course, you may specify several numbers.
On merge, [these issues and PRs will automatically get
closed](https://help.github.com/articles/closing-issues-using-keywords/).
If there is no issue or pull request that fits there, just remove the sentence
entirely.

# Filling the changelog

That one is really important to get right, because it will be used to craft a
release changelog such as [this
one](https://github.com/sonata-project/SonataAdminBundle/releases/tag/3.23.0).
The first thing you need to know about this is that if your pull request is
pedantic, meaning it does not fix a bug for the user, or adds a new feature, it
is not needed. This is typically the case for PRs that fix the build or add
some documentation.

3 important things:

1. Make sure to fill the changelog section with something directed at the end
user, that answers the question "What benefit do I get from upgrading?".
2. Make sure to remove empty sections.
3. Make sure to keep the markdown markup.

## Filling the TODO list

If you think your PR is ready, just remove it. If you have still some things
left to do, customize it.

## Filling the Subject

Last but not least, describe your PR as well as you can. Bear in mind that some
reviewers do not even use the library they are maintaining, so try to be
crystal clear.

PR #867, opened 2017-09-21
Get request_stack for Symfony 3

schauer:
I am targeting this branch (3.x), because of a bug fix.

## Changelog

```markdown
### Fixed
- deprecation message about the "request" service from the admin controller
```
## Subject

Use the `request_stack` when available, by using `CRUDController::getRequest()`

### DIFF
--- a/Controller/SiteAdminController.php
+++ b/Controller/SiteAdminController.php
@@ -32,11 +32,12 @@ class SiteAdminController extends Controller
      */
     public function snapshotsAction()
     {
+        $request = $this->getRequest();
         if (false === $this->get('sonata.page.admin.snapshot')->isGranted('CREATE')) {
             throw new AccessDeniedException();
         }
 
-        $id = $this->get('request')->get($this->admin->getIdParameter());
+        $id = $request->get($this->admin->getIdParameter());
 
         $object = $this->admin->getObject($id);
 
@@ -46,7 +47,7 @@ class SiteAdminController extends Controller
 
         $this->admin->setSubject($object);
 
-        if ($this->get('request')->getMethod() == 'POST') {
+        if ($request->getMethod() == 'POST') {
             $this->get('sonata.notification.backend')
                 ->createAndPublish('sonata.page.create_snapshots', array(
                     'siteId' => $object->getId(),


Controller/SiteAdminController.php
```diff
@@ -30,13 +31,13 @@ class SiteAdminController extends Controller
      * @throws NotFoundHttpException
      * @throws AccessDeniedException
      */
-    public function snapshotsAction()
+    public function snapshotsAction(Request $request)
```
greg0ire: This is a BC-break for extending classes

greg0ire:
> because of a bug fix

How is this a bug?

Controller/SiteAdminController.php
Soullivaneuh: I think `CRUDController` of admin-bundle handles the request with a different maner.

greg0ire: https://github.com/sonata-project/SonataAdminBundle/blob/c39403f1b801cf2d40d916d2cd00daf753adeec9/Controller/CRUDController.php#L948-L955

greg0ire:
> How is this a bug?

Answering my own question: I think this might fix a deprecation notice

Soullivaneuh:
@greg0ire Indeed. This is not really a bug, but the PR is a patch. :wink:

greg0ire:

How to do that?

greg0ire:
You should probably copy/paste this: https://github.com/sonata-project/SonataAdminBundle/blob/c39403f1b801cf2d40d916d2cd00daf753adeec9/Controller/CRUDController.php#L948-L955

greg0ire:
TODO:

- [ ] copy / paste the method I asked you to copy / paste
- [ ] amend your commit message to reflect what you are doing now.

schauer:
The requested changes were realized. Is everything correct now?

greg0ire:

Bad example :

```
```

Good example:

```
Remove call to defraculate()

Fixes #989
```

Controller/SiteAdminController.php
```diff
@@ -63,4 +64,16 @@ public function snapshotsAction()
             'object' => $object,
         ));
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getRequest()
```
greg0ire: Please make the method final. Not sure why we didn't do it in the piece of code you copied from.

schauer: The class \Sonata\PageBundle\Controller\SiteAdminController extends the class \Sonata\AdminBundle\Controller\CRUDController. The class \Sonata\AdminBundle\Controller\CRUDController contains the getRequest() I have to copy. I do not know why I had to copy the method at all.

I will do the method final and try to improve my commit message.

greg0ire: Oooooh I did not notice that. Ok then, just remove it, and sorry for the trouble!

greg0ire:
I updated your changelog and your PR subject.

Controller/SiteAdminController.php
greg0ire: `CRUDController` is aliased to `Controller`, that's probably why I did not notice that.

PR #869, opened 2017-09-29
 Removed deprecated calls

core23:
I am targeting this branch, because this is a patch.

## Subject

Removed some more deprecated calls.

PR #871, opened 2017-10-04
Fixed wrong class name

core23:
I am targeting this branch, because this is a patch of #869.

## Subject

Fixed some wrong class names.


greg0ire:
No need for a changelog since there is no release yet.

PR #872, opened 2017-10-05
Workaround for missing getRequest method on sf3

acrobat:
I am targeting this branch, because bugfix for an incompatibility with sf3.

## Changelog

```markdown
### Fixed
- The `getRequest` method is incompatible with sf3 in the page controller
```

## Subject

An incompatibility discovered when upgrading to sf3 and checking the error pages. The `getRequest` method on the controller class is deprecated since sf2.4 and removed in sf3. In the next major this workaround should be replaced with just injecting the service in the action method (possible if we don't support sf2.3 anymore).

### DIFF
--- a/Controller/PageController.php
+++ b/Controller/PageController.php
@@ -18,6 +18,7 @@ use Sonata\PageBundle\Exception\PageNotFoundException;
 use Sonata\PageBundle\Listener\ExceptionListener;
 use Sonata\PageBundle\Page\PageServiceManagerInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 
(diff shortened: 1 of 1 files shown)


halk:
I'm getting the following error after this commit on Symfony 2.8:

`Compile Error: Access level to Sonata\PageBundle\Controller\PageController::getRequest() must be public (as in class Symfony\Bundle\FrameworkBundle\Controller\Controller)`

acrobat:
I will take a look at it later tonight!

core23:
The method exists as public if we are using symfony 2.8. Maybe choose a differnt name?

greg0ire:
Maybe we could just remove if if it exists in 2.8 and in 3?

core23:
> Maybe we could just remove if if it exists in 2.8 and in 3?

The method doesn't exist in 3.x. That is the original issue

greg0ire:
Ooooh this is a PR not an issue, I see... got me confused.

acrobat:
I've renamed the method to fix compatibility with sf2.8. See #887

PR #876, opened 2017-10-09
use new sf3 choices structure

rvanginneken:
I am targeting this branch, because symfony 3 incompatibility.

In case of bug fix, `3.x` **MUST** be targeted.

## Changelog

```markdown
### Fixed
- use new sf3 choices structure
```

## Subject
Changed the choices structure to follow new sf3 standards. Added choices_as_values for BC.

### DIFF
--- a/Form/Type/PageTypeChoiceType.php
+++ b/Form/Type/PageTypeChoiceType.php
@@ -41,10 +41,17 @@ class PageTypeChoiceType extends AbstractType
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefaults([
+        $defaults = [
             'choices' => $this->getPageTypes(),
             'choice_translation_domain' => false,
-        ]);
+        ];
+
+        // NEXT_MAJOR: Remove (when requirement of Symfony is >= 3.0)
+        if (method_exists('Symfony\Component\Form\FormTypeInterface', 'setDefaultOptions')) {
+            $defaults['choices_as_values'] = true;
+        }
+
+        $resolver->setDefaults($defaults);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


greg0ire:
This needs to be a bit more complex to support all versions of sf: https://github.com/sonata-project/SonataAdminBundle/blob/1dd267f943cd1187485b5394fd260e076accacd0/Form/Type/Filter/ChoiceType.php#L81-L93

greg0ire:
This should make your life easier: https://github.com/sonata-project/SonataPageBundle/pull/877

rvanginneken:
@greg0ire I rebased my PR. Could you go a little bit deeper into why this needs to be more complex?

choices_as_values is available in atleast 2.8 and will only throw a deprecation when used in 3.x. Which seems correct to me, since it is a deprecated feature which should be fixed at some point.
There is no need to keep the old structure around, seeing support for <2.8 was just dropped.

greg0ire:
Yeah but you should avoid the deprecation notice (i.e. add the optional conditionally, see the link I provided for the right condition). Basically, take the same code, but remove the < 2.8 compat layer

rvanginneken:
@greg0ire Allright, I assumed deprecation warnings would be okay, since they don't interfere with the user experience and they need to be fixed.
But I applied the change, I assume this is what you meant?

greg0ire:
>  they don't interfere with the user experience

They kind of do, especially when you have hundreds of them you can't fix yourself.

greg0ire:
Please fix the tiny CI issue

rvanginneken:
@greg0ire done :)

greg0ire:
Please fix the TravisCI build

rvanginneken:
@greg0ire travis build (test) fixed :)

greg0ire:
Thanks @rvanginneken !

PR #877, opened 2017-10-09
Drop old versions of php/sf

greg0ire:

## Changelog

```markdown
### Removed
- support for old versions of php and Symfony
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^5.3 || ^7.0",
+        "php": "^7.1",
         "cocur/slugify": "^1.0 || ^2.0",
         "sonata-project/admin-bundle": "^3.4",
         "sonata-project/block-bundle": "^3.3.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^5.3 || ^7.0",
```
jordisala1991: Shouldn't we maintain PHP 7.0? It is still on the green zone

greg0ire: yeah but barely, so we decided to make an exception that time

acrobat: I think this jump is maybe to big for 3.x but something the next major. I would make the requirement `^5.6|^7.0` for now as php language is no big issue for now, I think?

greg0ire:
Please merge

composer.json
greg0ire: Woops too late, but anyway, we took our decision I'm afraid.

Soullivaneuh: @acrobat Please read: http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

Also: https://github.com/sonata-project/dev-kit/issues/285

acrobat: @Soullivaneuh yes I know, it's indeed not a problem to drop support for certain php versions. In my opinion it's not necessary to drop php 5.6 support now as it is still supported and doesn't make maintaining the repo harder (my opinion). But no problem for me as I'm running 7.1 on most applications! :)

PR #881, opened 2017-10-09
Removed php5 version checks

acrobat:
I am targeting this branch, because support for all version below php7.1 are removed on 3.x.

## Changelog

```markdown
### Removed
- Removed php5 version checks
```

## Subject

Some cleanup to remove php5 version checks

### DIFF
--- a/DependencyInjection/SonataPageExtension.php
+++ b/DependencyInjection/SonataPageExtension.php
@@ -84,10 +84,6 @@ class SonataPageExtension extends Extension
         $this->configurePageDefaults($container, $config);
         $this->configurePageServices($container, $config);
 
-        if (PHP_VERSION_ID < 70000) {
-            $this->configureClassesToCompile();
-        }
-
         $container->setParameter('sonata.page.assets', $config['assets']);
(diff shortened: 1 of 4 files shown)


DependencyInjection/SonataPageExtension.php
```diff
@@ -499,73 +495,4 @@ public function configurePageServices(ContainerBuilder $container, array $config
         $definition = $container->getDefinition('sonata.page.page_service_manager');
         $definition->addMethodCall('setDefault', [new Reference($config['default_page_service'])]);
     }
-
-    /**
-     * Add class to compile.
-     */
-    public function configureClassesToCompile()
```
core23: Technical, it's a BC break. Not sure if we could accept this, 'cause this method should never been called in this context

greg0ire: It's fine.

greg0ire:
Thanks @acrobat !

OskarStark:
minor?

acrobat:
I would say so because of the php and symfony version bump. It's possible to release this as a patch version but imo it's cleaner to this in a minor release

PR #882, opened 2017-10-09
fix site readonly attr

rvanginneken:
I am targeting this branch, because there was still an "old" read_only flag.

## Changelog

```markdown
### Fixed
- read_only error for site selection in page admin
```

## Subject

There still was an "old" read_only flag. I moved it to 'attr' and used the boolean flag in both cases where it was used.

### DIFF
--- a/Admin/PageAdmin.php
+++ b/Admin/PageAdmin.php
@@ -327,7 +327,7 @@ class PageAdmin extends AbstractAdmin
         if (!$this->getSubject() || (!$this->getSubject()->isInternal() && !$this->getSubject()->isError())) {
             $formMapper
                 ->with('form_page.group_main_label')
-                    ->add('url', TextType::class, ['attr' => ['readonly' => 'readonly']])
+                    ->add('url', TextType::class, ['attr' => ['readonly' => true]])
                 ->end()
             ;
         }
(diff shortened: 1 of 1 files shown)


greg0ire:

Bad example :

```
```

Good example:

```
Remove call to defraculate()

Fixes #989
```

rvanginneken:
@greg0ire I hope this is a better indication of what I did?

greg0ire:
It is, but what's unclear is the "why" part

Admin/PageAdmin.php
```diff
@@ -327,15 +327,15 @@ protected function configureFormFields(FormMapper $formMapper)
         if (!$this->getSubject() || (!$this->getSubject()->isInternal() && !$this->getSubject()->isError())) {
             $formMapper
                 ->with('form_page.group_main_label')
-                    ->add('url', TextType::class, ['attr' => ['readonly' => 'readonly']])
+                    ->add('url', TextType::class, ['attr' => ['readonly' => true]])
```
greg0ire: Wasn't that valid?

jlamur: I was checking that, it seems like it is not valid, but it was casted correctly: https://github.com/symfony/symfony/pull/10676/files#diff-22a03074077455f3d1128486ef5d9851R82

Using a boolean seems to be favored as this test shows: https://github.com/symfony/symfony/pull/10676/files#diff-e11656489f1e01413320bd5edff0f0c7R1917

rvanginneken: Also it's a flag, which should be able to be turned on or off -> boolean

jlamur:
Thank you @rvanginneken !

rvanginneken:
The why part: errors. I'll add it to the message tomorrow (on a different system atm).

jlamur:
Too late, it's merged! It's not a big deal anyway, thank you.

rvanginneken:
@jlamur Yeah I noticed.  it's still nice to get those pointers though, I'm very used to my local project development flow, so I really need to pay some more attention to my contribution commits :)

PR #751, opened 2016-12-01
fix sonata_seo_context to use the right breadcrumb service

grimpows:
I am targetting this branch, because it's a minor fix.

Closes #749

## Changelog

```markdown

### Fixed
- use page-bundle-specific breadcrumb service

```

PR #883, opened 2017-10-09
use page as context for breadcrumbs

rvanginneken:
I am targeting this branch, because breadcrumbs are not loaded correctly by default.

Relates to #751

## Changelog

```markdown
### Changed
- use 'page' instead of 'homepage' as default context for breadcrumbs
```

## Subject

As mentioned in #751, breadcrumbs are not loaded correctly. In that PR is alot of complexity, which I personally don't see the purpose of, but this may be related to the way I set up my project.
It appears that simply using the 'page' context will create correct breadcrumbs, pointing to the correct localised pages.

### DIFF
--- a/Resources/views/layout.html.twig
+++ b/Resources/views/layout.html.twig
@@ -19,10 +19,7 @@ file that was distributed with this source code.
 
             {% block sonata_page_breadcrumb %}
                 <div class="row">
-                    {% if sonata_seo_context is not defined %}
-                        {% set sonata_seo_context = 'homepage' %}
-                    {% endif %}
-                    {{ sonata_block_render_event('breadcrumb', { 'context': sonata_seo_context, 'current_uri': app.request.requestUri }) }}
+                    {{ sonata_block_render_event('breadcrumb', { 'context': 'page', 'current_uri': app.request.requestUri }) }}
                 </div>
             {% endblock %}
 


Resources/views/layout.html.twig
```diff
@@ -19,10 +19,7 @@ file that was distributed with this source code.
 
             {% block sonata_page_breadcrumb %}
                 <div class="row">
-                    {% if sonata_seo_context is not defined %}
-                        {% set sonata_seo_context = 'homepage' %}
-                    {% endif %}
-                    {{ sonata_block_render_event('breadcrumb', { 'context': sonata_seo_context, 'current_uri': app.request.requestUri }) }}
+                    {{ sonata_block_render_event('breadcrumb', { 'context': 'page', 'current_uri': app.request.requestUri }) }}
```
greg0ire: I don't really understand why we suddenly no longer need `sonata_seo_context`. Are there any docs about this variable?

rvanginneken: Hey @greg0ire,
I removed it because I didnt see any info about it in the docs. I also checked my project, which has all dependencies for the pagebundle installed; I couldn't find any usage of it besides the layout.html.twig.

greg0ire:
Thanks @rvanginneken ! If this breaks something, we'll revert it but make sure to add some docs ^^

rvanginneken:
Hey @greg0ire I'm not quite sure what I'm supposed to add docs about. There is something in the docs about creating breadcrumbs and using a custom context for that. "sonata_seo_context" is also not used there.

PR #884, opened 2017-10-15
Drop old versions of sf

jordisala1991:

## Changelog

No need for new changelog since it was already on a previous unreleased PR


jordisala1991:
This build is failing... super strange...

PR #886, opened 2017-10-21
Remove code checks for <sf2.8

acrobat:
I am targeting this branch, because of code cleanup after sf/php bump.

## Changelog

```markdown
### Removed
- pre sf2.8 code checks
```

## Subject

Some code cleanup after the symfony/php version bump

### DIFF
--- a/CmsManager/CmsManagerSelector.php
+++ b/CmsManager/CmsManagerSelector.php
@@ -15,7 +15,6 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
-use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
 use Symfony\Component\Security\Http\Logout\LogoutHandlerInterface;
@@ -79,7 +78,7 @@ class CmsManagerSelector implements CmsManagerSelectorInterface, LogoutHandlerIn
      */
     public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
     {
-        if ($this->getSecurityContext()->getToken() &&
+        if ($this->container->get('security.token_storage')->getToken() &&
             $this->container->get('sonata.page.admin.page')->isGranted('EDIT')) {
             $this->getSession()->set('sonata/page/isEditor', true);
         }
@@ -118,18 +117,5 @@ class CmsManagerSelector implements CmsManagerSelectorInterface, LogoutHandlerIn
         }
 
         return;
-    }
-
-    /**
-     * @return TokenStorageInterface
-     */
-    private function getSecurityContext()
-    {
-        // NEXT_MAJOR: Remove hack when bumping requirements to SF 2.6+
-        if (interface_exists('Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface')) {
-            return $this->container->get('security.token_storage');
-        }
-
-        return $this->container->get('security.context');
     }
 }
--- a/Controller/PageController.php
+++ b/Controller/PageController.php
@@ -68,7 +68,7 @@ class PageController extends Controller
 
         $cms->setCurrentPage($page);
 
-        // NEXT_MAJOR: remove the usage of $this->getRequest() by injecting the request action method (sf 2.4+)
+        // NEXT_MAJOR: remove the usage of $this->getRequest() by injecting the request action method in 4.0 (BC break)
         return $this->getPageServiceManager()->execute($page, $this->getRequest());
     }
 
@@ -109,10 +109,6 @@ class PageController extends Controller
      */
     private function getRequest()
     {
-        if ($this->container->has('request_stack')) {
-            return $this->container->get('request_stack')->getCurrentRequest();
-        }
-
-        return $this->container->get('request');
+        return $this->container->get('request_stack')->getCurrentRequest();
     }
 }
--- a/Tests/Validator/UniqueUrlValidatorTest.php
+++ b/Tests/Validator/UniqueUrlValidatorTest.php
@@ -14,6 +14,7 @@ namespace Sonata\PageBundle\Tests\Validator;
 use PHPUnit\Framework\TestCase;
 use Sonata\PageBundle\Validator\Constraints\UniqueUrl;
 use Sonata\PageBundle\Validator\UniqueUrlValidator;
+use Symfony\Component\Validator\Context\ExecutionContext;
 
 class UniqueUrlValidatorTest extends TestCase
 {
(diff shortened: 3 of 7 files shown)


Controller/PageController.php
```diff
@@ -52,7 +52,7 @@ public function exceptionsListAction()
      *
      * @return Response
      */
-    public function exceptionEditAction($code)
+    public function exceptionEditAction(Request $request, $code)
```
greg0ire: This is a BC-break

acrobat: Why is it a BC break? Symfony will resolve the typehinted parameter and the `$code` parameter will be mathed to the route parameter

greg0ire: > Or do you mean that the controller can possibly be extended?

Yes

acrobat: Or should I change this to `public function exceptionEditAction($code, Request $request = null)` and when this is null use the old method to get the request + deprecration when null? This way you can remove the BC method in the next major release and make the request parameter required

greg0ire: Nope: https://3v4l.org/Wmd05

acrobat: oh snap indeed :stuck_out_tongue:

acrobat: Fixed!

acrobat:
Removed the BC break!

acrobat:
Tests are failing now but I will take a look what's wrong tomorrow!

CmsManager/CmsManagerSelector.php
```diff
@@ -79,7 +78,7 @@ public function isEditor()
      */
     public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
     {
-        if ($this->getSecurityContext()->getToken() &&
+        if ($this->container->get('security.token_storage')->getToken() &&
```
Soullivaneuh: It would be great to start using dependency injection instead of using this ugly container.

Tests/Validator/UniqueUrlValidatorTest.php
```diff
@@ -111,10 +111,6 @@ public function testValidateWithPageDynamic()
 
     private function getContext()
     {
-        return $this->createMock(
-            class_exists('Symfony\Component\Validator\Context\ExecutionContextInterface') ?
-            'Symfony\Component\Validator\Context\ExecutionContextInterface' :
-            'Symfony\Component\Validator\ExecutionContextInterface'
-        );
+        return $this->createMock('Symfony\Component\Validator\Context\ExecutionContextInterface');
```
Soullivaneuh: `ExecutionContextInterface::class`

CmsManager/CmsManagerSelector.php
acrobat: good point! I will take a look at it!

Tests/Validator/UniqueUrlValidatorTest.php
OskarStark: @Soullivaneuh isn't there a fixer available which we could activate?

CmsManager/CmsManagerSelector.php
acrobat: @Soullivaneuh is it considered a BC break if I add extra arguments to this service? Or is this allowed?

Tests/Validator/UniqueUrlValidatorTest.php
acrobat: Fixed! And tests pass now

Tests/Validator/UniqueUrlValidatorTest.php
```diff
@@ -111,10 +112,19 @@ public function testValidateWithPageDynamic()
...
+        $context = new ExecutionContext($validator, 'root', $translator);
+
+        $context->setGroup('MyGroup');
+        $context->setNode('InvalidValue', null, null, 'property.path');
+        $context->setConstraint(new UniqueUrl());
+        $validator->expects($this->any())
+            ->method('inContext')
+            ->with($context)
+            ->will($this->returnValue($contextualValidator));
+        return $context;
```
acrobat: I've copied this initialization code from the symfony validator tests

acrobat:
Do I need to fix the styleci error in this PR too? As these are "errors" that were not introduced in this PR

Soullivaneuh:
@acrobat No, it's an external change.

Please wait #892 to be merged and do a rebase.

acrobat:
@Soullivaneuh rebase done!

CmsManager/CmsManagerSelector.php
Soullivaneuh: Well, maybe adding the service with `null` value as default and trigger a deprecated error if it's not provided?

But maybe it's quite overkill as this class should be only use as a service.

cc @sonata-project/contributors for more opinions.

Soullivaneuh: Well, maybe add the argument with `null` value as default and trigger a deprecation error if not provided?

Maybe quite overkill. @sonata-project/contributors toughs?

acrobat: Maybe we should move the "service container cleanup" to a separate pr?

greg0ire: I would remove the type hint on container, remove the argument, and throw if the provided argument is not a token storage. Also, I'd add a next major instruction asking for a type hint.

acrobat: If it's ok for you guys I will create a new issue to do this cleanup later on?

greg0ire: It is for me

acrobat: Let's do that, smaller PR's are easier to review and merge

Soullivaneuh: > I would remove the type hint on container, remove the argument, and throw if the provided argument is not a token storage. Also, I'd add a next major instruction asking for a type hint.

As this class only needs the `security.token_storage` service, we can indeed do that. :+1:

acrobat: It does need more services no? There are a lot `$this->container->get('')` calls in this service

greg0ire: I didn't see that, but indeed, you're right. Maybe add one argument per service, without type hinting, and make them optional, and then play with `func_get_args` /  `func_num_args` to simulate 2 different signatures?

acrobat: Yeah I'm looking at symfony how they do it now, adding/removing arguments. But indeed it will be something with `func_num_args`/`func_get_args`. I will come up with something and open a PR for it!

Tests/Validator/UniqueUrlValidatorTest.php
```diff
@@ -111,10 +112,20 @@ public function testValidateWithPageDynamic()
 
     private function getContext()
     {
-        return $this->createMock(
-            class_exists('Symfony\Component\Validator\Context\ExecutionContextInterface') ?
-            'Symfony\Component\Validator\Context\ExecutionContextInterface' :
-            'Symfony\Component\Validator\ExecutionContextInterface'
-        );
+        $translator = $this->getMockBuilder('Symfony\Component\Translation\TranslatorInterface')->getMock();
```
OskarStark: couldn't we use `createMock()` here?

acrobat: yes indeed, will replace those

acrobat: Done!

greg0ire:
Thanks @acrobat !

PR #895, opened 2017-11-05
DevKit updates for 3.x branch

SonataCI:


Tests/Route/CmsPageRouterTest.php
```diff
@@ -163,6 +163,9 @@ public function testGenerateWithPageSlugValid()
         $this->assertEquals('//localhost/my/path?foo=bar', $url);
     }
 
+    /**
+     * @group legacy
+     */
```
greg0ire: @pix-art please add tests for the new way.

core23:
This should fix travis: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/750

CONTRIBUTING.md
```diff
@@ -74,7 +74,7 @@ php-cs-fixer fix --verbose
 
 #### The documentation
 
-The documentation is mostly written with the `rst` format, and can be found in the `docs` directory.
+The documentation is mostly written with the `rst` format, and can be found in the `Resources/doc` directory.
```
core23: We should fix this on the dev-kit

core23:
It looks like that the `weak_vendors` env entry has some problems when mocking `getMockForAbstractClass(Doctrine\ORM\QueryBuilder::class)`.

The class has has some `@deprecated` methods which causes the warnings. After removing the annotation, everything is fine.

cc @greg0ire

greg0ire:
@core23 This might fix it: https://github.com/symfony/symfony/pull/24864

core23:
So we must wait for phpunit bridge 3.4 to fix this issue?

core23:
Maybe we should revert the `weak_vendors` mode.

greg0ire:
Or should we mark the test with `@legacy`?

core23:
This is not the only project that will fail. So we should revert his mode for now.

greg0ire:
Yeah probably :(

PR #905, opened 2017-11-14
Fix Twig 2.0 check for block 'content' existence

psyray:
This PR is complementary of this one #860 
While accessing a page with empty content the following error is returned if using Twig 2.0
image

## Changelog
```markdown
### Fixed
- compatibility with Twig 2.0 was improved
```


OskarStark:
Thank you @psyray 👍

PR #913, opened 2017-11-30
Fixed wrong route in pagelist block

core23:

## Changelog

```markdown
### Fixed
 - Fixed wrong route in pagelist block
```

## Subject

The wrong route generator was called.


greg0ire:
Thanks @core23 !

PR #880, opened 2017-10-09
Make it possible to use Page SEO title again

rvanginneken:
I am targeting this branch, because seo title was not working.

Related to #767 
Closes #800

## Changelog

```markdown
### Fixed
- Not working SEO page title
```

## Subject

All pages were showing only the default title. SEO title was not used. Now we override it when a title or name is set.

PR #916, opened 2017-12-01
Improved seo vs. page metadata

core23:
I am targeting this branch, because this is a bugfix (again)

## Changelog

```markdown
### Changed
 - The internal page name is not used as a seo title fallback anymore
```

## Subject

Shame on me, #880 introduced this stupid bug again. To be safe for the future, I added a short comment why this line should be keept.

### DIFF
--- a/src/Page/Service/DefaultPageService.php
+++ b/src/Page/Service/DefaultPageService.php
@@ -71,8 +71,12 @@ class DefaultPageService extends BasePageService
             return;
         }
 
-        if ($page->getTitle() || $page->getName()) {
-            $this->seoPage->setTitle($page->getTitle() ?: $page->getName());
+        /*
+         * Always prefer the page title, if set.
+         * Do not use the (internal) page name as a fallback
+         */
+        if ($page->getTitle()) {
+            $this->seoPage->setTitle($page->getTitle());
         }
 
         if ($page->getMetaDescription()) {
--- a/tests/Page/Service/DefaultPageServiceTest.php
+++ b/tests/Page/Service/DefaultPageServiceTest.php
@@ -61,7 +61,7 @@ class DefaultPageServiceTest extends TestCase
 
         // mock a page instance
         $page = $this->createMock('Sonata\PageBundle\Model\PageInterface');
-        $page->expects($this->exactly(2))->method('getTitle')->will($this->returnValue('page title'));
+        $page->expects($this->any())->method('getTitle')->will($this->returnValue('page title'));
         $page->expects($this->atLeastOnce())->method('getMetaDescription')->will($this->returnValue('page meta description'));
         $page->expects($this->atLeastOnce())->method('getMetaKeyword')->will($this->returnValue('page meta keywords'));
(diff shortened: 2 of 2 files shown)


src/Page/Service/DefaultPageService.php
```diff
@@ -71,7 +71,9 @@ protected function updateSeoPage(PageInterface $page)
             return;
         }
 
-        if ($page->getTitle() || $page->getName()) {
+        // Do not change this line if you really know what you are doing!
+        // The page name is technically, so always prefer the title
```
greg0ire: The first sentence means the opposite of what you mean, and I don't understand the second one

src/Page/Service/DefaultPageService.php
```diff
@@ -71,7 +71,9 @@ protected function updateSeoPage(PageInterface $page)
             return;
         }
 
-        if ($page->getTitle() || $page->getName()) {
+        // Only change this line if you really know what you are doing!
+        if (!$this->seoPage->getTitle()) {
+            // Prefer the page title, fallback to the technically page name
```
greg0ire: "technical page name"?

core23:
Please don't merge! 

This problem looks more complex!

core23:
Everything should be fine now @OskarStark

core23:
The old implementation will never take the seo title set by the seo extension (e.g. https://github.com/sonata-project/SonataNewsBundle/blob/3.x/src/Controller/PostController.php#L180) because the page name could not be empty and is always required.

src/Page/Service/DefaultPageService.php
```diff
@@ -71,8 +71,11 @@ protected function updateSeoPage(PageInterface $page)
             return;
         }
 
-        if ($page->getTitle() || $page->getName()) {
-            $this->seoPage->setTitle($page->getTitle() ?: $page->getName());
+        // Only change this line if you really know what you are doing!
+        // Always prefer the page title, if set
+        // Do not use the (internal) page name as a fallback
```
greg0ire: Please use multi-line style comments

core23:
cc @rvanginneken

rvanginneken:
If I remember correct, the previous implementation checked the seo title, which would always be set. I think I just reverted it back to _some_ older implementation. If this fix still gives the desired results, great.

As for 
> Only change this line if you really know what you are doing!

I don't see no reason for adding this, besides demotivating creating PR's.

core23:
> I don't see a reason for adding this, besides demotivating creating PR's. This could be added literally everywhere.

As far as I remembered, there were at least 3 PRs which try to change the behavior.

jordisala1991:
I wont say anything about having a comment on the code if it is really important or it is hard to understand why the line is there. But I would also prefer a better message explaining why that line is important and/or why it has to be that way. WDYT @core23 ?

core23:
Reworded the comment

greg0ire:
Reminds me of this: https://stackoverflow.com/questions/184618/what-is-the-best-comment-in-source-code-you-have-ever-encountered#482129

PR #925, opened 2017-12-09
Make services public

core23:

## Changelog

```markdown
### Changed
 - make services explicit public
```

## Subject

This patch is needed for sf 4.


greg0ire:
> This patch is needed for sf 4.

Maybe not: https://github.com/sonata-project/SonataAdminBundle/pull/4813/files

jordisala1991:
Maybe not the admin , but the others for sure. Other than that, isnt it better to make it explicit?

greg0ire:
It surely doesn't hurt

PR #930, opened 2017-12-12
Fixing type unsafe parameters being passed to some functions

janpetersen:
I am targeting this branch, because I use the 'master' branch in my current project.

## Changelog

```markdown
### Fixed
- Implementing type safe method call to DateTime::setTimestamp()
- Fixing not type safe parameters in calls to string functions
```

## Subject

I fixed two things that occured just as of today after updating 'master' via composer update and after updating Php to 7.1.12 (from 7.1.11; on Ubuntu 14.04 LTS). Not sure which one caused the issues, but my changes sure fixes them.

_Edit:_ Commit https://github.com/sonata-project/SonataPageBundle/commit/06ba25115ae65236a0fe072aec90cbf8df2971c2 introduced these errors, since "strict types" were declared.

**1) Implementing type safe method call to DateTime::setTimestamp()**

When loading serialized snapshot contents, the Transformer passed string values to the \DateTime::setTimestamp()" method, resulting in an uncaught FatalThrowableError:

> Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalThrowableError: "Type error: DateTime::setTimestamp() expects parameter 1 to be integer, string given" at Transformer.php line 165
> DateTime->setTimestamp('1510846884')
> in vendor/sonata-project/page-bundle/src/Entity/Transformer.php (line 165)

**2) Fixing not type safe parameters in calls to string functions**

There were two FatalThrowableErrors when passing null values to php's string functions "trim" and "substr":

> Type error: trim() expects parameter 1 to be string, null given
> in vendor/sonata-project/page-bundle/src/Model/Page.php (line 311)

> Type error: substr() expects parameter 1 to be string, null given
> in vendor/sonata-project/page-bundle/src/Model/Page.php (line 243)


janpetersen:
I just  found out why the errors occured:

Four days ago the "strict types" declarations were added:

> declare(strict_types=1);

https://github.com/sonata-project/SonataPageBundle/commit/06ba25115ae65236a0fe072aec90cbf8df2971c2

src/Model/Page.php
```diff
@@ -308,6 +308,11 @@ public function getName()
      */
     public function setSlug($slug): void
     {
+        if ($slug === null) {
+            $this->slug = null;
+            return;
+        }
+        
         $this->slug = self::slugify(trim($slug));
```
jordisala1991: isn't it better to simply do: `self::slugify(trim((string) $slug));`

src/Entity/Transformer.php
```diff
@@ -162,11 +162,11 @@ public function load(SnapshotInterface $snapshot)
         $page->setRequestMethod($content['request_method']);
 
         $createdAt = new \DateTime();
-        $createdAt->setTimestamp($content['created_at']);
+        $createdAt->setTimestamp((int) $content['created_at']);
```
jordisala1991: For this changes, isn't it better to modify instead this other code?

https://github.com/sonata-project/SonataPageBundle/blob/4f9b962f2bb5d44175bb33e1aefc3106fbec6d59/src/Entity/Transformer.php#L111

We could do instead of `format` use `getTimestamp` and we would get a integer without needing a conversion here. WDYT?

src/Model/Page.php
```diff
@@ -240,7 +240,7 @@ public function getRouteName()
      */
     public function setPageAlias($pageAlias): void
     {
-        if ('_page_alias_' != substr($pageAlias, 0, 12)) {
+        if ($pageAlias !== null && '_page_alias_' != substr($pageAlias, 0, 12)) {
```
jordisala1991: ins't is better to simply cast to string?

src/Entity/Transformer.php
janpetersen: This would store all new entries as integers, but it would not fix all the existing entries in the database which are serialized at strings.

Perhaps we could do both. :)

src/Model/Page.php
janpetersen: I don't know if it's better or not. It depends on whether you want to store NULL or '\_page\_alias\_' into the object's property.

That's something _you_ should decide, since you know the framework way better than me. :)

src/Model/Page.php
janpetersen: This depends on whether you want to store NULL or an empty string into the object.

You decide! :)

jordisala1991: https://3v4l.org/i2Ito

Seems like cast to string preserves better the old code

janpetersen:
Hi, I changed the code, it now casts to string.

Feel free to review and merge! :)

jordisala1991:
ping @janpetersen can you do a rebase on master?

janpetersen:
@jordisala1991 I updated my branch to upstream/master.

jordisala1991:
This is already fixed on 3.x and tonight will be updated on master by an automatic merge. Thank you @janpetersen

janpetersen:
Okay thank you @jordisala1991 ! :)

PR #940, opened 2018-01-03
Update installation.rst

AntonioCS:
Duplicated configuration

Closes #{put_issue_number_here}

## Changelog

```markdown
### Added
- Added some `Class::newMethod` to do great stuff

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -153,11 +153,6 @@ configuration file.
             page_defaults:
                 homepage: {decorate: false} # disable decoration for homepage, key - is a page route
 
-            default_template: default # template key from templates section, used as default for pages
-            templates:
-                default:  { path: 'SonataPageBundle::layout.html.twig',          name: 'default' }
-                2columns: { path: 'SonataPageBundle::2columns_layout.html.twig', name: '2 columns layout' }
-
             # manage the http errors
             catch_exceptions:
                 not_found: [404]    # render 404 page with "not_found" key (name generated: _page_internal_error_{key})


greg0ire:
I remove the changelog, it is not needed for docs.

AntonioCS:
Sure, makes sense.

greg0ire:
Copying the most important part for people who want to understand this:

> @greg0ire I reported a bug in Sonata Admin which originates in Sonata Page documentation so please forgive me obvious inclination to blame the documentation.
BUT reexamining the documentation you can see this:

```
sonata_page:
    slugify_service:   sonata.core.slugify.cocur # old BC value is sonata.core.slugify.native
    multisite:        host
    default_template: default # template key from templates section, used as default for new pages
    templates:
        default:  { path: 'SonataPageBundle::layout.html.twig',          name: 'default' }
        2columns: { path: 'SonataPageBundle::2columns_layout.html.twig', name: '2 columns layout' }

    # Generates a snapshot when a page is saved (from the admin)
    direct_publication: false # or %kernel.debug% if you want to publish in dev mode (but not in prod)
```

> and later on we have similar documentation:
```

sonata_page:
    multisite: host
    use_streamed_response: true # set the value to false in debug mode or if the reverse proxy does not handle streamed response
    ignore_route_patterns:
        - ^(.*)admin(.*)   # ignore admin route, ie route containing 'admin'
        - ^_(.*)          # ignore symfony routes

    ignore_routes:
        - sonata_page_cache_esi
        - sonata_page_cache_ssi
        - sonata_page_js_sync_cache
        - sonata_page_js_async_cache
        - sonata_cache_esi
        - sonata_cache_ssi
        - sonata_cache_js_async
        - sonata_cache_js_sync
        - sonata_cache_apc

    ignore_uri_patterns:
        - ^/admin\/   # ignore admin route, ie route containing 'admin'

    page_defaults:
        homepage: {decorate: false} # disable decoration for homepage, key - is a page route

    default_template: default # template key from templates section, used as default for pages
    templates:
        default:  { path: 'SonataPageBundle::layout.html.twig',          name: 'default' }
        2columns: { path: 'SonataPageBundle::2columns_layout.html.twig', name: '2 columns layout' }

    # manage the http errors
    catch_exceptions:
        not_found: [404]    # render 404 page with "not_found" key (name generated: _page_internal_error_{key})
        fatal:     [500]    # so you can use the same page for different http errors or specify specific page for each error
```

> On this [page](https://sonata-project.org/bundles/page/master/doc/reference/installation.html)
On the first piece of configuration you do get the warning that you will need to overwrite this but not for the second piece and you can see that you have the **templates** part again.

> So what happened is that I didn't notice this and just copied **your** sample of the documentation.
I did remove the first section of templates but not the second and then I just added what is on this [page](https://sonata-project.org/bundles/page/master/doc/reference/page_composer.html).

docs/reference/installation.rst
```diff
@@ -153,11 +153,6 @@ configuration file.
             page_defaults:
                 homepage: {decorate: false} # disable decoration for homepage, key - is a page route
 
-            default_template: default # template key from templates section, used as default for pages
-            templates:
-                default:  { path: 'SonataPageBundle::layout.html.twig',          name: 'default' }
-                2columns: { path: 'SonataPageBundle::2columns_layout.html.twig', name: '2 columns layout' }
-
```
greg0ire: Not sure if that is the right fix… what if I don't use the page Composer? Will things work without this? What do you think of adding a warning on top of [this page](https://sonata-project.org/bundles/page/master/doc/reference/page_composer.html) instead?

greg0ire: The page Composer doc is just a code block, adding a nice introduction to it would make sense IMO.

AntonioCS: Yeah I just removed the part that I thought would make sense, since it's the part that is duplicated but since I am still new to this bundle this might affect other parts that I am not aware. 
The linked page also has issues:
- Missing sonata_page section
- The template files don't exist, the default and 2columns can be changed to the ones provided by sonata but not the 3columns
- There is no description (that I can see) of what the sections in the config actually are.

Might be more, but again, with my limited knowledge of the project those are the ones I could point out and got bitten by in my installation of the sonata page bundle.

greg0ire: I think the `sonata_page` is missing because it is "obvious" and that would imply some indent, but maybe this was a bad idea? Also, in Symfony 4, this would be in `sonata_page.yml` I guess, so no indention anymore.

I think with the limited knowledge you have yet, just a warning on top of the Composer file should be the less risky, right? Bear in mind that I don't use this bundle, so, I have no more knowledge than you do :P

AntonioCS: Is there someone here that does use the bundle? I ask because I don't understand:

> … what if I don't use the page Composer? 

Don't you just ignore (in the config) the routes that you don't want the bundle to touch?

greg0ire: I'm not sure if it makes sense, bug if you don't use the page Composer, you won't even read the doc about it, and that means you will not have the block you just removed (should this PR be merged). Will things still work in that case?

AntonioCS: I am not sure I fully understand. The section I am removing is mentioned earlier in the documentation page. I am just trying to prevent a repetition here.

greg0ire: People might think they have to replace the whole section with the second code sample. Especially if they read this (emphasis is mine):

> we need to add some lines which we will *override* later (we need them now only for the following step):

Plus if we follow your logic, we'd have to remove `multisite: host` too, correct?

AntonioCS: Yeah, I guess any trace of duplication would be better removed

PR #914, opened 2017-11-30
Removed deprecated inline editor

core23:
I am targeting this branch, because this is a BC break.

## Changelog

```markdown
### Removed
 - Removed deprecated inline editor
```

## Subject

Inline editing is deprecated since version 2.3.

PR #949, opened 2018-01-25
bugfix: create new container position corrected

n3o77:
I am targeting this branch, because it's a bugfix.

## Changelog

```markdown
### Fixed
- Add new block to container position starts with 0.
```
## Subject

When adding a new block in a container it uses the length of all children (including itself) which leads to an offset of 1. If you use the drag & drop sorting the count is one less as it starts with 0. This is usually not a big problem as you just have a gap in the positions but the sorting still works fine, but if you depend on the position in the template or controller this can lead to unexpected behavior. 

I.E. you want to give the first child in a container an extra class. If you just add the items (without sorting) the position starts at 1. If you use the drag & drop sort feature the position starts at 0.

### DIFF
--- a/src/Resources/public/sonata-page.back.js
+++ b/src/Resources/public/sonata-page.back.js
@@ -362,7 +362,7 @@
                     $parentFormControl.parent().parent().hide();
                 } else if (self.isFormControlTypeByName(formControlName, 'position')) {
                     $positionFormControl = $formControl;
-                    $positionFormControl.val($containerChildren.find('> *').length);
+                    $positionFormControl.val($containerChildren.find('> *').length - 1);
                     $positionFormControl.closest('.form-group').hide();
                 }
             });
--- a/assets/src/js/composer.js
+++ b/assets/src/js/composer.js
@@ -354,7 +354,7 @@
                     $parentFormControl.parent().parent().hide();
                 } else if (self.isFormControlTypeByName(formControlName, 'position')) {
                     $positionFormControl = $formControl;
-                    $positionFormControl.val($containerChildren.find('> *').length);
+                    $positionFormControl.val($containerChildren.find('> *').length - 1);
                     $positionFormControl.closest('.form-group').hide();
                 }
             });


n3o77:
How do I handle the compiled files? Just realised I need to commit them too. As mentioned in #914 the output of the files isn't optimal. Should I just adjust them manually?

greg0ire:
Hmmm… maybe this helps: https://github.com/sonata-project/dev-kit/pull/213/files#r140774356 ?

n3o77:
Sorry, not really. I know how to compile them. The problem is that the output seems not to be OK, there is a slightly different banner format, different formatting and no newline at the end. I can adjust that manually, but it doesn't seem to make sense to manually adjust generated files. 

In this [PR](https://github.com/sonata-project/SonataPageBundle/pull/839/files) it looks like it was done manually.

greg0ire:
Please push the generated files as is so that we can see what this means. Maybe we just need to adjust some configuration. How do you generate the files, by the way?

n3o77:
Pushed them, but they basically look the same as in #914 . `npm install` or `yarn install` in the `assets_src` folder and then just run `gulp` (or `gulp js` etc.).

greg0ire:
Ok so regarding the header, it's automatically generated, and prepended to the one of the original file. I think it should be removed from the build

src/Resources/public/sonata-page.back.js
```diff
@@ -571,32 +586,36 @@
...
-                        enabled = !enabled;
-                        $switchButton.toggleClass('bg-yellow bg-green');
-                        $switchBtnIcon.toggleClass('fa-toggle-off fa-toggle-on');
-
-                        if (enabled) {
-                            $switchButton.html($switchLblDsbl);
-                        } else {
-                            $switchButton.html($switchLblEnbl);
-                        }
+                        if (resp.status && resp.status === 'OK') {
```
greg0ire: That does not look like "just a formatting change"

greg0ire:
This looks like a big mess. I think it should be handled in a separate PR, and you could do your change manually for now.

src/Resources/public/sonata-page.back.js
n3o77: You're right. Looks like it was introduced here (from blame); https://github.com/sonata-project/SonataPageBundle/commit/c2112c0ed4f7278cb38d629c153ca0296f33529f

n3o77:
Yeah seems messy. Especially as the file contents don't seem to match up. The header usually should replace the existing one but I guess it doesn't work because the format is slightly different.

greg0ire:
The mess has been handled

n3o77:
@greg0ire what's the proper way now to handle the compiled js files?

greg0ire:
@covex-nn can you answer this please?

covex-nn:
@greg0ire @n3o77 i guess, there is only one way: rebase, resolve conflicts only in `src/Resources/assets_src` (if any), execute `cd src/Resources/assets_src && gulp`. Then push changes in `assets_src` and `public` dirs

n3o77:
@covex-nn thanks

OskarStark:
@n3o77 would be great if you could finish this PR!

Thank you

n3o77:
@OskarStark i rebased and removed the compiled JS file. That should be it, right?

OskarStark:
I merged #1027 , please rebase again

n3o77:
@OskarStark thank you. rebased and pushed

OskarStark:
Thank you!

greg0ire:
Thanks @n3o77 !

PR #952, opened 2018-02-06
Changed default templates for system pages

core23:

## Changelog

```markdown
### Changed
 - Remove default template from exception list
 - Use default template in page create template

```

## Subject

The old address won't work in with newer twig versions.

### DIFF
--- a/src/Resources/views/Exceptions/list.html.twig
+++ b/src/Resources/views/Exceptions/list.html.twig
@@ -8,8 +8,6 @@ For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
 
 #}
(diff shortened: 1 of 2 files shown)


src/Resources/views/Exceptions/list.html.twig
```diff
@@ -8,8 +8,6 @@ For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
 
 #}
-{% extends "::base.html.twig" %}
```
core23: We don't need a template for this special page.

core23:
Please review again @greg0ire

greg0ire:
Thanks @core23 !

PR #957, opened 2018-02-11
Minor composer improvements

jordisala1991:

## Subject

- Replace security by the specific components we need
- Up some dependencies to avoid long composer resolution on --prefer-lowest


jordisala1991:
Seems like there is not 1.x release with php 5.6 compatibility since 1.0.7 (we should have had one before releasing 2.0) so moving back to 1.0.2.

greg0ire:
Thanks @jordisala1991 !

PR #961, opened 2018-02-15
Drop compatibility with FOSRest <2.1

jordisala1991:

## Changelog

```markdown
### Removed
- Removed compatibility with older versions of FOSRestBundle (<2.1)
```

## Subject

Since dropping a dependency is not a BC break, this is a good start, because we are doing a lot of code just to keep compat with FOSRest 1.


composer.json
```diff
@@ -51,13 +51,14 @@
         "twig/twig": "^1.34 || ^2.0"
     },
     "conflict": {
+        "friendsofsymfony/rest-bundle": "<2.1 || >=3.0",
```
core23: Do we need an upper limit?

jordisala1991: It was added on a bundle, and I add it on this one too. We had a discussion about that some time ago, and we decided to keep the upper limit on that other case

greg0ire:
Thanks @jordisala1991 !

Issue #965, opened 2018-02-20
site-install error: "Type error: trim() expects parameter 1 to be string, null given)"


ob1201:
For a Symfony4 install
```
doctrine/cache                           v1.7.1             = v1.7.1             Caching library offering an object-oriented API for many cache backends
doctrine/doctrine-cache-bundle           1.3.2              = 1.3.2              Symfony Bundle for Doctrine Cache
doctrine/phpcr-bundle                    dev-master bb380a5 = dev-master bb380a5 Symfony DoctrinePHPCRBundle
doctrine/phpcr-odm                       1.4.4              = 1.4.4              Object-Document-Mapper for PHPCR
friendsofsymfony/http-cache              2.1.2              = 2.1.2              Tools to manage HTTP caching proxies with PHP
jackalope/jackalope                      1.3.5              = 1.3.5              Jackalope PHPCR library
phpcr/phpcr                              2.1.4              = 2.1.4              PHP Content Repository interfaces
phpcr/phpcr-utils                        1.4.0              = 1.4.0              PHP Content Repository implementation independant utilities
psr/cache                                1.0.1              = 1.0.1              Common interface for caching libraries
psr/simple-cache                         1.0.0              = 1.0.0              Common interfaces for simple caching
sonata-project/admin-bundle              3.31.1             = 3.31.1             The missing Symfony Admin Generator
sonata-project/block-bundle              3.12.0             = 3.12.0             Symfony SonataBlockBundle
sonata-project/cache                     1.1.1              ~ 2.0.1              Cache library
sonata-project/cache-bundle              2.x-dev c0de243    ~ dev-master 4010ae3 This bundle provides caching services
sonata-project/core-bundle               3.x-dev a71caca    ~ dev-master 885ce54 Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.3.1              = 2.3.1              Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.0.2              = 1.0.2              Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.4.2              = 3.4.2              Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.5.0              = 2.5.0              Symfony SonataEasyExtendsBundle
sonata-project/exporter                  1.8.0              = 1.8.0              Lightweight Exporter library
[...]

PR #966, opened 2018-02-21
Update Page.php

ob1201:
I am targeting this branch, because fails on site load

Addresses #965

## Changelog
```markdown
### Fixed
missing string wrapper around trim
```

### DIFF
--- a/src/Model/Page.php
+++ b/src/Model/Page.php
@@ -306,7 +306,7 @@ abstract class Page implements PageInterface
      */
     public function setSlug($slug)
     {
-        $this->slug = self::slugify(trim($slug));
+        $this->slug = self::slugify(trim((string)$slug));
     }
 
     /**
--- a/src/Model/Page.php
+++ b/src/Model/Page.php
@@ -306,7 +306,7 @@ abstract class Page implements PageInterface
      */
     public function setSlug($slug)
     {
-        $this->slug = self::slugify(trim((string)$slug));
+        $this->slug = self::slugify(trim((string) $slug));
     }
 
     /**


greg0ire:

Bad example :

```
```

Good example:

```
Remove call to defraculate()

Fixes #989
```

src/Model/Page.php
```diff
@@ -306,7 +306,7 @@ public function getName()
      */
     public function setSlug($slug)
     {
-        $this->slug = self::slugify(trim($slug));
+        $this->slug = self::slugify(trim((string)$slug));
```
greg0ire: missing space between the cast and the variable, this is why the build fails. Please amend your commit.

ob1201: this is what you suggested
```
so the fix would be https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Model/Page.php#L309, adding (string) before $slug
```
that's simply what was done above, no?

kunicmarko20: @ob1201 you just need to add ` `  between `(string)` and `$slug` so you get `$this->slug = self::slugify(trim((string) $slug));`

src/Model/Page.php
```diff
@@ -306,7 +306,7 @@ public function getName()
      */
     public function setSlug($slug)
     {
-        $this->slug = self::slugify(trim($slug));
+        $this->slug = self::slugify(trim((string) $slug));
```
greg0ire: I hope nothing breaks if we change a slug from `null` to `''`. BTW, when does this occur? Can't see why the slug of a page would be `null` or `''`

jordisala1991: trim(null) = '' AFAIK

kunicmarko20: https://3v4l.org/71CT5 vs https://3v4l.org/ufpTl, not sure this is our problem?

jordisala1991: https://3v4l.org/3egP6

greg0ire:
Thanks @ob1201 !

greg0ire:
Wait what version of php are you using to get this warning? hhvm?

ob1201:
@greg0ire 
```
php -v
PHP 7.2.1-dev (cli) (built: Jan  8 2018 15:40:39) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
   with Zend OPcache v7.2.1-dev, Copyright (c) 1999-2018, by Zend Technologies
```

updating to this^^
```
composer require sonata-project/page-bundle:3.x-dev
```

_does_ fix the trim error, both at console, and on site frontpage -- where I'd been seeing it

greg0ire:
`7.2.1-dev`

You're using a dev branch of php?!

ob1201:
This is fully reproducible with 7.2.1 *release* as well

greg0ire:
Must have been fix in between then: 

```PHP 7.2.2 (cli) (built: Feb  6 2018 22:51:22) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.2, Copyright (c) 1999-2018, by Zend Technologies
bash-4.4$ php -r "var_dump(trim(null));"
string(0) ""

```

greg0ire:
Can't reproduce here either: https://3v4l.org/ufoMl am I doing the wrong test?

PR #968, opened 2018-02-22
Backport type fixes

greg0ire:

Closes #930 

## Changelog

```markdown
### Fixed
- forward-compatibility with strict mode
```

### DIFF
--- a/src/Model/Page.php
+++ b/src/Model/Page.php
@@ -238,7 +238,7 @@ abstract class Page implements PageInterface
      */
     public function setPageAlias($pageAlias)
     {
-        if ('_page_alias_' != substr($pageAlias, 0, 12)) {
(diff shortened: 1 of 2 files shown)


src/Model/Page.php
```diff
@@ -238,7 +238,7 @@ public function getRouteName()
      */
     public function setPageAlias($pageAlias)
     {
-        if ('_page_alias_' != substr($pageAlias, 0, 12)) {
+        if (null !== $pageAlias && '_page_alias_' != substr((string) $pageAlias, 0, 12)) {
```
jordisala1991: why this null comparison?

kunicmarko20: hm, you are right, not needed https://3v4l.org/fAJVX because of `(string)`

greg0ire: Yeah can't see how that would be useful

PR #973, opened 2018-03-07
Regenerate js and css files with gulp

covex-nn:
I am targeting this branch, because this is patch.

## Subject

Since #839 or even earlier, javascript and css files inside `src/Resources/public` directory was unsynchronized with a source in `src/Resources/assets_src`.

With this PR i synced them back, and also i removed duplicates of header comment with copyright at the beginning of a result files.

To check my changes: clone my fork, checkout `sync-gulp` branch and run `gulp` inside `src/Resources/assets_src` directory. And changes from #839 are not lost, because file `src/Resources/public/sonata-page.back.js` has not been changed (except header comment)


jordisala1991:
labelled as pedantic since it change mostly the source code of javascript, it does not change the compiled ones. (except for some comments)

src/Resources/assets_src/src/js/composer.js
```diff
@@ -1,14 +1,6 @@
 /**
- *
- * This file is part of the Sonata package.
```
OskarStark: is it correct, that all headers get removed?

covex-nn: @OskarStark i tried to leave only one header comment for files in `Resources/public` directory after running `gulp`. So, i removed all comments about `This file is part of the Sonata` in `Resources/assets_src/src` and modified `Resources/assets_src/gulp/util/banner.js` file.

PR #980, opened 2018-04-26
Require sonata-project/doctrine-orm-admin-bundle

covex-nn:
I am targeting this branch, because this a patch.

See https://github.com/sonata-project/dev-kit/issues/409

## Subject

SonataPageBundle supports only DoctrineORM. This PR will allow to install fully functional AdminBundle with DoctrineORM support after `composer require sonata-project/page-bundle`. And it will be still possible to create a recipe for Flex if this PR won't be accepted, these changes are optional

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "sonata-project/core-bundle": "^3.9",
         "sonata-project/datagrid-bundle": "^2.3",
         "sonata-project/doctrine-extensions": "^1.0",
+        "sonata-project/doctrine-orm-admin-bundle": "^3.5",
         "sonata-project/easy-extends-bundle": "^2.5",
         "sonata-project/notification-bundle": "^3.3",
         "sonata-project/seo-bundle": "^2.5",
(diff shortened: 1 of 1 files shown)


kunicmarko20:
I think this doesn't need changelog, am I right?

covex-nn:
It does not require release, i guess. But, with this PR and with Flex recipe SonataPageBundle could be installed and be fully fuctional just with:

```bash
composer create-project symfony/skeleton .
composer require sonata-project/page-bundle
```

So, lets label it with `pedantic` =) But, lets release it, if it will be accepted

greg0ire:
Uuuuh why do we add this dependency? Isn't this bundle usable with other storage backends?

covex-nn:
@greg0ire there is no Document or PHPCR classes and xml mappings, only Entity for DoctrineORM

greg0ire:
Please put this in the commit message then.

greg0ire:
Thanks! I squash merged and changed the commit message so that the subject is under 50 chars.

PR #992, opened 2018-06-11
#590 duplicate root route when it in config

### LINKED ISSUE #590: Duplicate / route when running sonata:page:update-core-routes command
Hello!

I work with PageBundle on 2.4.x-dev and first created a new Site and afterwards I run the command 

```
app/console sonata:page:update-core-routes --site=1
```

Now there are 2 pages as homepage created.
The first one with route_name page_slug and path /
the second one with route_name homepage and path /

If I try to edit one of the two homepages, I will get an error, that / already exists.

I found this solution: Delete the second one in the database and change the route_name to homepage. 

This is the part, which is wrong imo.
https://github.com/sonata-project/SonataPageBundle/blob/master/Route/RoutePageGenerator.php#L90

Can someone reproduce the problem?

greetings from munich,
Sebastian

moshenskyDV:
I am targeting this branch, because its patch.

Closes #590
Code is not my idea, thank @grimpows

## Changelog

Updated RoutePageGenerator. Now, homepage creates statically, and it stay as root page for another config pages by default. If  we will create in config page with root path ("/") - they same as other will child page (it means it will available by two slashes path ("example.com//"). 
This patch, before creating homepage looks in config page and search any page with root path. If it will found - we will use that config page as root.
So, after patch, possible configure bundle for homepage in routing file.

### Fixed

Creating homepage from router config (check if page available in config - we will not create default Homepage.

### Security

I checked this on my project. Looks good. Checked generating new routes and update previously routes.
Please double check this code.

### DIFF
--- a/src/Route/RoutePageGenerator.php
+++ b/src/Route/RoutePageGenerator.php
@@ -91,16 +91,7 @@ class RoutePageGenerator
 
         // no root url for the given website, create one
         if (!$root) {
-            $root = $this->pageManager->create([
-                'routeName' => PageInterface::PAGE_ROUTE_CMS_NAME,
-                'name' => 'Homepage',
-                'url' => '/',
-                'site' => $site,
-                'requestMethod' => $requirements['_method'] ??
-                    'GET|POST|HEAD|DELETE|PUT',
-                'slug' => '/',
-            ]);
-
+            $root = $this->createRootPage($site);
             $this->pageManager->save($root);
         }
 
@@ -247,5 +238,38 @@ MSG
         if ($output instanceof OutputInterface) {
             $output->writeln($message);
         }
+    }
+
+    /**
+     * Generate root page (parent for another generated page).
+     * If root path available in config - create from config
+     * Else - generate it statically.
+     */
+    private function createRootPage(SiteInterface $site): PageInterface
+    {
+        foreach ($this->router->getRouteCollection()->all() as $name => $route) {
+            if ('/' === $route->getPath()) {
+                $requirements = $route->getRequirements();
+                $name = trim($name);
+
+                return $this->pageManager->create([
+                    'routeName' => $name,
+                    'name' => $name,
+                    'url' => $route->getPath(),
+                    'site' => $site,
+                    'requestMethod' => $requirements['_method'] ?? 'GET|POST|HEAD|DELETE|PUT',
+                    'slug' => '/',
+                ]);
+            }
+        }
+
+        return $this->pageManager->create([
+            'routeName' => PageInterface::PAGE_ROUTE_CMS_NAME,
+            'name' => 'Homepage',
+            'url' => '/',
+            'site' => $site,
+            'requestMethod' => 'GET|POST|HEAD|DELETE|PUT',
+            'slug' => '/',
+        ]);
     }
 }


moshenskyDV:
Wait pls, forgot to check lint.

src/Route/RoutePageGenerator.php
```diff
@@ -89,7 +89,22 @@ public function update(SiteInterface $site, OutputInterface $output = null, $cle
...
+                        'routeName' => $name,
+                        'name' => $name,
+                        'url' => $route->getPath(),
+                        'site' => $site,
+                        'requestMethod' => isset($requirements['_method']) ? $requirements['_method'] : 'GET|POST|HEAD|DELETE|PUT',
+                        'slug' => '/',
+                    ]);
+                }
+            }
+            $root = ($root) ? $root : $this->pageManager->create([
```
greg0ire: Why the parenthesis? Also `$root ?: $this->pageManager->create([…` should be equivalent, right?

greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

src/Route/RoutePageGenerator.php
```diff
@@ -89,7 +89,22 @@ public function update(SiteInterface $site, OutputInterface $output = null, $cle
 
         // no root url for the given website, create one
         if (!$root) {
-            $root = $this->pageManager->create([
+            // first search for a route
+            foreach ($this->router->getRouteCollection()->all() as $name => $route) {
```
core23: Can you please create a separate method for this.

moshenskyDV:
Travis said, that lint issues in another files (tests). Should i fix this, or not?

src/Route/RoutePageGenerator.php
```diff
@@ -252,6 +227,45 @@ public function update(SiteInterface $site, OutputInterface $output = null, $cle
         }
     }
 
+    /**
+     * Generate root page (parent for another generated pages).
+     * If root path available in config - create from config
+     * Else - generate it statically.
+     *
+     * @param SiteInterface $site
```
greg0ire: Please remove this

src/Route/RoutePageGenerator.php
```diff
@@ -252,6 +227,45 @@ public function update(SiteInterface $site, OutputInterface $output = null, $cle
...
+     *
+     * @return object root page
+     */
+    public function createRootPage(SiteInterface $site)
+    {
+        foreach ($this->router->getRouteCollection()->all() as $name => $route) {
+            if ('/' === $route->getPath()) {
+                $requirements = $route->getRequirements();
+                $name = trim($name);
+                $root = $this->pageManager->create([
```
greg0ire: Why not return here directly? This would make the rest of the code less complex.

greg0ire:
> Please remove this phpdoc

I only meant the line with `@param`

moshenskyDV:
Why param should be removed from PHPDoc?

greg0ire:
What information does it bring? None. It's just noise, and we want people to read our comments, not ignore them because 90% of them are tautologic.

greg0ire:
Use `make cs-fix` to fix the build.

greg0ire:

How to do that?

covex-nn:
Just tested:

* first, i created a Controller with one action, available on "/" route; then i created a site, then executed `bin/console sonata:page:update-core-routes --site=all`
* in pages tree view a controller action became a root for all other pages. so, everything is good here
* then i updated controller action annotation from `@Route(path="/")` to `@Route(path="/test")` and executed `sonata:page:update-core-routes` again
* in pages tree view a path to controller action was updated to "/test", no Homepage route created
* i cleared a cache and executed `sonata:page:update-core-routes` again, a new Homepage page appeared, controller action became a child page for Homepage
* then i updated controller action annotation from `@Route(path="/test")` to `@Route(path="/")`, cleared cache and executed `sonata:page:update-core-routes` again
* from this moment two pages have Url === '/', and this cannot be solved easily

After i added controller action with "/" route into existing project with Homepage page, two pages with '/' route have '/' as Url to. I could not update a page-for-controller-action (there was a error about Url), and when i deleted Homepage, all pages were deleted too =)

moshenskyDV:
@covex-nn i tried to reproduce your commend, and i saw, that we don't check duplicating pages in config.
If we will create someithing like this in rouote config:
```
contact:
    resource: "@ContactBundle/Resources/config/routing.yml"
    prefix:   /contact
contact2:
    resource: "@Contact2Bundle/Resources/config/routing.yml"
    prefix:   /contact
```
System will create 2 pages with same route, and we can not edit any of this page in admin panel (we will see error about already used URL).

So I think - we need to add check in this file for routes duplicates inside this loop (RoutePageGenerator.php):
```
// Iterate over declared routes from the routing mechanism
        foreach ($this->router->getRouteCollection()->all() as $name => $route) {
```

What do you think, guys? It need to be fixed in this PR, or will better to create new PR for this?

rande:
@sonata-project/ekino can you review this PR ?

moshenskyDV:
Changed access modifier. Any update of this?
(I not understood what coveralls wants from me, other tests passed)

src/Route/RoutePageGenerator.php
```diff
@@ -249,4 +239,39 @@ protected function writeln(OutputInterface $output = null, $message)
             $output->writeln($message);
         }
     }
+
+    /**
+     * Generate root page (parent for another generated pages).
+     * If root path available in config - create from config
+     * Else - generate it statically.
+     *
+     * @return object root page
```
core23: Sure, this is just an `object`?

moshenskyDV: Not sure, can you tell me, what is better to write?

core23: PageInterface

src/Route/RoutePageGenerator.php
```diff
@@ -249,4 +239,39 @@ protected function writeln(OutputInterface $output = null, $message)
             $output->writeln($message);
         }
     }
+
+    /**
+     * Generate root page (parent for another generated pages).
```
greg0ire:
```suggestion
     * Generate root page (parent for another generated page).
```

moshenskyDV: Usually, there are couple of pages, so I think it's correct.

greg0ire: It isn't, because "another" is singular. So it's either "another page" or "other pages"

moshenskyDV: Thanks, understood, one minute.

moshenskyDV: @greg0ire could you check again?

OskarStark:
Can you please rebase your PR?

moshenskyDV:
Hi. I did. Unfortunately i have no time to test it manually.

src/Route/RoutePageGenerator.php
```diff
@@ -248,4 +239,39 @@ protected function writeln(OutputInterface $output = null, $message)
...
     }
+
+    /**
+     * Generate root page (parent for another generated page).
+     * If root path available in config - create from config
+     * Else - generate it statically.
+     *
+     * @return PageInterface Root page
+     */
+    private function createRootPage(SiteInterface $site)
```
greg0ire:
```suggestion
    private function createRootPage(SiteInterface $site): PageInterface
```

src/Route/RoutePageGenerator.php
```diff
@@ -248,4 +239,39 @@ protected function writeln(OutputInterface $output = null, $message)
             $output->writeln($message);
         }
     }
+
+    /**
+     * Generate root page (parent for another generated page).
+     * If root path available in config - create from config
+     * Else - generate it statically.
+     *
+     * @return PageInterface Root page
```
greg0ire:
```suggestion
```

src/Route/RoutePageGenerator.php
```diff
@@ -248,4 +239,39 @@ protected function writeln(OutputInterface $output = null, $message)
             $output->writeln($message);
         }
     }
+
+    /**
+     * Generate root page (parent for another generated page).
+     * If root path available in config - create from config
+     * Else - generate it statically.
+     *
```
greg0ire:
```suggestion
```

core23:
Thanks @moshenskyDV 👍

PR #993, opened 2018-06-13
Remove default value for parent association mappings

covex-nn:
I am targeting this branch, because this is a bug fix.

See https://github.com/sonata-project/SonataPageBundle/pull/983 and https://github.com/sonata-project/SonataAdminBundle/pull/5058

## Changelog

```markdown
### Fixed
- Removed default value for parent association mappings
```

## Subject

I am not sure if this is a BC break or not, but without this PR PageBundle is broken.

Now, an exception `Warning: Illegal string offset 'sonata.page.admin.page'` is occured when `Sonata\AdminBundle\Admin\addParentAssociationMapping` tries to set a value, but there is a default string instead of array.

### DIFF
--- a/src/Admin/BlockAdmin.php
+++ b/src/Admin/BlockAdmin.php
@@ -31,11 +31,6 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
 class BlockAdmin extends BaseBlockAdmin
 {
     /**
-     * {@inheritdoc}
-     */
-    protected $parentAssociationMapping = 'page';
-
-    /**
      * @var array
      */
     protected $blocks;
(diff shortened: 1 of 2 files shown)


covex-nn:
@kunicmarko20 there is `protected $classnameLabel` for translations, i will create another PR for that =) But `protected $parentAssociationMapping`, that i removed in this PR, is for another stuff, after https://github.com/sonata-project/SonataAdminBundle/pull/5058 that property must not be a string by default

kunicmarko20:
My bad, didnt read the names. But shouldn’t you update the config then?

kunicmarko20:
Already done by me, I get it now. Sorry 😔

greg0ire:
Thanks @covex-nn !

PR #996, opened 2018-06-14
Auto-register datepicker form theme

covex-nn:

## Changelog

```markdown
### Changed
- Auto-register datepicker form theme
```

## Subject

If datepicker form theme was not added after installation, `Enabled from` and `Enabled to` fields looks like a regular text fields, and it is almost imposible to fill that fields manualy. To fix this, developer must add this configuration:
```yaml
twig:
    form_themes:
        - '@SonataCore/Form/datepicker.html.twig'
```

This PR will add this configuration for developer automaticaly in SonataPageExtension

### DIFF
--- a/src/DependencyInjection/SonataPageExtension.php
+++ b/src/DependencyInjection/SonataPageExtension.php
@@ -17,6 +17,7 @@ use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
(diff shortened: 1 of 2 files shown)


OskarStark:
Could you add a testcase for this?

src/DependencyInjection/SonataPageExtension.php
```diff
@@ -17,15 +17,26 @@
...
  */
-class SonataPageExtension extends Extension
+class SonataPageExtension extends Extension implements PrependExtensionInterface
 {
+    public function prepend(ContainerBuilder $container)
+    {
+        if ($container->hasExtension('twig')) {
+            // add custom form widgets
+            $container->prependExtensionConfig('twig', [
+                'form_themes' => ['@SonataCore/Form/datepicker.html.twig'],
```
kunicmarko20: why isn't this done in the core?

covex-nn: Because it is not necessary in CoreBundle, but mandatory in PageBundle

covex-nn:
@greg0ire there is no docs in PageBundle about datepicker. But i found how to set up datepicker in [CoreBundle docs](https://sonata-project.org/bundles/core/3-x/doc/reference/form_types.html#datepickertype-and-datetimepickertype)

src/DependencyInjection/SonataPageExtension.php
core23: Will this append or replace the current `form_themes` config?

covex-nn: This will prepend ) Application can have a lot of configuration files for a bundle, container will merge these files into one. This array is something like a new file, but its content is placed to the begining of the list.

I found `prepend` method in [symfony-cmf/core-bundle](https://github.com/symfony-cmf/core-bundle/blob/master/src/DependencyInjection/CmfCoreExtension.php) and in [sonata-project/user-bundle](https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/DependencyInjection/SonataUserExtension.php). A bundle can configure another bundle =) But this configuration can be overriden with real config files

kunicmarko20:
Thank you @covex-nn!

PR #999, opened 2018-06-22
Moved asset to root directory

core23:
I am targeting this branch, because this is affects only the build.

## Subject

Moved all assets source and build files to the root inside an new `assets` directory. This will help to get a better overview between symfony bundle files and dev-only files.

Refs https://github.com/sonata-project/dev-kit/pull/213


covex-nn:
Docs https://sonata-project.org/bundles/page/3-x/doc/reference/assets.html should be updated too

core23:
Fixed it @covex-nn

greg0ire:
Thanks @core23 !

PR #1008, opened 2018-09-08
Make services public

core23:

## Changelog

```markdown
### Fixed
- Make `sonata.page.kernel.exception_listener` service public
```

## Subject

This service is used here: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Controller/PageController.php#L80


kunicmarko20:
You probably need a lint PR on brach as we needed it everywhere

PR #1014, opened 2018-10-01
Added support for latest cache bundle

core23:
I am targeting this branch, because this feature is BC.

## Changelog

```markdown
### Added
- Added support for latest `sonata-project/cache-bundle`
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "sonata-project/admin-bundle": "^3.35",
         "sonata-project/block-bundle": "^3.11",
         "sonata-project/cache": "^1.0.2 || ^2.0",
-        "sonata-project/cache-bundle": "^2.4",
+        "sonata-project/cache-bundle": "^2.4 || ^3.0",
         "sonata-project/core-bundle": "^3.9",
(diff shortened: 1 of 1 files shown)


kunicmarko20:
Build is failing

greg0ire:
> Declaration of Sonata\PageBundle\Cache\BlockEsiCache::get(array $keys) must be compatible with Sonata\Cache\CacheAdapterInterface::get(array $keys): Sonata\Cache\CacheElementInterface in /home/travis/build/sonata-project/SonataPageBundle/src/Cache/BlockEsiCache.php

This is what is going to make supporting 2 versions very hard. A solution would be to have 2 classes and a class alias that will use the correct code depending on the version of the block bundle, but that would mean changing the signature of a method, and that is a BC-break.

kunicmarko20:
Stop adding support for new lib versions where we can't do that and release a new major with support for the new lib versions?

core23:
> Stop adding support for new lib versions where we can't do that and release a new major with support for the new lib versions?

Sorry. 

Depenency management is hell :( We can't support it, because it relys on newer php versions

PR #545, opened 2015-10-07
[ADD] allow hybrid pages to be root on the hierarchy

juanolon:
This changes where made for the version 2.3. I think, they should also be compatible with master. but i cannot test it. currently only using 2.3

PR #1016, opened 2018-10-18
Allow hybrid page on root hierarchy

### LINKED ISSUE #537: Hybrid Homepage
Hi,

I am having issues creating a Hybrid homepage. I have created a route/action that looks like this: 

```
    /**
     * @Route("/", name="homepage")
     */
    public function indexAction(Request $request)
    {
        return $this->render('home.html.twig');
    }
```

and I am running:

```
 php app/console sonata:page:update-core-route --site=all
```

However I end up with two Homepage pages. One of them is a CMS page and the other is my Hybrid page. 

Looking at the code I think the issue can be traced back to here: https://github.com/sonata-project/SonataPageBundle/blob/master/Route/RoutePageGenerator.php#L89

The RoutePageGenerator will create a root CMS homepage page that is the very similar to my Hybrid page. This means that I need to go into the DB and update ID's and then delete the CMS homepage page. 

Could we perhaps offer the user an option to enter a default existing route as the root page?

I am happy to have a look at fixing the issue but I want to make sure I am not barking up the wrong tree first.

Thanks,
Stephen

Th3Mouk:
Closes #537 #545 #590

### DIFF
--- a/src/Route/RoutePageGenerator.php
+++ b/src/Route/RoutePageGenerator.php
@@ -89,14 +89,27 @@ class RoutePageGenerator
 
         // no root url for the given website, create one
         if (!$root) {
-            $root = $this->pageManager->create([
+            // first search for a route
+            foreach ($this->router->getRouteCollection()->all() as $name => $route) {
+                if ('/' === $route->getPath()) {
+                    $requirements = $route->getRequirements();
+                    $name = trim($name);
+                    $root = $this->pageManager->create([
+                        'routeName' => $name,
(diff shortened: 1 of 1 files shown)


core23:
Can you please use the issue template @Th3Mouk

moshenskyDV:
I already made a pull request about this ([link](https://github.com/sonata-project/SonataPageBundle/pull/992))

core23:
Duplicate of #992

PR #1017, opened 2018-10-29
Add support for mssql in the SnapshotManager to clean up snapshots

pierredup:
I am targeting this branch, because it's a new feature that is backwards compatible.

## Changelog

```markdown
### Added
- Add support for mssql in `SnapshotManager` to clean up snapshots
```

## Subject

Cleaning up snapshots aren't currently supported on MSSQL databases. This PR adds support for MSSQL when cleaning up snapshots.

### DIFF
--- a/src/Entity/SnapshotManager.php
+++ b/src/Entity/SnapshotManager.php
@@ -288,6 +288,30 @@ class SnapshotManager extends BaseEntityManager implements SnapshotManagerInterf
                 $keep
             ));
         }
+        
+        if ('mssql' === $platform) {
+            return $this->getConnection()->exec(sprintf(
+                'DELETE FROM %s
+                WHERE
+                    page_id = %d
+                    AND id NOT IN (
+                        SELECT id
+                        FROM (
+                            SELECT TOP %d id, publication_date_end
+                            FROM %s
+                            WHERE
+                                page_id = %d
+                            ORDER BY
+                                publication_date_end DESC
+                        ) AS table_alias
+                )',
+                $tableName,
+                $page->getId(),
+                $keep,
+                $tableName,
+                $page->getId()
+            ));
+        }
 
         throw new \RuntimeException(sprintf('The %s database platform has not been tested yet. Please report us if it works and feel free to create a pull request to handle it ;-)', $platform));
     }
--- a/src/Entity/SnapshotManager.php
+++ b/src/Entity/SnapshotManager.php
@@ -236,84 +236,28 @@ class SnapshotManager extends BaseEntityManager implements SnapshotManagerInterf
             throw new \RuntimeException(sprintf('Please provide an integer value, %s given', \gettype($keep)));
         }
 
-        $tableName = $this->getTableName();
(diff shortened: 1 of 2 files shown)


OskarStark:
Thank you for your feedback, IMO a unit test would be enough. We can expect the correct method is called with the expected arguments, do you think you are able todo this?

I would be very thankful if you could cover the other platforms, too. The most stuff should be copy paste

src/Entity/SnapshotManager.php
```diff
@@ -289,6 +289,30 @@ public function cleanup(PageInterface $page, $keep)
             ));
         }
 
+        if ('mssql' === $platform) {
```
core23: This method is already very long... Can you try to extra this to an own method / class

pierredup:
Is there a specific reason why DQL wasn't used for this method? It would help with supporting any platform without having to create queries for each one.

I pushed changes to use DQL instead which also makes the method quite smaller. Please have a look and let me know if you are happy with this approach or if there is anything I'm missing.

I'll add a test for this shortly.

OskarStark:
> Is there a specific reason why DQL wasn't used for this method? It would help with supporting any platform without having to create queries for each one.

I would say yes, maybe the git history can help?

pierredup:
Git history doesn't help giving any information on why it was done like this, which is why I asked if there is a specific reason that was maybe part of another discussion somewhere that might not be available in Github

https://github.com/sonata-project/SonataPageBundle/commit/5ea71c900bdafc45702a51d8af43e46f8cbaa24d#diff-e79e02424892dacf03a0257d5752dd66R255
https://github.com/sonata-project/SonataPageBundle/commit/8629f4224d88e73ce3becb0b1c010607626563b0

OskarStark:
hmm @mremi can you give us some insights or at least give your 2 cents here? Thank you!

src/Entity/SnapshotManager.php
```diff
@@ -236,60 +236,28 @@ public function cleanup(PageInterface $page, $keep)
...
+            ->getArrayResult();
+
+        $query = $em->createQueryBuilder()
+            ->delete()
+            ->from($this->class, 's')
+            ->where('s.page = :page')
+            ->andWhere('s.id NOT IN (:pageId)')
+            ->setParameter('page', $page)
+            ->setParameter('pageId', array_column($snapShots, 'id'))
+            ->getQuery();
```
greg0ire: Why use a query builder here over plain DQL? There are no conditional clauses…

pierredup: Probably just out of habit to use a query builder, it's a personal preference to use a query builder over DQL.
But if plain DQL is the preferred way then I'll happily change it

greg0ire: Please do. Plain DQL is not the preferred way, because you should pick different APIs for different use cases. Here is a table to make your decision: http://talks.nekland.fr/Doctrine/#/32 (:fr:)

core23:
Can you please finish this @pierredup ?

core23:
This PR looks dead. Feel free to reopen if you will continue working on this.

PR #1019, opened 2018-11-01
Catch empty locale in exception listener

core23:

## Changelog

```markdown
### Fixed
- Catch empty locale in exception listener
```

## Subject

The Request object in symfony expects only string, not null values.


kunicmarko20:
Thanks @core23

PR #1028, opened 2019-01-02
[Docs] small changes

OskarStark:


docs/reference/advanced_configuration.rst
```diff
@@ -9,7 +9,7 @@ Full configuration options:
 
     .. code-block:: yaml
 
-        # app/config/config.yml
+        # config/packages/symfony_cmf_routing_extra.yaml
```
dbu: this should be cmf_routing.yaml.

i notice that the part below is really outdated. the very first experiments had this config key, but since then we changed to `cmf_routing`. routers_by_id is then correct, however.

dbu: the service id of the dynamic router in the comment below is cmf_routing.dynamic_router

docs/reference/multisite.rst
```diff
@@ -22,37 +22,33 @@ With this strategy it is possible to handle sites like :
 * http://sonata-project.com
 * http://sonata-project.net
 
-Just configure the ``sonata_page`` section as:
+Configure the ``sonata_page`` section as:
```
dbu: thanks!

docs/reference/advanced_configuration.rst
OskarStark: thank you, I updated it, can you please verify again?

greg0ire:
Thanks @OskarStark !

PR #1046, opened 2019-03-10
Use page title in breadcrumb

core23:
## Subject
The breadcrumb was using the (technical) page name if no special SEO title was defined. The page title is now used if it is defined, otherwise the translated page name is used.

## Changelog

```markdown
### Changed
- Translate page name in breadcrumb if no title is defined

### Fixed
- Use page title in breadcrumbs
```

### DIFF
--- a/src/Block/BreadcrumbBlockService.php
+++ b/src/Block/BreadcrumbBlockService.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sonata\PageBundle\Block;
 
 use Knp\Menu\FactoryInterface;
+use Knp\Menu\ItemInterface;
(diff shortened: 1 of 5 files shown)


src/Block/BreadcrumbBlockService.php
```diff
@@ -109,4 +100,28 @@ protected function getCurrentPage()
 
         return $cms->getCurrentPage();
     }
+
+    /**
+     * @param ItemInterface $menu
+     * @param PageInterface $page
+     */
```
greg0ire: Please remove this phpdoc

core23: Fixed

greg0ire:
Thanks @core23 !

PR #1050, opened 2019-04-15
Fix "Custom URL"

### LINKED ISSUE #502: Custom URL (not fully implemented ?)
Hello everyone,

I've tried to analyse the workflow of the router, because the use of field 'custom URL' is obscure. 
In my project, except the corresponding field in the database, he's never considered.

I understand how the `CmsPageRouter.php` generate and match route of different page type, but the implementation the customURL seem forgotten for a part.

When a snapshot is created the `Transformer.php` do :

``` php
$snapshot->setUrl($page->getUrl());
$snapshot->setRouteName($page->getRouteName());
$snapshot->setPageAlias($page->getPageAlias());
$content['slug'] = $page->getSlug();
```

But when a snap is load :

``` php
$page->setCustomUrl($snapshot->getUrl());
$page->setUrl($snapshot->getUrl());
$page->setRouteName($snapshot->getRouteName());
$page->setPageAlias($snapshot->getPageAlias());
$page->setSlug($content['slug']);
```

So i conclude, that the snap doesn't care about it.

And in the `CmsPageRouter.php` I've found this :

``` php
/**
 * Returns the Url from a Page object
 *
 * @param PageInterface $page
 *
 * @return string
 */
protected function getUrlFromPage(PageInterface $page)
{
    return $page->getCustomUrl() ?: $page->getUrl();
}

/**
     * Generates an URL from a Page object
     *
     * @param PageInterface $page          Page object
     * @param array         $parameters    An array of parameters
     * @param bool|string   $referenceType The type of reference to be generated (one of the constants)
     *
     * @return string
     *
     * @throws \RuntimeException
     */
    protected function generateFromPage(PageInterface $page, array $parameters = array(), $referenceType = self::ABSOLUTE_PATH)
    {
        // hybrid pages use, by definition, the default routing mechanism
        if ($page->isHybrid()) {
            return $this->router->generate($page->getRouteName(), $parameters, $referenceType);
        }

        $url = $this->getUrlFromPage($page);

        if ($url === false) {
            throw new \RuntimeException(sprintf('Page "%d" has no url or customUrl.', $page->getId()));
        }

        return $this->decorateUrl($url, $parameters, $referenceType);
    }
```

IMO I miss part of the reasoning or a part of the workflow is missing.

Thanks for your help

RSalo:
## Subject

Probably, it's a not good solution but it works though.

Closes #502 

## Changelog

```markdown

### Fixed
Fix issue when doesn't work field "Custom url"
```

### DIFF
--- a/src/CmsManager/CmsManagerInterface.php
+++ b/src/CmsManager/CmsManagerInterface.php
@@ -47,7 +47,7 @@ interface CmsManagerInterface
     public function getPageByUrl(SiteInterface $site, $slug);
 
     /**
-     * Returns a fully loaded page ( + blocks ) from a url.
+     * Returns a fully loaded page ( + blocks ) from a custom url.
      *
      * @param SiteInterface $site
      * @param string        $slug
--- a/src/CmsManager/CmsManagerInterface.php
+++ b/src/CmsManager/CmsManagerInterface.php
@@ -47,6 +47,16 @@ interface CmsManagerInterface
     public function getPageByUrl(SiteInterface $site, $slug);
 
     /**
+     * Returns a fully loaded page ( + blocks ) from a url.
+     *
+     * @param SiteInterface $site
+     * @param string        $slug
+     *
+     * @return PageInterface
+     */
+    public function getPageByCustomUrl(SiteInterface $site, $slug);
+
+    /**
      * Returns a fully loaded page ( + blocks ) from a route name.
      *
      * @param SiteInterface $site
--- a/src/Resources/views/base_layout.html.twig
+++ b/src/Resources/views/base_layout.html.twig
@@ -174,6 +174,8 @@ file that was distributed with this source code.
             {% endblock %}
         {% endblock %}
 
+        {% block sonata_block_page_monitoring %}
         <!-- monitoring:3e9fda56df2cdd3b039f189693ab7844fbb2d4f6 -->
+        {% endblock %}
     </body>
 </html>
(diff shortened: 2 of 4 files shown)


src/CmsManager/CmsManagerInterface.php
```diff
@@ -46,6 +46,16 @@ public function findContainer($name, PageInterface $page, BlockInterface $parent
...
 
+    /**
+     * Returns a fully loaded page ( + blocks ) from a custom url.
+     *
+     * @param SiteInterface $site
+     * @param string        $slug
+     *
+     * @return PageInterface
+     */
+    public function getPageByCustomUrl(SiteInterface $site, $slug);
```
core23: This is a BC break :(

You can't add new methods to an interface with a minor release

RSalo: > This is a BC break :(
> 
> You can't add new methods to an interface with a minor release

Oh, should i add this to master branch?

core23: You can comment the method, with a hint to uncomment it with the next major release

src/Resources/views/base_layout.html.twig
```diff
@@ -174,6 +174,8 @@ file that was distributed with this source code.
             {% endblock %}
         {% endblock %}
 
+        {% block sonata_block_page_monitoring %}
         <!-- monitoring:3e9fda56df2cdd3b039f189693ab7844fbb2d4f6 -->
```
core23: I think you can remove this comment instead of adding a new block for this

PR #1057, opened 2019-04-24
fix compare with a different value type in Tree action

phucwan91:
## Subject

I am targeting this branch, because of an issue in the `PageAdmincontroller`.

## Changelog

```markdown

### Fixed
compare with a different value type in the `treeAction`
```


OskarStark:
Thank you

PR #1067, opened 2019-07-16
Manual merge

greg0ire:


src/Admin/PageAdmin.php
```diff
@@ -204,10 +201,7 @@ public function getBatchActions()
         return $actions;
     }
 
-    /**
-     * @param SiteManagerInterface $siteManager
-     */
-    public function setSiteManager(SiteManagerInterface $siteManager): void
```
phansys: Why the return type declaration was removed?

greg0ire: Good catch, my mistake!

PR #1071, opened 2019-08-17
Use more specific PHPDoc

core23:
## Subject

I am targeting this branch, because this is pedantic.

## Changelog

```markdown
### Changed
- Match PHPDoc with doctrine model 

### Removed
- Remove superfluous PHPDoc 
```

### DIFF
--- a/src/Entity/BlockManager.php
+++ b/src/Entity/BlockManager.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Entity;
 
+use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\BlockBundle\Model\BlockManagerInterface;
 use Sonata\DatagridBundle\Pager\Doctrine\Pager;
 use Sonata\DatagridBundle\ProxyQuery\Doctrine\ProxyQuery;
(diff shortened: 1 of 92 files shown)


src/Entity/BlockManager.php
```diff
@@ -43,8 +40,6 @@ public function save($block, $andFlush = true)
      * @param int  $parentId Parent block Id (needed when partial = true)
      * @param int  $pageId   Page Id (needed when partial = true)
      * @param bool $partial  Should we use partial references? (Better for performance, but can lead to query issues.)
-     *
-     * @return mixed
```
phansys: If we will remove the `mixed` types, we should update the rule at [dev-kit](https://github.com/sonata-project/dev-kit/blob/165dd7bb80c3e3a6ce69f0f8f440dd737ea25a47/project/.php_cs.dist.twig#L31).

src/Model/SnapshotInterface.php
```diff
@@ -17,6 +17,13 @@
  * Defines methods to interact with the persistency layer of a SnapshotInterface.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @method int|null           getTargetId()
```
phansys: Are these methods intended to be part of the interface in a next major release?

core23: We should add all @methods in the next major version

src/Entity/BlockManager.php
core23: IMHO we could leave the settings as they are. It is no problem to keep some mixin settings. I reverted some of my changes.

tests/Block/BlockContextManagerTest.php
```diff
@@ -28,7 +28,7 @@ class BlockContextManagerTest extends TestCase
 {
     public function testGetWithValidData()
     {
-        $service = $this->getMockForAbstractClass(AbstractBlockService::class);
+        $service = $this->getMockBuilder(AbstractBlockService::class)->disableOriginalConstructor()->getMock();
```
OskarStark: Why not createMock?

core23: Fixed

PR #1077, opened 2019-09-26
fix(decorator-strategy): check headers values as strings

fancyweb:
## Subject

I am targeting this branch, because it's a bug fix.

This project does not properly compare HttpFoundation headers values. The returned value can be a string, an array of strings or null. The following commit : https://github.com/symfony/symfony/commit/ef5ead00059fc321e7fda20c2dfdd4440585be26#diff-5f886132f28113d74a1a219e967e88fcR124 recently enforced the returned values (by casting them to strings). Thus, the strict comparisons against booleans does not work anymore.

The bug was not detected by the CI because the wrong bag is used for headers in the unit test. I updated it and I also added assertions for the non tested header `x-sonata-page-not-decorable`.

I think it's better to handle both cases (bool & string) to not force a `symfony/http-foundation` bump.

## Changelog

```markdown
### Fixed
- Fixed `DecoratorStrategy` compatibility with `symfony/http-foundation` >= 3.4.31
```

### DIFF
--- a/src/CmsManager/DecoratorStrategy.php
+++ b/src/CmsManager/DecoratorStrategy.php
@@ -57,12 +57,14 @@ class DecoratorStrategy implements DecoratorStrategyInterface
             return false;
         }
 
-        if (true === $response->headers->get('x-sonata-page-not-decorable', false)) {
+        // TODO: drop support for `true` value when bumping symfony/http-foundation dependency to >= 3.4.31
+        if (\in_array($response->headers->get('x-sonata-page-not-decorable', '0'), [true, '1'], true)) {
             return false;
         }
 
+        // TODO: drop support for `true` value when bumping symfony/http-foundation dependency to >= 3.4.31
         // the main controller explicitly force the the page to be decorate
-        if (true === $response->headers->get('x-sonata-page-decorable', false)) {
+        if (\in_array($response->headers->get('x-sonata-page-decorable', '0'), [true, '1'], true)) {
             return true;
         }
 
--- a/tests/CmsManager/DecoratorStrategyTest.php
+++ b/tests/CmsManager/DecoratorStrategyTest.php
@@ -15,13 +15,14 @@ namespace Sonata\PageBundle\Tests\Page;
 
 use PHPUnit\Framework\TestCase;
 use Sonata\PageBundle\CmsManager\DecoratorStrategy;
-use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
 class DecoratorStrategyTest extends TestCase
 {
+    // TODO: drop support for headers' boolean values when bumping symfony/http-foundation dependency to >= 3.4.31
     public function testIsDecorable()
     {
         $response = new Response('dummy');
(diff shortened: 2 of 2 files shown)


src/CmsManager/DecoratorStrategy.php
```diff
@@ -57,12 +57,14 @@ public function isDecorable(Request $request, $requestType, Response $response)
             return false;
         }
 
-        if (true === $response->headers->get('x-sonata-page-not-decorable', false)) {
+        // TODO: remove true support when the symfony/http-foundation dependency is > 3.4.31
```
phansys:
```suggestion
        // TODO: drop support for `true` value when bumping symfony/http-foundation dependency to > 3.4.31
```

src/CmsManager/DecoratorStrategy.php
```diff
@@ -57,12 +57,14 @@ public function isDecorable(Request $request, $requestType, Response $response)
...
+        if (\in_array($response->headers->get('x-sonata-page-not-decorable', '0'), [true, '1'], true)) {
             return false;
         }
 
+        // TODO: remove true support when the symfony/http-foundation dependency is > 3.4.31
```
phansys:
```suggestion
        // TODO: drop support for `true` value when bumping symfony/http-foundation dependency to > 3.4.31
```

tests/CmsManager/DecoratorStrategyTest.php
```diff
@@ -15,13 +15,14 @@
...
 use Sonata\PageBundle\CmsManager\DecoratorStrategy;
-use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
 class DecoratorStrategyTest extends TestCase
 {
+    // TODO: remove headers boolean support when the symfony/http-foundation dependency is > 3.4.31
```
phansys:
```suggestion
    // TODO: drop support for `true` value when bumping symfony/http-foundation dependency to > 3.4.31

```

fancyweb:
@phansys Thanks for your suggestions. I could not apply them directly because I had to correct the `>` to `>=`.

src/CmsManager/DecoratorStrategy.php
```diff
@@ -57,12 +57,14 @@ public function isDecorable(Request $request, $requestType, Response $response)
             return false;
         }
 
-        if (true === $response->headers->get('x-sonata-page-not-decorable', false)) {
+        // TODO: drop support for `true` value when bumping symfony/http-foundation dependency to >= 3.4.31
+        if (\in_array($response->headers->get('x-sonata-page-not-decorable', '0'), [true, '1'], true)) {
```
core23:
```suggestion
        if (true === (bool) $response->headers->get('x-sonata-page-not-decorable', false)) {
```
Is more understandable

fancyweb: More understandable but casting is less strict as it allows us to support "out of scope" HttpFoundation implementation (for example if userland extends the headers bag and break the contract) + the default value passed to `get` must be a string.

greg0ire: Why not use `getBoolean` here?

fancyweb: `getBoolean` doesn't exist in `ResponseHeaderBag`

greg0ire: Oh right, I got confused with `ParameterBag`, nevermind

fancyweb: It was a great suggestion actually, I was disappointed too 😅

greg0ire:
Note that 2.8 will no longer be supported, even for security fixes, so a bump to 3.4.31 would really not be an issue to me.

core23:
Thanks @fancyweb

PR #1084, opened 2019-11-02
3.13.1

greg0ire:


core23:
Do we need this?

greg0ire:
What do you mean?

core23:
This is a dev dependency. So we do we need a new release?

greg0ire:
Oh right. We don't.

OskarStark:
But if you are using npm install and not, how you should, npm install —production, you end up with this dependency in prod.
So why not just create a release? Doesn’t hurt

core23:
Why would you do this for this *library*? If you call `npm install` on your *project*, you should only load all non-dev dependencies on your project.

OskarStark:
You are right, I mixed things up 😄

PR #1089, opened 2019-11-11
Dropping support for symfony < 3.4

franmomu:
## Subject

This is the same thing as https://github.com/sonata-project/SonataAdminBundle/pull/5733, dropping support of Symfony < 3.4 and >= 4, < 4.2

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Removed
- Support for Symfony < 3.4
- Support for Symfony >= 4, < 4.2
```


franmomu:
There are some:
```
@deprecated This property is deprecated since version 2.4 and will be removed in 3.0
```

Should this be removed?

greg0ire:
> @deprecated This property is deprecated since version 2.4 and will be removed in 3.0

Woops looks like we forgot those when releasing 3.0.0 . I think the right call would be to adapt that comment cause removing the property would be a BC-break.

core23:
Thanks @franmomu

PR #1103, opened 2020-01-12
3.14.0

greg0ire:


DidierPais:
Hello,

I'm currently using version 3.13.0 and I can't update to 3.14.0 (I need the fix on `BlockInteractor`) because I have `symfony/dependency-injection` 4.4.2 but the composer.json says `"symfony/dependency-injection": "^3.4 || ^4.2,<4.4.0",`

Is it an oversight ?

Thank you.

greg0ire:
IIRC tests fail with DI 4.4, consider sending a PR without the constraint and with the tests fixed

PR #1126, opened 2020-03-11
Remove cocur/slugify dependency

fancyweb:
## Subject

I wanted to allow cocur/slugify ^4.0 but I actually didn't find any usage in the
code so I guess we can remove the dependency entirely.

I am targeting this branch, because this change is backward compatible.

## Changelog
```markdown
### Removed
- Removed "cocur/slugify" dependency
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
     ],
     "require": {
         "php": "^7.1",
-        "cocur/slugify": "^1.0 || ^2.0 || ^3.0",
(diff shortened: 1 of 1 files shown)


VincentLanglet:
It was added by @soullivaneuh Here https://github.com/sonata-project/SonataPageBundle/commit/223881d5832e61cde859d0b97c2cacfb0717f161
Any idea why ?

OskarStark:
Maybe because it was an explicit dependency, but not anymore? Can you checkout the this version and search for the occurrence?

PR #1127, opened 2020-03-14
Make page names more readable when using generator

core23:
## Subject

When calling `sonata:page:update-core-route`, generated page names will not use the route name as default anymore. A meaningful page title will guessed by replacing the underscore with words.

E.g. the generated page for the route `sonata_news_view` will get `Sonata News View` as default page name. 

The name is only displayed on the admin backend and it will only affect newly generated pages, so this is BC.

## Changelog

```markdown
### Changed
- Removed underscores in page names when calling `sonata:page:update-core-route`
```

### DIFF
--- a/src/Route/RoutePageGenerator.php
+++ b/src/Route/RoutePageGenerator.php
@@ -98,6 +98,7 @@ class RoutePageGenerator
         // Iterate over declared routes from the routing mechanism
         foreach ($this->router->getRouteCollection()->all() as $name => $route) {
             $name = trim($name);
+            $displayName = $this->displayName($name);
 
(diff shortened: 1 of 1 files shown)


src/Route/RoutePageGenerator.php
```diff
@@ -274,4 +275,14 @@ private function createRootPage(SiteInterface $site): PageInterface
             'slug' => '/',
         ]);
     }
+
+    /**
+     * @param string $name
+     *
+     * @return string
+     */
```
VincentLanglet: Phpdoc should be removed when you use typehint

greg0ire:
Thanks @core23 !

PR #1136, opened 2020-04-13
Applied fixes from FlintCI

soullivaneuh:


franmomu:
Fixed in https://github.com/sonata-project/SonataPageBundle/pull/1137

PR #1137, opened 2020-04-13
Drop support for Symfony < 4.3

franmomu:
## Subject
Updating the project dropping Symfony < 4.3

I am targeting this branch, because this is BC.

Closes #1135

## Changelog

```markdown
### Removed
- Support for Symfony < 4.3
### Fixed
- Missing var type declaration
```

PR #1139, opened 2020-04-13
Fix redirect after a batch snapshot

### LINKED ISSUE #1134: Error during "Create a Publication" from list view in page:  An exception has been thrown during the rendering of a template ("Notice: Array to string conversion")
Hello!
Please help me to solve the issue.

Steps to reproduce the error:

Install the sonata page bundle and do the required configuration.

Go to Pages at back-end:

Pages > Manage Pages >> List View 

and checked any of the pages which you have added already or add new one and do the same.

and select the option "Create a publication" >> "OK" >> "Yes Execute" and then there will be error "An exception has been thrown during the rendering of a template ("Notice: Array to string conversion")."

screenshot2020 03 30-14_15_55

Please note: I have tried with a fresh installation of the Sonata Page Bundle it seems there is the same error.

Please help me to solve the error or suggest me any version which is resolved with the mentioned error.

Many Thanks!!

franmomu:
## Subject

This is made on top of #1137

Closes #1134

## Changelog

```markdown
### Fixed
- Redirecting after batch snapshot
```

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -4,7 +4,7 @@ Installation
 Prerequisites
 -------------
 
-PHP 7.1 and Symfony >=4.3 are needed to make this bundle work, there are
+PHP 7.2 and Symfony >=4.3 are needed to make this bundle work, there are
 also some Sonata dependencies that need to be installed and configured beforehand:
 
     - SonataAdminBundle_
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -28,9 +28,6 @@ Add the dependant bundles to the vendor/bundles directory:
 .. code-block:: bash
 
     composer require sonata-project/page-bundle --no-update
-
-    # for SonataPageBundle > 2.3.6
-    composer require sonata-project/datagrid-bundle --no-update
     composer require sonata-project/doctrine-orm-admin-bundle --no-update
 
     # optional when using API
@@ -69,10 +66,6 @@ Add these bundles in the config mapping definition (or enable `auto_mapping`_):
                     mappings:
                         ApplicationSonataPageBundle: ~ # only once the ApplicationSonataPageBundle is generated
                         SonataPageBundle: ~
-
-        dbal:
-            types:
-                json: Sonata\Doctrine\Types\JsonType
 
 CMF Routing Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/Model/PageTest.php
+++ b/tests/Model/PageTest.php
@@ -28,7 +28,11 @@ class PageTest extends TestCase
     public function testSlugify(): void
     {
         setlocale(LC_ALL, 'en_US.utf8');
-        setlocale(LC_CTYPE, 'en_US.utf8');
+
+        $reflectionClass = new \ReflectionClass(Page::class);
+        $property = $reflectionClass->getProperty('slugifyMethod');
+        $property->setAccessible(true);
+        $property->setValue(null);
 
         $this->assertSame(Page::slugify('test'), 'test');
         $this->assertSame(Page::slugify('S§!@@#$#$alut'), 's-alut');
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.2",
         "doctrine/doctrine-bundle": "^1.12.3 || ^2.0",
         "doctrine/persistence": "^1.3.3",
         "sonata-project/admin-bundle": "^3.35",
--- a/composer.json
+++ b/composer.json
@@ -61,6 +61,8 @@
         "jms/serializer-bundle": "^1.0 || ^2.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
         "nelmio/api-doc-bundle": "^2.4",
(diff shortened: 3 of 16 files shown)


docs/reference/installation.rst
```diff
@@ -28,9 +28,6 @@ Add the dependant bundles to the vendor/bundles directory:
 .. code-block:: bash
 
     composer require sonata-project/page-bundle --no-update
-
-    # for SonataPageBundle > 2.3.6
-    composer require sonata-project/datagrid-bundle --no-update
```
franmomu, line 33: [datagrid-bundle is already required in composer.json](https://github.com/sonata-project/SonataPageBundle/blob/3.x/composer.json#L32)

docs/reference/installation.rst
```diff
@@ -70,10 +67,6 @@ Add these bundles in the config mapping definition (or enable `auto_mapping`_):
                         ApplicationSonataPageBundle: ~ # only once the ApplicationSonataPageBundle is generated
                         SonataPageBundle: ~
 
-        dbal:
-            types:
-                json: Sonata\Doctrine\Types\JsonType
```
franmomu, line 75: [JsonType is deprecated ](https://github.com/sonata-project/sonata-doctrine-extensions/blob/1.x/src/Types/JsonType.php)in favor of the Doctrine one.

tests/Model/PageTest.php
```diff
@@ -28,7 +28,11 @@ class PageTest extends TestCase
     public function testSlugify(): void
     {
         setlocale(LC_ALL, 'en_US.utf8');
-        setlocale(LC_CTYPE, 'en_US.utf8');
+
+        $reflectionClass = new \ReflectionClass(Page::class);
+        $property = $reflectionClass->getProperty('slugifyMethod');
+        $property->setAccessible(true);
+        $property->setValue(null);
```
franmomu, line 35: This test was failing because with the functional test, when this bundle is loaded, [`Page::slugifyMethod` is set](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/SonataPageBundle.php#L64). So it was using the cocur slugify method provided by the CoreBundle.

phansys: If we need to check the implementation provided by this bundle, should not be better to use the corresponding configuration?
https://github.com/sonata-project/SonataPageBundle/blob/926cbbf124e214cc9b9824051b7273836635fd35/tests/App/config/sonata.yaml#L14

franmomu: [The current implementation is deprecated](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Model/Page.php#L590), there is no `slugify_service` in this bundle. [If you don't set it, it will use `sonata.core.slugify.native`](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/DependencyInjection/Configuration.php#L118). Not sure if this is what you meant 🤔

phansys: I'm not a user of this bundle, so I haven't experience with its behavior.
Re-readig, I thought this test is functional, but it isn't, right? If not, why is it being affected by a functional test?
>This test was failing because with the functional test, when this bundle is loaded, Page::slugifyMethod is set.

Based in your comment:
>So it was using the cocur slugify method provided by the CoreBundle instead of the own implementation.

I understood what you're confirming here:
>If you don't set it, it will use sonata.core.slugify.native.

About my comment
>If we need to check the implementation provided by this bundle, should not be better to use the corresponding configuration?

I understand there is no service defining a slugger in this bundle. I was referencing the implementation you've described before. What I'm asking is if there is a way to avoid the "hack" to manipulate a property value with via reflection.

Finally, since this test isn't checking all the available implementations, I think it should be renamed in order to make clear which one is intended to checked (own, "sonata.core.slugify.native", or "sonata.core.slugify.cocur").

franmomu: > Re-readig, I thought this test is functional, but it isn't, right? If not, why is it being affected by a functional test?

What I did was the funcional test, I didn't write this test. 

I had to add this "hack" in this because when using the bundle in a normal way (by normal I mean loading `SonataPageBundle` class), it calls `Page::setSlugify` method as I said.

> Finally, since this test isn't checking all the available implementations, I think it should be renamed in order to make clear which one is intended to checked (own, "sonata.core.slugify.native", or "sonata.core.slugify.cocur").

This test I modified is in the class `tests/Model/PageTest.php`, to me it is clear that is testing the own deprecated and very unlikely to be executed implementation, IMHO that "hack" in a test that is going to be removed in the next major version is not really a big deal.

phansys: >This test I modified is in the class tests/Model/PageTest.php, to me it is clear that is testing the own deprecated and very unlikely to be executed implementation

Since `Page::slugify()` can use any implementation based on the callable passed to `Page::setSlugifyMethod()`, I'm sorry, but it's not clear for me that the own implementation is being tested, since it depends on the value of the property you're manipulating here:

https://github.com/sonata-project/SonataPageBundle/blob/926cbbf124e214cc9b9824051b7273836635fd35/src/Model/Page.php#L200-L203

https://github.com/sonata-project/SonataPageBundle/blob/926cbbf124e214cc9b9824051b7273836635fd35/src/Model/Page.php#L590-L615

Whatever, if you think this is the way, go ahead.

greg0ire:
Thanks @franmomu !

PR #1148, opened 2020-05-07
Ignore subrequests in SiteSelector

core23:
## Subject

When rendering a controller, the existing seo matadata of the current page is overridden by the rendered sub controller. This is a problem if the rendered controller does not have any meta information.

```
    {{ render(controller('SonataNewsBundle:Post:comments', {'postId': post.id})) }}
```

## Changelog

```markdown
### Fixed
- Ignore subrequests in `SiteSelector`
```


greg0ire:
Thanks @core23 !

PR #1150, opened 2020-05-08
Truncate texts in page composer

core23:
## Subject

When editing a block in the page composer, the html is parsed and put into an input field. 

Add missing trim to remove whitespaces from the parsed html.

image

## Changelog

```markdown
### Fixed
- Truncate texts in page composer
```


VincentLanglet:
Thanks @core23

PR #1166, opened 2020-06-22
Replace truncate filter with u filter

wbloszyk:
## Subject

This bundle was using `truncate` filter without explicitly require `twig/extensions`, with this PR it uses `u` filter.

I added `twig/extra-bundle` to autoload the Twig extension and a fallback in case Flex is not used.

Ref: sonata-project/SonataAdminBundle#6132

I am targeting this branch, because these changes are BC.

## Changelog

```markdown
### Added
- Added `twig/string-extra` dependency.
### Changed
- Changed use of `truncate` filter with `u` filter.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,8 @@
         "symfony/security-http": "^4.3",
         "symfony/templating": "^4.3",
         "symfony/validator": "^4.3",
-        "twig/twig": "^2.10"
+        "twig/string-extra": "^3.0",
+        "twig/twig": "^2.12.1"
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.1 || >=3.0",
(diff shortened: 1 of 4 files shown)


wbloszyk:
@core23 @jordisala1991 @phansys RTM

composer.json
```diff
@@ -65,6 +66,10 @@
         "symfony/css-selector": "^4.3",
         "symfony/phpunit-bridge": "^5.0"
     },
+    "suggest": {
+        "sonata-project/notification-bundle": "^2.2",
```
jordisala1991, line 70: The suggest section is not for the version

wbloszyk: This should not be in suggest. My mistake while copy it from other truncate PR.

jordisala1991:
Thank you @wbloszyk

PR #1170, opened 2020-06-29
Fix StringExtension registration

wbloszyk:
## Subject

This PR contains:
- move StringExtension registration to Compiler pass and check all `twig.extension` service to avoid duplication
- improve tests
- fix previus broken checking
```php
... !is_a($container->getDefinition('twig.extension.string')->getClass(), StringExtension::class)) ...
// it is like this:   !is_a('Twig\Extra\String\StringExtension', StringExtension::class)) ...
```

PATCH for #1166.

## Changelog

```markdown
### Fixed
- Fix `Twig\Extra\String\StringExtension` optional auto-registration to avoid duplication `twig.extension` service 
```

### DIFF
--- a/src/DependencyInjection/Compiler/TwigStringExtensionCompilerPass.php
+++ b/src/DependencyInjection/Compiler/TwigStringExtensionCompilerPass.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Twig\Extra\String\StringExtension;
+
+final class TwigStringExtensionCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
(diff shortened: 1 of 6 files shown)


wbloszyk:
Anyone have idea why CompillerPass is not used in functional tests?

jordisala1991:
Did you debug it and the function is not being called?

wbloszyk:
It is called. Probably some tests do not load it when it is needed. I found which test should be fix here. Need do it the same in AdminBundle. I will do it Tomorrow and finnaly CoreBundle will be drop in sandbox = whole Sonata 3. :tada:

src/DependencyInjection/Compiler/TwigStringExtensionCompilerPass.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Twig\Extra\String\StringExtension;
+
+final class TwigStringExtensionCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        foreach ($container->findTaggedServiceIds('twig.extension') as $id => $attributes) {
+            if (StringExtension::class === $container->getDefinition($id)->getClass()) {
```
phansys, line 26: I just see one case escaping from this condition. I've hit that scenario some time ago in other package and had to check that case too, but I'm not able to remember where.
The fact is, `$container->getDefinition($id)->getClass()` can return a parameter name (like "%class%"). So, this condition should take care about this situation making the conversion (`ContainerBuilder::getParameter()`) before the current check.

wbloszyk: Technicly it is possible but it will be very strange to use final class in this case.

jordisala1991:
Thank you @wbloszyk

PR #1173, opened 2020-07-06
Fix RestFul API

wbloszyk:
## Subject

Part of https://github.com/sonata-project/dev-kit/issues/778

## Changelog

```markdown
### Added
- Added public alias `Sonata\PageBundle\Controller\Api\BlockController` for `sonata.page.controller.api.block` service
- Added public alias `Sonata\PageBundle\Controller\Api\PageController` for `sonata.page.controller.api.page` service
- Added public alias `Sonata\PageBundle\Controller\Api\SiteController` for `sonata.page.controller.api.site` service
- Added public alias `Sonata\PageBundle\Controller\Api\SnapshotController` for `sonata.page.controller.api.snapshot` service

### Fixed
- fix RestFul API - `Class could not be determined for Controller identified` Error
```

### DIFF
--- a/src/Resources/config/api_controllers.xml
+++ b/src/Resources/config/api_controllers.xml
@@ -5,6 +5,7 @@
             <argument type="service" id="sonata.page.manager.block"/>
             <argument type="service" id="form.factory"/>
         </service>
+        <service id="Sonata\PageBundle\Controller\Api\BlockController" alias="sonata.page.controller.api.block" public="true"/>
         <service id="sonata.page.controller.api.page" class="Sonata\PageBundle\Controller\Api\PageController">
             <argument type="service" id="sonata.page.manager.site"/>
             <argument type="service" id="sonata.page.manager.page"/>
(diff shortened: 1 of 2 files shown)


greg0ire:
Why do we suddenly have this "Class could not be determined for Controller identified" error?

wbloszyk:
Suddenly? This was broken for long time. Propably service cannot be usage from Symfony4

greg0ire:
> This was broken for long time.

Maybe it has been broken for 2 years, but one day in worked, and the day after, it didn't, right? Why is that?

wbloszyk:
> > This was broken for long time.
> 
> Maybe it has been broken for 2 years, but one day in worked, and the day after, it didn't, right? Why is that? Maybe it has to do with [changes in visibility of services in Symfony](https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default)? They were public by default in the past, were they not? If yes, why do we need to add an alias with the FQCN as opposed to just making the existing services public? Not saying that's a bad thing, not even saying I didn't do similar things in the past myself, I'm just trying to understand the changes before I approve them.

It is not visibility of service issue. It is more like change in routing system. Something like change `AppBundle:Blog:list` (with auto add Action) in Symfony3 to `AppBundle\Controller\BlogController::listAction` in Symfony4. 

I set public in existing service and it not working but when I create alias based on FQNC all works. 

Next thing is it:
```yaml
# old - api.xml is require
sonata_api_page:
    type: rest
    prefix: /api/page
    resource: "@SonataPageBundle/Resources/config/routing/api.xml"

# new - api.xml is not require
# loads routes from the PHP annotations of the controllers found in that directory
sonata_api_page:
    type: rest # I don't know it is working with rest, but sure it will be with annotation
    prefix: /api/page
    resource: '@SonataPageBundle/Controller/API/'

``` 

@greg0ire 
This is quick fix for it. How to improve it disscusion is under [this issue](https://github.com/sonata-project/dev-kit/issues/778).

 https://symfony.com/doc/4.2/routing/external_resources.html

greg0ire:
Thanks @wbloszyk !

PR #1202, opened 2020-09-22
Bump sonata-project/datagrid-bundle to version ^3.0

wbloszyk:
## Subject

Both datagrid version can be used becouse `PageableInteface` is not used. When we remove deprecated from `PageableManagerInterface` then lower version should be drop with little BC-break.

## Changelog

```markdown
### Added
- support for sonata-project/datagrid-bundle v3
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "sonata-project/block-bundle": "^3.20",
         "sonata-project/cache": "^1.0.3 || ^2.0",
         "sonata-project/cache-bundle": "^2.4.1",
-        "sonata-project/datagrid-bundle": "^2.5",
+        "sonata-project/datagrid-bundle": "^2.5 || ^3.0",
         "sonata-project/doctrine-extensions": "^1.8",
         "sonata-project/doctrine-orm-admin-bundle": "^3.19",
         "sonata-project/form-extensions": "^0.1.1 || ^1.4",


VincentLanglet (approved):

core23 (approved):

phansys (commented):
Shouldn't we add `NEXT_MAJOR` comments where the changes must be done when dropping support for sonata-project/datagrid-bundle < 3?

wbloszyk:
:thinking: https://github.com/sonata-project/SonataMediaBundle/pull/1814#discussion_r492047513

`Sonata\Doctrine\Model\PageableManagerInterface` is deprecated  - `sonata-project/doctrine-exentsions` 
```php
public function getPager(array $criteria, $page, $limit = 10, array $sort = []);
```
`Sonata\DatagridBundle\Pager\PageableInterface`  - `sonata-project/doctrine-bundle` in 2.x 
```php
public function getPager(array $criteria, $page, $limit = 10, array $sort = []);
```
`Sonata\DatagridBundle\Pager\PageableInterface`  - `sonata-project/doctrine-bundle` in 3.x 
```php
public function getPager(array $criteria, int $page, int $limit = 10, array $sort = []): PagerInterface;
```

So unil we use 'getPager()' from `doctrine-extensions` we can drop support for `datagrid-bundle` without BC-break.

phansys:
>sonata-project/SonataMediaBundle#1814 (comment)

In sonata-project/SonataMediaBundle#1814 there is no need to drop support for any dependency or to make any change related to that PR in a next major release, since there is full compatibility with both versions (2 and 3) of aws/aws-sdk-php.

>So until we use 'getPager()' from doctrine-extensions we can drop support for datagrid-bundle without BC-break.

In your description, I understand the change made in this PR can lead to a little BC (which surely will be resolved in the next major). That's why I'm wondering if a `NEXT_MAJOR` comment should be required to address that path:

>When we remove deprecated from PageableManagerInterface then lower version should be drop with little BC-break.

wbloszyk:
> > When we remove deprecated from PageableManagerInterface then lower version should be drop with little BC-break.

Correct. `PageableManagerInterface` is from `doctrine-extensions`. When we remove deprecations, then we must use `datagrid-bundle` which force to update `getPager()` method.

PR #1209, opened 2020-10-01
Replace mentions of 'whitelist' with 'allowlist'

### LINKED ISSUE #1174: Remove oppressive words from code
## Feature Request

We should avoid terms like *blacklist*, *whitliste, *master*, *slave*, ... and use more tech-related words like *blocklist*, *prmary*, *secondary*,...

In this repo, we use the term *whitelist* which could be replaced with *allowlist*:
https://github.com/sonata-project/SonataPageBundle/blob/3.x/assets/src/js/composer.js#L787
https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Resources/views/PageAdmin/compose.html.twig#L57

jlt10:
## Replace mentions of 'whitelist' with 'allowlist'

This repo uses the term _whitelist_ instead of _allowlist_. This PR replaces every instance of "whitelist" in the code with "allowlist".

I am targeting this branch, because this is a backward compatible bug fix.

Closes #1174.


src/Resources/views/PageAdmin/compose.html.twig
```diff
@@ -54,7 +54,7 @@ file that was distributed with this source code.
                                 {% if container.block is defined and container.block %}
                                     <a class="page-composer__page-preview__container block-preview-{{ container.block.id }}"
                                        data-block-id="{{ container.block.id }}"
-                                       data-block-whitelist="{{ container.area.blocks|join(',') }}"
+                                       data-block-allowlist="{{ container.area.blocks|join(',') }}"
```
franmomu, line 57:
```suggestion
                                       data-block-allowlist="{{ container.area.blocks|join(',') }}"
                                       {# NEXT_MAJOR: Remove the following line #}
                                       data-block-whitelist="{{ container.area.blocks|join(',') }}"
```
It's very unlikely to happen, but just in case someone is using this attribute, we could keep it until next major.

jlt10: I see, that makes sense. Should I change the pull request to master?

assets/src/js/composer.js
```diff
@@ -784,15 +784,15 @@
                     revert:            true,
                     connectToSortable: '.page-composer__container__children',
                     accept: function (source) {
-                        var blockWhitelist = $(this).attr('data-block-whitelist');
-                        if (blockWhitelist === '') {
+                        var blockAllowlist = $(this).attr('data-block-allowlist');
```
core23, line 787: Technically speaking, this is a BC break.

I think we could risk this, unless you leave an upgrade note.

core23 (changes requested):

src/Resources/views/PageAdmin/compose.html.twig
```diff
@@ -54,6 +54,8 @@ file that was distributed with this source code.
                                 {% if container.block is defined and container.block %}
                                     <a class="page-composer__page-preview__container block-preview-{{ container.block.id }}"
                                        data-block-id="{{ container.block.id }}"
+                                       data-block-allowlist="{{ container.area.blocks|join(',') }}"
+                                       {# NEXT_MAJOR: Remove the following line #}
                                        data-block-whitelist="{{ container.area.blocks|join(',') }}"
```
core23, line 59: This will not solve the problem when someone is overriding this template.

assets/src/js/composer.js
jlt10: I'm new to this, what is the best way to leave an upgrade note?

VincentLanglet: There is a UPGRADE.md file, you could add an entry to this file to explain the change.

Another solution would be to select
´´´
This.data(whitelist) || this.data(allowlist)
´´´
With a next major to remove the whitelist part

VincentLanglet (approved):

core23 (approved):

VincentLanglet:
Thanks @jlt10

jlt10:
Happy to help! @VincentLanglet do you mind adding the label "hacktoberfest-accepted" to this PR?

PR #1274, opened 2021-01-27
Allow support PHP 8.0

Yozhef:
## Changelog

```markdown
### Added
- Add support for PHP 8.x
```


Yozhef:
Blocked. by https://github.com/sonata-project/SonataCacheBundle/pull/309

jordisala1991:
Blocked because requires cache bundle v2 and that is not maintained anymore, compat for php8 is on cache-bundle v3.

One step could be try to add compat with v3 (dont know if that is possible because of type hints), the other way is to remove that integration and move to use symfony cache. This option will require next major tho.

wdyt @sonata-project/contributors

phansys:
I'm a little confused about the status of `sonata-project/cache-bundle`. Is it deprecated or not? sonata-project/SonataCacheBundle#279.
BTW, I agree with the proposal of using `symfony/cache` as  replacement.

VincentLanglet:
> I'm a little confused about the status of `sonata-project/cache-bundle`. Is it deprecated or not? [sonata-project/SonataCacheBundle#279](https://github.com/sonata-project/SonataCacheBundle/issues/279).

It's deprecated, but since some bundle are still using it, we're still maintaining it.

wbloszyk:
> Blocked because requires cache bundle v2 and that is not maintained anymore, compat for php8 is on cache-bundle v3.
> 
> One step could be try to add compat with v3 (dont know if that is possible because of type hints), the other way is to remove that integration and move to use symfony cache. This option will require next major tho.
> 
> wdyt @sonata-project/contributors

IMO the best way will add PHP 8.0 support in cache-bundle v2. We can bump Symfony to ^4.0 too. Oher solution will require provide BC-break.

PR #1284, opened 2021-03-06
Remove admin deprecations

core23:
## Subject

I am targeting this branch, because this is BC.

## Changelog

```markdown
### Removed
- Remove admin deprecations
```

PR #1295, opened 2021-04-09
Fixes parent call and return int in commands

franmomu:
## Subject

We missed in https://github.com/sonata-project/SonataPageBundle/pull/1284 the parent call, otherwise `PageAdmin` overrides:

https://github.com/sonata-project/SonataPageBundle/blob/9f808b3608feef0ec78752051a6f3dcc0f900bb0/src/Admin/BaseBlockAdmin.php#L228-L237

## Changelog

```markdown
### Fixed
- Fixed deprecations using commands because of not returning `int`
```

### DIFF
--- a/src/Admin/BlockAdmin.php
+++ b/src/Admin/BlockAdmin.php
@@ -62,7 +62,7 @@ class BlockAdmin extends BaseBlockAdmin
 
     protected function configurePersistentParameters(): array
     {
-        $parameters = [];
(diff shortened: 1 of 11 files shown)


VincentLanglet (approved):

core23 (approved):

VincentLanglet:
Thanks @franmomu

Issue #513, opened 2015-07-30
ChildrenPagesBlockService doesn't save page


jaroslavas:
When using children pages block, only the current page works. If I set another page, it isn't saved. As far as I debugged it, pageId doesn't even gets set in prePersist/preUpdate methods.

composer.json:

```
...
        "sonata-project/admin-bundle": "^2.3",
        "sonata-project/doctrine-orm-admin-bundle": "^2.3",
        "sonata-project/cache-bundle": "^2.2",
        "sonata-project/seo-bundle": "~1.1",
        "sonata-project/easy-extends-bundle": "^2.1",
        "sonata-project/notification-bundle": "^2.3",
        "sonata-project/datagrid-bundle": "dev-postgre-fix as 2.2.*-dev",
        "sonata-project/page-bundle": "^2.3",
        "sonata-project/user-bundle": "^2.2",
        "sonata-project/formatter-bundle": "^2.3",
        "sonata-project/markitup-bundle": "^2.1",
        "sonata-project/news-bundle": "^2.3",
        "sonata-project/classification-bundle": "^2.2",
        "symfony-cmf/routing-bundle": "^1.3",
        "mopa/bootstrap-bundle": "^2.3",
        "twbs/bootstrap": "^3.3",
        "sonata-project/media-bundle": "^2.3",
        "sonata-project/intl-bundle": "^2.2",
...
```

core23:
Can you provide more information? Please use the new issue template too  @jaroslavas

nnlugovoy:
The problem is here ` 'class' => $formMapper->getAdmin()->getClass(),` - on update form tries to convert settings.pageId parameter to *Block* object, instead of *Page*.
```php
           array('pageId', 'sonata_page_selector', array(
                    'model_manager' => $formMapper->getAdmin()->getModelManager(),
                    'class' => $formMapper->getAdmin()->getClass(),
                    'site' => $block->getPage()->getSite(),
                    'required' => false,
                    'label' => 'form.label_page',
                )),
```
also, if I provide for example specific `'class' => 'PageBundle\\Entity\\Page'` then receive error
``` No entity manager defined for class Sonata\PageBundle\Model\SnapshotPageProxy```
due to load in 
```php
    public function load(BlockInterface $block)
    {
        if (is_numeric($block->getSetting('pageId', null))) {
            $cmsManager = $this->cmsManagerSelector->retrieve();
            $site = $block->getPage()->getSite();
[...]

PR #1297, opened 2021-04-13
Remove usage of sonata_admin twig global variable

### LINKED ISSUE #1282: Remove `sonata_admin` deprecations in twig templates
## Feature Request

The global twig constant `sonata_admin` was deprecated.

Refs: https://github.com/sonata-project/SonataAdminBundle/pull/6640#pullrequestreview-602353439

franmomu:
## Subject

In order to fix #1282, I've added a `sonata_page_admin` variable which holds the `sonata.page.admin.page` service.

I am targeting this branch, because these changes are BC.

Closes #1282 and #1299.

## Changelog

```markdown
### Added
- Added `sonata_page_admin` Twig global variable which holds `sonata.page.admin.page` service
### Removed
- Removed deprecations from `sonata-project/admin-bundle` using `sonata_admin` Twig global variable
```

### DIFF
--- a/src/DependencyInjection/Compiler/GlobalVariablesCompilerPass.php
+++ b/src/DependencyInjection/Compiler/GlobalVariablesCompilerPass.php
@@ -26,7 +26,12 @@ class GlobalVariablesCompilerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
+        if (!$container->hasDefinition('twig')) {
+            return;
+        }
+
         $container->getDefinition('twig')
-            ->addMethodCall('addGlobal', ['sonata_page', new Reference('sonata.page.twig.global')]);
+            ->addMethodCall('addGlobal', ['sonata_page', new Reference('sonata.page.twig.global')])
+            ->addMethodCall('addGlobal', ['sonata_page_admin', new Reference('sonata.page.admin.page')]);
     }
 }
--- a/src/Resources/views/base_layout.html.twig
+++ b/src/Resources/views/base_layout.html.twig
@@ -50,15 +50,21 @@ file that was distributed with this source code.
         {% block sonata_page_top_bar %}
             {% block page_top_bar %} {# Deprecated block #}
                 {% if sonata_page.isEditor or ( app.user and is_granted('ROLE_PREVIOUS_ADMIN') ) %}
-
+                    {#
+                        The "page" variable is an instance of SnapshotPageProxy and cannot be used in admin service calls
+                        which expect an instance of the Page model.
+                    #}
                     {% if sonata_page.isEditor and sonata_page.isInlineEditionOn %}
                         <!-- CMS specific variables -->
                         <script>
                             jQuery(document).ready(function() {
                                 Sonata.Page.init({
                                     url: {
-                                        block_save_position: '{{ sonata_admin.objectUrl('sonata.page.admin.page', 'edit', page) }}',
-                                        block_edit:          '{{ sonata_admin.url('sonata.page.admin.page|sonata.page.admin.block', 'edit', {'id': 'BLOCK_ID'}) }}'
+                                        block_save_position: '{{ sonata_page_admin.generateUrl('edit', {(sonata_page_admin.idParameter): page.id}) }}',
+                                        block_edit:          '{{ sonata_page_admin.generateUrl('sonata.page.admin.page|sonata.page.admin.block.edit', {
+                                            (sonata_page_admin.idParameter): page.id,
+                                            (sonata_page_admin.getChild('sonata.page.admin.block').idParameter): 'BLOCK_ID'
+                                        }) }}
                                     }
                                 });
                             });
@@ -69,8 +75,11 @@ file that was distributed with this source code.
                     {% if sonata_page.isEditor and sonata_page.isInlineEditionOn %}
                         data-page-editor='{{ {
                             url: {
-                                block_save_position: sonata_admin.objectUrl('sonata.page.admin.page', 'edit', page),
-                                block_edit:          sonata_admin.url('sonata.page.admin.page|sonata.page.admin.block', 'edit', {'id': 'BLOCK_ID'})
+                                block_save_position: sonata_page_admin.generateUrl('edit', {(sonata_page_admin.idParameter): page.id}),
+                                block_edit:          sonata_page_admin.generateUrl('sonata.page.admin.page|sonata.page.admin.block.edit', {
+                                    (sonata_page_admin.idParameter): page.id,
+                                    (sonata_page_admin.getChild('sonata.page.admin.block').idParameter): 'BLOCK_ID'
+                                })
                             }
                         }|json_encode()|raw }}'
                     {% endif %}>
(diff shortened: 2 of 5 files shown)


src/Resources/views/base_layout.html.twig
```diff
@@ -57,8 +57,8 @@ file that was distributed with this source code.
                             jQuery(document).ready(function() {
                                 Sonata.Page.init({
                                     url: {
-                                        block_save_position: '{{ sonata_admin.objectUrl('sonata.page.admin.page', 'edit', page) }}',
-                                        block_edit:          '{{ sonata_admin.url('sonata.page.admin.page|sonata.page.admin.block', 'edit', {'id': 'BLOCK_ID'}) }}'
+                                        block_save_position: '{{ sonata_page_admin.generateObjectUrl('edit', page) }}',
+                                        block_edit:          '{{ sonata_page_admin.generateUrl('sonata.page.admin.page|sonata.page.admin.block.edit', {'id': page.id, 'childId': 'BLOCK_ID'})
```
franmomu, line 61: I've been not able to test this, but given `'sonata.page.admin.page|sonata.page.admin.block.edit'` route contains a reference of `page` and `block`, I guess it should have `id` and `childId`, I'm not sure if this was working before.

If you can give it a try @core23 or someone using the bundle.

VincentLanglet (dismissed):

src/Resources/views/base_layout.html.twig
phansys: I'm not using this bundle either, but should not we use `admin.idParameter` for these route parameters instead of having them hardcoded? Something like:
```twig
{% set sonata_page_block_admin = sonata_admin.adminPool
    .getAdminByAdminCode('sonata.page.admin.page|sonata.page.admin.block')
%}
{{
    sonata_page_block_admin.generateUrl(
        'edit',
        {
            (sonata_page_block_admin.idParameter): 'BLOCK_ID',
            (sonata_page_block_admin.parent.idParameter): sonata_page_block_admin.parent.id(sonata_page_block_admin.parent.subject) 
        }
    )
}}
```
BTW, the closing ` }}'` is missing in https://github.com/sonata-project/SonataPageBundle/pull/1297#discussion_r612153544.

franmomu: Oops thanks! we should probably use a twig linter 🤦 

I've changed the code a bit to not use `sonata_admin`.

VincentLanglet (approved):

core23 (dismissed):

jordisala1991 (approved):

src/Resources/views/base_layout.html.twig
phansys: >Oops thanks! we should probably use a twig linter facepalm

Yes, I think we could catch syntax errors faster.

>I've changed the code a bit to not use sonata_admin.

That's great. Did you have chance to analyze my concern about the hardcoded route parameters?

src/Resources/views/base_layout.html.twig
```diff
@@ -57,8 +57,11 @@ file that was distributed with this source code.
                             jQuery(document).ready(function() {
                                 Sonata.Page.init({
                                     url: {
-                                        block_save_position: '{{ sonata_admin.objectUrl('sonata.page.admin.page', 'edit', page) }}',
-                                        block_edit:          '{{ sonata_admin.url('sonata.page.admin.page|sonata.page.admin.block', 'edit', {'id': 'BLOCK_ID'}) }}'
+                                        block_save_position: '{{ sonata_page_admin.generateObjectUrl('edit', page) }}',
+                                        block_edit:          '{{ sonata_page_admin.generateUrl('sonata.page.admin.page|sonata.page.admin.block.edit', {
+                                            (sonata_page_admin.idParameter): page.id,
```
phansys, line 62:
```suggestion
                                            (sonata_page_admin.idParameter): sonata_page_admin.id(page),
```

src/Resources/views/base_layout.html.twig
```diff
@@ -69,8 +72,11 @@ file that was distributed with this source code.
                     {% if sonata_page.isEditor and sonata_page.isInlineEditionOn %}
                         data-page-editor='{{ {
                             url: {
-                                block_save_position: sonata_admin.objectUrl('sonata.page.admin.page', 'edit', page),
-                                block_edit:          sonata_admin.url('sonata.page.admin.page|sonata.page.admin.block', 'edit', {'id': 'BLOCK_ID'})
+                                block_save_position: sonata_page_admin.generateObjectUrl('edit', page),
+                                block_edit:          sonata_page_admin.generateUrl('sonata.page.admin.page|sonata.page.admin.block.edit', {
+                                    (sonata_page_admin.idParameter): page.id,
```
phansys, line 77:
```suggestion
                                    (sonata_page_admin.idParameter): sonata_page_admin.id(page),
```

src/Resources/views/base_layout.html.twig
```diff
@@ -114,7 +120,7 @@ file that was distributed with this source code.
 
                                     {% if page is defined %}
                                         <li>
-                                            <a href="{{ sonata_admin.url('sonata.page.admin.page', 'compose', {'id': page.id}) }}">
+                                            <a href="{{ sonata_page_admin.generateUrl('compose', {'id': page.id}) }}">
```
phansys, line 123: IMO, we should not hardcode the parameter name for the identifier nor how its value is obtained.
`AdminInterface` exposes the `getIdParameter()` and `id()` methods for these values.

franmomu: I agree, but trying to test this, something else has arisen, looks like this `page` variable is not an instance of let's say `App\Entity\Page`, it's an instance of `Sonata\PageBundle\Model\SnapshotPageProxy`, so it throws:

```
An exception has been thrown during the rendering of a template ("No entity manager defined for class Sonata\PageBundle\Model\SnapshotPageProxy").
```

Maybe related of https://github.com/sonata-project/SonataPageBundle/issues/513

phansys: In that case, I think we should leave a comment exposing the problem and open a issue (if it isn't reported yet).

franmomu: ok, so I've ended up changing all the `objectUrl`|`generateObjectUrl` calls that failed by `generateUrl` and that should address the issues with that `SnapshotPageProxy` problem and added a comment.

VincentLanglet (approved):

src/Resources/views/base_layout.html.twig
```diff
@@ -50,15 +50,21 @@ file that was distributed with this source code.
         {% block sonata_page_top_bar %}
             {% block page_top_bar %} {# Deprecated block #}
                 {% if sonata_page.isEditor or ( app.user and is_granted('ROLE_PREVIOUS_ADMIN') ) %}
-
+                    {#
```
phansys, line 53: I think we should add this comment in every template where we can not use `admin.id()`.

phansys (approved):

src/Resources/views/base_layout.html.twig
franmomu: To be honest, I have no idea where is that. 

In 
https://github.com/sonata-project/SonataPageBundle/blob/78ae79486e5caf849f241bd14d97040106a853d5/src/Resources/views/Block/block_pagelist.html.twig#L24

I've changed it to `generateObjectUrl` because they come from 

https://github.com/sonata-project/SonataPageBundle/blob/78ae79486e5caf849f241bd14d97040106a853d5/src/Block/PageListBlockService.php#L85-L97
and they are actual "pages".

core23 (approved):

src/DependencyInjection/Compiler/GlobalVariablesCompilerPass.php
```diff
@@ -26,7 +26,12 @@ class GlobalVariablesCompilerPass implements CompilerPassInterface
...
     public function process(ContainerBuilder $container)
     {
+        if (!$container->hasDefinition('twig')) {
+            return;
+        }
+
         $container->getDefinition('twig')
-            ->addMethodCall('addGlobal', ['sonata_page', new Reference('sonata.page.twig.global')]);
+            ->addMethodCall('addGlobal', ['sonata_page', new Reference('sonata.page.twig.global')])
+            ->addMethodCall('addGlobal', ['sonata_page_admin', new Reference('sonata.page.admin.page')]);
```
core23, line 35: The `SonataAdminBundle` is an optional dependency, so this can be null @VincentLanglet

VincentLanglet: Why is it optional ? It's in the require
https://github.com/sonata-project/SonataPageBundle/blob/3.x/composer.json#L28

VincentLanglet: cc @franmomu

franmomu: Fixed in https://github.com/sonata-project/SonataPageBundle/pull/1310, the services are loaded if `SonataAdminBundle` is available:

https://github.com/sonata-project/SonataPageBundle/blob/654fc11cffe2131657cf6e26d84c7a5adf32d344/src/DependencyInjection/SonataPageExtension.php#L77-L85

but I guess another PR is needed to make it really optional.

Issue #1096, opened 2019-11-30
Remove Sonata Cache


core23:
We should use symfony cache instead.

Even the doctrine cache was just deprecated in favor of symfony cache:
https://github.com/doctrine/DoctrineCacheBundle#deprecation-warning

eerison:
@VincentLanglet can you add this in https://github.com/sonata-project/SonataPageBundle/milestone/1

please :)

Hanmac:
I did some brain storming about Removing Sonata Cache(Bundle) from Page Bundle.

* Block [ESI Cache](https://symfony.com/doc/current/http_cache/esi.html) and Block [SSI Cache](https://symfony.com/doc/4.4/http_cache/ssi.html) use features more related to symfony http cache than [symfony cache](https://symfony.com/doc/current/components/cache.html).
* I think the change should be done in two parts:
  * Make Page Bundle use [PSR-6](https://www.php-fig.org/psr/psr-6/) or [PSR-16](https://www.php-fig.org/psr/psr-16/) Cache Interface to store rendered blocks if wanted. (What Cache PSR should be prefered?) Symfony Cache is only an implementation of these two.
  * Make the ESI/SSI Cache be able use the previous PSR-6/16 cached page content and wrap it into ESI/SSI

i don't know yet how it is controlled which pages/blocks are cached in which way.

I need to say that i didn't used page bundle yet, so that's just my thoughts of uninvolved.

eerison:
> I did some brain storming about Removing Sonata Cache(Bundle) from Page Bundle.
> 
> * Block [ESI Cache](https://symfony.com/doc/current/http_cache/esi.html) and Block [SSI Cache](https://symfony.com/doc/4.4/http_cache/ssi.html) use features more related to symfony http cache than [symfony cache](https://symfony.com/doc/current/components/cache.html).
> * I think the change should be done in two parts:
>   
>   * Make Page Bundle use [PSR-6](https://www.php-fig.org/psr/psr-6/) or [PSR-16](https://www.php-fig.org/psr/psr-16/) Cache Interface to store rendered blocks if wanted. (What Cache PSR should be prefered?) Symfony Cache is only an implementation of these two.
>   * Make the ESI/SSI Cache be able use the previous PSR-6/16 cached page content and wrap it into ESI/SSI
> 
> i don't know yet how it is controlled which pages/blocks are cached in which way.
> 
[...]

PR #1303, opened 2021-04-20
Support php 8

4c0n:
## Subject

Support php 8

I am targeting this branch, because I need the current version to support PHP 8.

## Changelog

Support PHP 8.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,6 +24,7 @@ jobs:
                 php-version:
                     - '7.3'
                     - '7.4'
+                    - '8.0'
                 dependencies: [highest]
                 allowed-to-fail: [false]
                 variant: [normal]
@@ -62,10 +63,6 @@ jobs:
 
             - name: Add PHPUnit matcher
               run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
-
-            - name: Configuration required for PHP 8.0
-              if: matrix.php-version == '8.0'
-              run: composer config platform.php 7.4.99
 
             - name: Install variant
               if: matrix.variant != 'normal'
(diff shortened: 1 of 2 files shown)


VincentLanglet:
The sonata-project/cache-bundle need to support PHP8 first

4c0n:
@VincentLanglet Actually it seems that it already does: https://github.com/sonata-project/SonataCacheBundle/blob/d0b13b4dcc85272f2235e7fbd50578dd403222b0/composer.json#L22

However this bundle should support the newer version of that bundle:
https://github.com/sonata-project/SonataPageBundle/blob/fed26c06856de8d6b1161d0f23001db7b8377e2b/composer.json#L31

VincentLanglet:
Oh yes indeed, it required the sonata-project/cache-bundle 3.
Another option is to stop using the cache bundle, since it's deprecated.

https://github.com/sonata-project/SonataPageBundle/issues/1096
https://github.com/sonata-project/SonataCacheBundle/issues/279

4c0n:
@VincentLanglet Is that not a BC break? I'd rather remove it to be honest, but I guess it should be replaced with some integration of the standard Symfony cache support?

VincentLanglet:
> @VincentLanglet Is that not a BC break? I'd rather remove it to be honest, but I guess it should be replaced with some integration of the standard Symfony cache support?

It depends how it's done, but I'm not sure. I know nothing about this bundle.

Another easy option could be to add support for php8 to the cache bundle v2.

4c0n:
@VincentLanglet turns out it wasn't too hard to use version 3: https://github.com/sonata-project/SonataPageBundle/pull/1304
It's probably better to migrate away from the cache bundle to the default Symfony cache in a new major version.

PR #1196, opened 2020-09-02
Update conflict constraint for "nelmio/api-doc-bundle" in order to allow version 3

wbloszyk:
## Subject
Update conflict constraint for "nelmio/api-doc-bundle" in order to allow version 3.

I am targeting this branch, because these changes almost respect BC (with one, very little BC-break potential).

## Changelog

```markdown
### Added
- support for `nelmio/api-doc-bundle` >= 3.6
### Changed
- Updated docs in order to expose how to configure custom serialization paths under jms_serializer configuration node (jms_serializer.metadata.directories).
```

PR #1304, opened 2021-04-20
Upgrade cache bundle to version 3

4c0n:
## Subject

Upgrading to the latest cache bundle version.
I realize it is deprectated, but it is currently supported and this seems to be the most natural path to take before removing the dependency completely.
The reason I would like to upgrade the cache bundle, is because version 3 supports PHP 8.

I am targeting this branch, because I would like the current version to support PHP 8.

Related to: https://github.com/sonata-project/SonataPageBundle/pull/1303

## Changelog

Refactored cache classes to reflect the new cache bundle version.


VincentLanglet:
Changing the constructor args are BC-break...

@sonata-project/contributors What should we do ? Introducing a BC break to support php 8 ?
Should we instead add php 8 support to the 2.x version of SonataCacheBundle ?

jordisala1991:
technically it is bc break but, in order to upgrade to new 3.x version, you will also need to upgrade cache-bundle. And if you override any of this class... you should probably add cache-bundle to your composer.json, so you will end up upgrading cache-bundle and because of that you will also upgrade your clases.

IMO if this is the only bundle missing an upgrade of this package, I would allow this bc break to allow moving to better support...

VincentLanglet:
> technically it is bc break but, in order to upgrade to new 3.x version, you will also need to upgrade cache-bundle. And if you override any of this class... you should probably add cache-bundle to your composer.json, so you will end up upgrading cache-bundle and because of that you will also upgrade your clases.

Yes indeed.
Let's add an upgrade note and merge this ! :)

VincentLanglet:
@4c0n Can you add some explanations in https://github.com/sonata-project/SonataPageBundle/blob/3.x/UPGRADE-3.x.md about the introduced breaking change ?

VincentLanglet:
You have an example in this PR: https://github.com/sonata-project/SonataPageBundle/pull/1196
You should also use the following format for the Changelog in your first PR (the test will be used automatically for the release)

```markdown
### Added
- support for `nelmio/api-doc-bundle` >= 3.6
### Changed
- Updated docs in order to expose how to configure custom serialization paths under jms_serializer configuration node (jms_serializer.metadata.directories).
```

VincentLanglet:
Hi @4c0n, are you still interested in this PR ?

4c0n:
@VincentLanglet My apologies for leaving this hanging. I got very busy and ended up not using this bundle anymore.
Sorry I kind of forgot about it and don't have any urgent needs to get this done anymore :)

VincentLanglet:
@Yozhef, you were interested by PHP8 support (https://github.com/sonata-project/SonataPageBundle/pull/1274), would you be interested to finish this PR ?

PR #1305, opened 2021-04-20
There is a direct dependency on the jms serializer bundle in api_form.xml

4c0n:
## Subject

There is a direct dependency on the jms serializer bundle in api_form.xml, so I moved the bundle from the dev dependencies into the dependencies section.

The dependency is there for all environments and will cause issues if the `jms_serializer.metadata_factory` service id does not exist.

I am targeting this branch, because it is the current version.

## Changelog

```markdown
### Fixed
- Do not load `api_form.xml` if `JMSSerializerBundle` is not installed
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "cocur/slugify": "^3.0 || ^4.0",
         "doctrine/doctrine-bundle": "^1.12.3 || ^2.0",
         "doctrine/persistence": "^1.3.4 || ^2.0",
+        "jms/serializer-bundle": "^2.0 || ^3.0",
         "sonata-project/admin-bundle": "^3.91.1",
         "sonata-project/block-bundle": "^3.20",
         "sonata-project/cache": "^1.0.3 || ^2.0",
@@ -62,7 +63,6 @@
     "require-dev": {
         "doctrine/annotations": "^1.10.3",
         "friendsofsymfony/rest-bundle": "^2.1 || ^3.0",
-        "jms/serializer-bundle": "^2.0 || ^3.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1.1",
         "nelmio/api-doc-bundle": "^2.4",
         "symfony/browser-kit": "^4.4",
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,6 @@
         "cocur/slugify": "^3.0 || ^4.0",
         "doctrine/doctrine-bundle": "^1.12.3 || ^2.0",
         "doctrine/persistence": "^1.3.4 || ^2.0",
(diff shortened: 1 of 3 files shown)


4c0n:
https://github.com/sonata-project/SonataPageBundle/blob/fed26c06856de8d6b1161d0f23001db7b8377e2b/src/Resources/config/api_form.xml#L6

VincentLanglet (approved):

franmomu:
`api_form.xml` only gets loaded when 

https://github.com/sonata-project/SonataPageBundle/blob/fed26c06856de8d6b1161d0f23001db7b8377e2b/src/DependencyInjection/SonataPageExtension.php#L70-L75

I think we could also add a check for `JMSSerializerBundle` instead of move it to the `require` section.

4c0n:
@franmomu That makes sense, hope this behavior is well documented somewhere, as it isn't great when services in a bundle "magically" appear or disappear based on the configuration of another bundle. Actually for a long time having one bundle depend on another bundle was considered bad practice, maybe that has changed?

Another approach would be to depend on the components that the other bundle uses and define the services in this bundle. That way there is no need to depend on the other bundle for its services.
Let me know what you think :)

core23 (changes requested):
The api is an optional dependency:
https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/DependencyInjection/SonataPageExtension.php#L70-L75

core23:
> @franmomu That makes sense, hope this behavior is well documented somewhere, as it isn't great when services in a bundle "magically" appear or disappear based on the configuration of another bundle. Actually for a long time having one bundle depend on another bundle was considered bad practice, maybe that has changed?

:+1: 

IMHO the best way is a separate bundle which adds an API layer to this bundle, but it would require some more refactoring. Using this `magic` thing is the current solution across all our bundles.

franmomu:
It should be documented for sure.

About dependencies, in an ideal world, we could maybe remove the API part and create a package with the code and a bundle maybe that integrates the API part to this bundle, but that would be a lot of work and maintenance.

The thing about components, IMO, the problem I see is that usually those services depend on another services, so you would have to create (and maintain) several services that don't really add value, instead of using it. And also sometimes this services are configurable and you want to configure them from the original bundle.

core23 (dismissed):

VincentLanglet (dismissed):

VincentLanglet:
tests need to be fixed

jordisala1991 (approved):

VincentLanglet (approved):

VincentLanglet:
Please review @core23 :)

core23 (approved):

greg0ire:
Thanks @4c0n

PR #1370, opened 2021-10-14
Upgrade to sonata 4

core23:
## Subject

I am targeting this branch, because this is a BC break.

## Changelog

```markdown
### Added
- Add compatiblity with all SonataAdmin 4.0 versions
```

## To do

- [ ] Open block bundle classes for extension
- [x] Require seo-bundle 3


VincentLanglet:
> * [ ]  Open block bundle classes for extension

Shouldn't we prefer injection/composition instead ?

core23:
I'm not using this bundle anymore.

haivala:
@core23 just out of curiosity: what are you using now for similar tasks?

core23:
Simple symfony controller with custom entities

PR #1382, opened 2021-11-16
Configure PHPStan and Psalm

jordisala1991:
## Subject

## Changelog

```markdown
### Added
- Added PHPStan and Psalm.
```


jordisala1991:
I do not know how to fix the last error:

```
 ------ --------------------------------------------------------------------- 
  Line   src/Entity/SnapshotManager.php                                       
 ------ --------------------------------------------------------------------- 
  184    Class Sonata\PageBundle\Model\SnapshotPageProxy constructor invoked  
         with 2 parameters, 3 required.                                       
 ------ --------------------------------------------------------------------- 
```

Appears both on phpstan and psalm, any ideas @core23 @VincentLanglet ?

VincentLanglet:
> I do not know how to fix the last error:
> 
> ```
>  ------ --------------------------------------------------------------------- 
>   Line   src/Entity/SnapshotManager.php                                       
>  ------ --------------------------------------------------------------------- 
>   184    Class Sonata\PageBundle\Model\SnapshotPageProxy constructor invoked  
>          with 2 parameters, 3 required.                                       
>  ------ --------------------------------------------------------------------- 
> ```
> 
> Appears both on phpstan and psalm, any ideas @core23 @VincentLanglet ?

Phpstan and psalm are right, the SnapshotPageProxy require three arguments
https://github.com/sonata-project/SonataPageBundle/blob/e91edae879bc15fadde09b8be7effa20a719b535/src/Model/SnapshotPageProxy.php#L58

A SnapshotManager, `$this`, a transformer (we're passing none). and a snapshot `$snapshot`.

This code seems broken. Since 2013
https://github.com/sonata-project/SonataPageBundle/commit/4152f9069b5ce5d3b1b9e9c7f594af4d45e23ce7#

composer.json
```diff
@@ -65,10 +65,17 @@
         "jms/serializer-bundle": "^2.0 || ^3.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1.1",
         "nelmio/api-doc-bundle": "^2.4",
+        "phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan": "^1.0",
+        "phpstan/phpstan-phpunit": "^1.0",
+        "phpstan/phpstan-symfony": "^1.0",
```
VincentLanglet, line 71: phpstan strict rules too ?

jordisala1991: I would like to but they generate too many errors, probably is safer to do it on next major instead to avoid bugs on stable releases.

jordisala1991:
> > I do not know how to fix the last error:
> > ```
> >  ------ --------------------------------------------------------------------- 
> >   Line   src/Entity/SnapshotManager.php                                       
> >  ------ --------------------------------------------------------------------- 
> >   184    Class Sonata\PageBundle\Model\SnapshotPageProxy constructor invoked  
> >          with 2 parameters, 3 required.                                       
> >  ------ --------------------------------------------------------------------- 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Appears both on phpstan and psalm, any ideas @core23 @VincentLanglet ?
> 
> Phpstan and psalm are right, the SnapshotPageProxy require three arguments
> 
> https://github.com/sonata-project/SonataPageBundle/blob/e91edae879bc15fadde09b8be7effa20a719b535/src/Model/SnapshotPageProxy.php#L58
> 
> A SnapshotManager, `$this`, a transformer (we're passing none). and a snapshot `$snapshot`.
> 
> This code seems broken. Since 2013 [4152f90](https://github.com/sonata-project/SonataPageBundle/commit/4152f9069b5ce5d3b1b9e9c7f594af4d45e23ce7)

I added a ignore on both tools because the method is deprecated and I think there is no need to add a parameter and a BC change to fix something that is deprecated anyway.

jordisala1991:
next levels and phpstan strict rules should be done on 4.x IMO, there are too many deprecated things this bundle is using.

VincentLanglet (approved):

core23 (approved):

gremo:
Please please please pay attention with such "innocent" updates. You added `protected $id` to the `Model/Site` class,  potentially breaking thousands with a "patch" updated.

Thanks.

jordisala1991:
If you want to open a pr to make them public it will be welcome.

gremo:
Sorry @jordisala1991 but I think I'm not in a position to help with that. I didn't looked at this PR in deep, so I can't understand why `protected $id` was added _in the first place_, without relying on a `public function getId()` instead 🤕 
 
Don't want to break something 😄

jordisala1991:
> Sorry @jordisala1991 but I think I'm not in a position to help with that. I didn't looked at this PR in deep, so I can't understand why `protected $id` was added _in the first place_, without relying on a `public function getId()` instead 🤕 
> 
>  
> 
> Don't want to break something 😄 

Then it will stay this way I guess, we are working on the 4.x branch.

PR #1392, opened 2021-12-09
Closing API

core23:
## Subject

## Changelog

```markdown
### Added
- Add `SnapshotInterface::getId()` method

### Changed
- Mark all classes as final
```

### DIFF
--- a/src/Admin/BlockAdmin.php
+++ b/src/Admin/BlockAdmin.php
@@ -32,6 +32,8 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
  * Admin class for the Block model.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @final since sonata-project/page-bundle 3.x
  */
 class BlockAdmin extends BaseBlockAdmin
 {
--- a/src/Model/SnapshotInterface.php
+++ b/src/Model/SnapshotInterface.php
@@ -18,6 +18,7 @@ namespace Sonata\PageBundle\Model;
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
+ * @method int|null           getId()
  * @method int|null           getTargetId()
  * @method void               setTargetId(?int $targetId)
  * @method PageInterface|null getTarget()
--- a/tests/Entity/SnapshotManagerTest.php
+++ b/tests/Entity/SnapshotManagerTest.php
@@ -29,7 +29,7 @@ use Sonata\PageBundle\Model\SnapshotPageProxyFactoryInterface;
 use Sonata\PageBundle\Model\SnapshotPageProxyInterface;
 use Sonata\PageBundle\Model\TransformerInterface;
 
-class SnapshotManagerTest extends TestCase
+final class SnapshotManagerTest extends TestCase
 {
     public function testSetTemplates(): void
     {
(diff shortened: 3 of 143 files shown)


src/Admin/BlockAdmin.php
```diff
@@ -32,6 +32,8 @@
  * Admin class for the Block model.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @final
```
jordisala1991:
```suggestion
 * @final since sonata-project/page-bundle 3.x
```

Same apply for the rest of final keywords.

tests/Entity/SnapshotManagerTest.php
```diff
@@ -226,7 +226,11 @@ public function testEnableSnapshots(): void
         $page = $this->createMock(PageInterface::class);
         $page->expects(static::once())->method('getId')->willReturn(456);
 
-        $snapshot = $this->createMock(Snapshot::class);
+        $snapshot = $this->getMockBuilder(SnapshotInterface::class)
+            ->disableOriginalConstructor()
+            ->addMethods(['getId'])
+            ->getMockForAbstractClass();
+        // @phpstan-ignore-next-line
```
VincentLanglet: If we need this method, we should add it to the interface.

VincentLanglet (changes requested):

tests/Entity/SnapshotManagerTest.php
core23: Not possible on the stable branch...

VincentLanglet: At least `@method` should be added https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Model/SnapshotInterface.php#L28
And a next major can be added in the test to update the code in the test.

But in this special case, I'm not sure we can consider that adding `getId` to the interface is a BC-break.
It's called here: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Entity/SnapshotManager.php#L91, so the method must be implemented anyway.

core23: done

VincentLanglet: the phpstan ignore is not needed anymore i think

VincentLanglet (approved):

core23:
Ping @sonata-project/contributors

VincentLanglet:
Can you take a look @jordisala1991 ?

src/Model/SnapshotInterface.php
```diff
@@ -18,6 +18,7 @@
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
+ * @method int|null           getId()
```
franmomu, line 21: Should this be `int`? or `mixed`?

Based on `PageInterface`

https://github.com/sonata-project/SonataPageBundle/blob/d102ce8843f3fb9456ce7d44d3af1194c959e8cb/src/Model/PageInterface.php#L27-L32

(I don't know if TargetId should be `mixed` as well)

core23: The `parentId` is a int type at the moment: https://github.com/sonata-project/SonataPageBundle/blob/fed26c06856de8d6b1161d0f23001db7b8377e2b/src/Model/SnapshotInterface.php#L26

If we change this, we should change it everywhere.

franmomu (approved):

VincentLanglet:
Thanks

Issue #1415, opened 2022-04-22
Remove SonataNotificationBundle from dependencies and code


eerison:
## Remove SonataNotificationBundle 

As discussed in #1410 , we need to remove `SonataNotificationBundle` from the next major release

- [x] #1418 
- [x] #1434
- [x] #1460 

**Create tickets**
- [x] #1433 Create a ticket to remove all `consume code`
- [x] #1453 
- [x]  remember to `set self deprecation = 0`
- [x] #1426

eerison:
I'm on it!

eerison:
Hey @core23 

I didn't see your Issue regarding `NotificationBundle` #1238 , my bad

https://github.com/sonata-project/SonataPageBundle/issues/1238#issue-749833646

But if I understand correct we do not need to use symfony/messager component, or is it necessary?

could you give some example(s) how was you thinking to remove this decency?, I guess you understand this functionality better then me 😄

eerison:
> The original idea was to allow async snapshot creation / cleanup. 

do you mean create and remove snapshots from the database?

> To allow deprecating the usage, we could introduce a new SnapshotManagerInterface with two methods createSnapshotsForSite and cleanupSnashotsForSite.

do you know where we are managing this logical (to create and remove snapshots) using NotificationBundle? just to have a better understand :)  


Edit 1: Hmmm I saw that it was already created `SnapshotManagerInterface` and `SnapshotManager`, I'll take a look why we are still using NotificationBundle in the code ;)

eerison:
Started to deprecate NotificationBundle in #1418

core23:
> But if I understand correct we do not need to use symfony/messager component, or is it necessary?
> 
> could you give some example(s) how was you thinking to remove this decency?, I guess you understand this functionality better then me 😄

The idea of my issue was to remove the whole async processing, because a "normal" website should be able to create snapshots on the fly. 

The notification bundle or messager component adds a complexity to this bundle that is not necessary.

eerison:
> > But if I understand correct we do not need to use symfony/messager component, or is it necessary?
> > could you give some example(s) how was you thinking to remove this decency?, I guess you understand this functionality better then me 😄
> 
> The idea of my issue was to remove the whole async processing, because a "normal" website should be able to create snapshots on the fly.
> 
> The notification bundle or messager component adds a complexity to this bundle that is not necessary.

[...]

PR #1418, opened 2022-04-28
Decouple and Deprecate notification bundle from create snapshot code

eerison:
## Decouple and Deprecate notification bundle from **create snapshot** code

I am targeting this branch, because NoficationBundle code wasn't deprecated.

Issues #1415.

## Changelog

```markdown
### Added
- Added `CreateSnapshotService` to create snapshots
- Added `CreateSnapshotBySiteInterface`
- Added `CreateSnapshotByPageInterface`

### Deprecated
- Deprecated `async` mode in `CreateSnapshotsCommand`
- Deprecated `sonata.notification.backend` code into `SiteAdminController`
- Deprecated `sonata.notification.backend` code into `PageAdminController`
- Deprecated `sendMessage` and `$backend` property into the `CreateSnapshotAdminExtension`
```

## To do

**Create Snapshot**
- [x] Created test for create snapshot
- [x] Add BC in create snapshot command
- [x] Decouple NotificationBundle for create snapshot
- [x] Create snapshot by command line
- [x] see where `CreateSnapshotAdminExtension` is used, and remove notification bundle  from there (it was added here: https://github.com/sonata-project/SonataPageBundle/commit/23657026942fd6b171a30073265b20b3335d4358)
- [x] Create snapshot, page, and site by admin (it was leading some strange behaviour )
- [x] Edit doc (remove --site=all, because it's the default value)
- [x] Fix pipelines
- [x] validate the same create snapshot behaviour from 3.x branch
- [x] Add --site=all, just to provide a security execution like "doctrine:schema:update --force"


src/Command/BaseCommand.php
```diff
@@ -91,9 +92,11 @@ public function getErrorListener()
      * @param string $mode
      *
      * @return BackendInterface
+     * @deprecated This method will be removed on release 4.x
      */
     public function getNotificationBackend($mode)
     {
+        trigger_deprecation('sonata-project/page-bundle', '3.27.0', 'This "getNotificationBackend" will be removed on release 4.x.');
```
VincentLanglet: Deprecation are made this way
```
@trigger_error(sprintf(
            'The method "%s()" is deprecated since sonata-project/admin-bundle 4.7 and will be removed in 5.0.'
            .'  Use "%s::%s()" instead.',
            __METHOD__,
            RenderElementRuntime::class,
            __FUNCTION__
        ), \E_USER_DEPRECATED);
```
so far.

src/Command/BaseCommand.php
```diff
@@ -91,9 +92,11 @@ public function getErrorListener()
      * @param string $mode
      *
      * @return BackendInterface
+     * @deprecated This method will be removed on release 4.x
```
VincentLanglet: @deprecated since ..., will be removed in 4.0.


Also, we need a 
```
NEXT_MAJOR: Remove this method.
```

src/Command/CreateSnapshotsCommand.php
```diff
@@ -24,9 +25,18 @@
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
  * @final since sonata-project/page-bundle 3.26
+ * NEXT_MAJOR: Remove code related with mode option!
```
VincentLanglet: It's less important since I suppose you'll be the one removing the code on 4.x ; but when we're adding NEXT_MAJOR to remove code we try to add comment on every code to remove

```
NEXT_MAJOR: Remove this line
```
etc.

src/Command/CreateSnapshotsCommand.php
```diff
@@ -24,9 +25,18 @@
...
  */
 class CreateSnapshotsCommand extends BaseCommand
 {
+    private SnapshotManagerInterface $snapshotManager;
+
+    public function __construct(string $name = null, SnapshotManagerInterface $snapshotManager)
+    {
+        parent::__construct($name);
+        $this->snapshotManager = $snapshotManager;
+    }
```
VincentLanglet:
```suggestion
    public function __construct(SnapshotManagerInterface $snapshotManager)
    {
        parent::__construct();
        $this->snapshotManager = $snapshotManager;
    }
```
Seems ok to me.


But you'll have to support all the possible call
```
new Command(); // To deprecate
new Command($name); // To deprecate
new Command($snapshotManager);
```

Doing
```
public function __construct($snapshotManagerOrDeprecatedName = null)
    {
        if (snapshotManagerOrDeprecatedName instanceof SnapshotManagerInterface) {}
        elseif (is_string(snapshotManagerOrDeprecatedName) || === null) { // deprecation }
    }
```

src/Command/CreateSnapshotsCommand.php
```diff
@@ -65,6 +79,8 @@ public function execute(InputInterface $input, OutputInterface $output)
                     'mode' => $input->getOption('mode'),
                 ]);
 
+                $this->snapshotManager->save();
```
VincentLanglet: You have to handle (for BC) the case where this property is null.

src/Command/CreateSnapshotsCommand.php
eerison: Yeah probably, I'll remove this!

But I'm adding this just for not forget!

src/Command/CreateSnapshotsCommand.php
eerison: hmmmm I get it, because `CreateSnapshotsCommand` is not a final class :'( 

I'll add some support for this, thanks for the reminder ;)

src/Command/CreateSnapshotsCommand.php
eerison: But @VincentLanglet why should it be null? always I'll inject this **snapshotManager**, it's not from an outside code, it's in **PageBundle**

VincentLanglet: What if someone use it manually by doing
```
(new CreateSnapshotCommand())->execute(...)
```

eerison: Make sense 😅

eerison:
Note: I'll check all suggestion added in my PR, I didn't forgot this :D

eerison:
Ok, just a explanation about future remove for this piece of code

https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/CreateSnapshotsCommand.php#L69-L75

I'm trying to coverage the `CreateSnapshotsCommand.php` before make changes, and I can not coverage this [else condition](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/CreateSnapshotsCommand.php#L69), the reason: when we pass one value or more in `--site`, always it'll be an **array**, then never it'll be equals [all](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/CreateSnapshotsCommand.php#L56) (and enter into the "else")

maybe in the past it worked different or maybe it get a different behaviour int his [PR](https://github.com/sonata-project/SonataPageBundle/commit/359ec6460b5048531180206d3cbf9256aba5ec2c#diff-3782f26d1d23790f57b46418f648ad5f8d1537a923c44c1b15ef8c3f09f9aa1eR32)

just for your information :)

VincentLanglet:
> maybe in the past it worked different or maybe it get a different behaviour int his [PR](https://github.com/sonata-project/SonataPageBundle/commit/359ec6460b5048531180206d3cbf9256aba5ec2c#diff-3782f26d1d23790f57b46418f648ad5f8d1537a923c44c1b15ef8c3f09f9aa1eR32)

Indeed, prior to this PR, this was working.
Now the check would be `!\in_array('all', ....)`

eerison:
> > maybe in the past it worked different or maybe it get a different behaviour int his [PR](https://github.com/sonata-project/SonataPageBundle/commit/359ec6460b5048531180206d3cbf9256aba5ec2c#diff-3782f26d1d23790f57b46418f648ad5f8d1537a923c44c1b15ef8c3f09f9aa1eR32)
> 
> Indeed, prior to this PR, this was working. Now the check would be `!\in_array('all', ....)`

But this is not necessary, because if you pass `all`, it will return all sites and it'll create the snaps shots

it's handled here : https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/BaseCommand.php#L107

Edit 1: it'll work for `1 site`, `many sites` or `all` we don't need this else to call the same command, and it's working without this "else" ;)

VincentLanglet:
Indeed, you might remove the else then. (and the if check)

eerison:
@VincentLanglet what do you think I make this ticket in two parts?

one to **create snapshot** and other to **clean up the snapshot**

I guess the clean up code is not too big, but maybe do this in a separated pull request is better to review.

Edit1: @VincentLanglet I moved the checklist regarding `cleanup snapshot` for the issue #1415

src/Command/BaseCommand.php
eerison: @VincentLanglet I updated the deprecation code as you suggested, can you check again please :)

src/Command/BaseCommand.php
eerison: this too.

src/Command/CreateSnapshotsCommand.php
eerison: can you check again please

src/Command/CreateSnapshotsCommand.php
eerison: I changed this ;)

phpunit.xml.dist
```diff
@@ -32,6 +32,7 @@ It's auto-generated by sonata-project/dev-kit package.
 
     <php>
         <ini name="precision" value="8" />
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=8" />
+        <server name="KERNEL_CLASS" value="Sonata\PageBundle\Tests\App\AppKernel" />
```
VincentLanglet: This file is generated by dev kit and cannot be updated here.

You have to solve the deprecation (or use group legacy).
And what's the issue with the kernel class ?

src/Admin/Extension/CreateSnapshotAdminExtension.php
```diff
@@ -18,18 +18,22 @@
...
 
 /**
  * @final since sonata-project/page-bundle 3.26
  */
 class CreateSnapshotAdminExtension extends AbstractAdminExtension
 {
     /**
-     * @var BackendInterface
+     * @var BackendInterface|CreateSnapshotByPageInterface
+     *                                                     NEXT_MAJOR: rename this variable, for example: createSnapshotByPage
```
VincentLanglet:
```suggestion
     * NEXT_MAJOR: rename this variable to createSnapshotByPage and restrict type to CreateSnapshotByPageInterface
```

src/Admin/Extension/CreateSnapshotAdminExtension.php
```diff
@@ -18,18 +18,22 @@
...
+     *
+     * @deprecated since 3.27, and it will be removed in 4.0.
      */
     protected $backend;
 
-    public function __construct(BackendInterface $backend)
+    public function __construct($backend)
```
VincentLanglet, line 38: Add phpdoc and a next major comment to restrict to CreateSnapshotByPageInterface

src/Command/BaseCommand.php
```diff
@@ -102,7 +106,7 @@ public function getNotificationBackend($mode)
     }
 
     /**
-     * @return array
+     * @return array|Site[]
```
VincentLanglet:
```suggestion
     * @return Site[]
```

src/Command/BaseCommand.php
```diff
@@ -91,6 +92,9 @@ public function getErrorListener()
      * @param string $mode
      *
      * @return BackendInterface
+     *                          NEXT_MAJOR: Remove this method
```
VincentLanglet:
```suggestion
     * NEXT_MAJOR: Remove this method
```

src/Command/CreateSnapshotsCommand.php
```diff
@@ -27,52 +27,67 @@
  */
 class CreateSnapshotsCommand extends BaseCommand
 {
+    protected static $defaultName = 'sonata:page:create-snapshots';
+
+    /**
+     * @var ?string|CreateSnapshotBySiteInterface
```
VincentLanglet, line 33: NEXT_MAJOR restrict to CreateSnapshotBySiteInterface

src/Model/SnapshotManagerInterface.php
```diff
@@ -24,6 +24,7 @@
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
  * @method SnapshotPageProxyInterface createSnapshotPageProxy(TransformerInterface $transformer, SnapshotInterface $snapshot)
+ * @method get
```
VincentLanglet: ?

src/Service/CreateSnapshotService.php
```diff
@@ -0,0 +1,74 @@
...
+        $entityManager = $this->snapshotManager->getEntityManager();
+        $pages = $this->pageManager->findBy(['site' => $site->getId()]);
+
+        // start a transaction
+        $entityManager->beginTransaction();
+
+        foreach ($pages as $page) {
+            $this->createByPage($page);
+        }
+//
```
VincentLanglet:
```suggestion

```

tests/Admin/Extension/CreateSnapshotAdminExtensionTest.php
```diff
@@ -115,4 +116,26 @@ public function testPostRemoveOnBlock(): void
         $extension = new CreateSnapshotAdminExtension($backend);
         $extension->postRemove($admin, $block);
     }
+
+    /**
+     * @test
```
VincentLanglet: We're prefixing test, by test instead.

tests/Command/CreateSnapshotsCommandTest.php
```diff
@@ -0,0 +1,166 @@
...
+            ->willReturn([$siteMock]);
+
+        // Setup SymfonyKernel
+        $kernel = self::bootKernel();
+        $this->application = new Application($kernel);
+        $this->siteManagerMock = $siteManagerMock;
+    }
+
+    /**
+     * @test
```
VincentLanglet: Same

tests/Command/CreateSnapshotsCommandTest.php
```diff
@@ -0,0 +1,166 @@
...
+            '--mode' => 'async',
+        ]);
+
+        $output = $commandTester->getDisplay();
+
+        static::assertStringContainsString('done!', $output);
+    }
+
+    /**
+     * @test
```
VincentLanglet: Same

tests/Command/CreateSnapshotsCommandTest.php
```diff
@@ -0,0 +1,166 @@
...
+
+        $inputMock = $this->createMock(InputInterface::class);
+        $outputMock = $this->createMock(OutputInterface::class);
+
+        //Run code
+        $createSnapshotCommandMock->execute($inputMock, $outputMock);
+    }
+
+    /**
+     * @test
```
VincentLanglet: Same

tests/Command/CreateSnapshotsCommandTest.php
```diff
@@ -0,0 +1,166 @@
...
+     * @testdox it's using notificationBundle when mode option is equals "async"
+     * @testWith ["sync", 0, 1]
```
VincentLanglet: We're using @dataProvider instead

tests/Service/CreateSnapshotServiceTest.php
```diff
@@ -0,0 +1,65 @@
...
+use Sonata\PageBundle\Model\PageManagerInterface;
+use Sonata\PageBundle\Model\SiteInterface;
+use Sonata\PageBundle\Model\SnapshotInterface;
+use Sonata\PageBundle\Model\TransformerInterface;
+use Sonata\PageBundle\Service\CreateSnapshotService;
+
+final class CreateSnapshotServiceTest extends TestCase
+{
+    /**
+     * @test
```
VincentLanglet: Same

VincentLanglet (changes requested):

tests/Service/CreateSnapshotServiceTest.php
eerison: done

tests/Command/CreateSnapshotsCommandTest.php
eerison: done

tests/Command/CreateSnapshotsCommandTest.php
eerison: done

src/Service/CreateSnapshotService.php
eerison: done

src/Command/BaseCommand.php
eerison: done

src/Admin/Extension/CreateSnapshotAdminExtension.php
eerison: applied

phpunit.xml.dist
eerison: ~~well I was using this  `KERNEL_CLASS` to run the test from the IDE and from command line too~~

I need this to run test that depends of Kernel class

```
LogicException : You must set the KERNEL_CLASS environment variable to the fully-qualified class name of your Kernel in phpunit.xml / phpunit.xml.dist or override the "Sonata\PageBundle\Tests\Command\CreateSnapshotsCommandTest::createKernel()" or "Sonata\PageBundle\Tests\Command\CreateSnapshotsCommandTest::getKernelClass()" method.
```

eerison: do you have some example for this `or use group legacy`?

src/Command/BaseCommand.php
eerison: @VincentLanglet when I remove this space from begginner I get the lint broken :/ 

https://github.com/sonata-project/SonataPageBundle/runs/7130759831?check_suite_focus=true#step:5:27

phpunit.xml.dist
eerison: ~~well regarding this I'll open a new Pull request, appeared many deprecations :'(~~

src/Admin/Extension/CreateSnapshotAdminExtension.php
eerison: I added ;)

phpunit.xml.dist
VincentLanglet: `@group legacy` over some tests

src/Command/BaseCommand.php
VincentLanglet: I think it's because you need a blank line around the @return phpdoc.
The linter is thinking you're adding a multi-line comment here.

phpunit.xml.dist
eerison: Nice, the deprecation problem was solved :) 

what is missing is this `KERNEL_CLASS` thing. I need this to run one of my tests

https://github.com/sonata-project/SonataPageBundle/runs/7131719464?check_suite_focus=true#step:8:33

src/Command/BaseCommand.php
eerison: yes it's the problem, thanks :)

phpunit.xml.dist
eerison: well I guess I saw some tests using command line, I will check how was it build and refactor my tests ;)

VincentLanglet: @jordisala1991 Introduce multiple times tests with AppKernel.
the last one was https://github.com/sonata-project/SonataFormatterBundle/pull/672 if it can help you.

VincentLanglet: @eerison Seems like you can do it directly in the classes: https://github.com/sonata-project/SonataMediaBundle/blob/f121156da9ea2e08e1ed872c70e278bac989925a/tests/Functional/Admin/GalleryAdminTest.php#L90-L96

eerison: > @eerison Seems like you can do it directly in the classes: https://github.com/sonata-project/SonataMediaBundle/blob/f121156da9ea2e08e1ed872c70e278bac989925a/tests/Functional/Admin/GalleryAdminTest.php#L90-L96

ohhh, thanks for the info :)

eerison: Nice @VincentLanglet  it works :)

VincentLanglet (dismissed):

docs/reference/command_line.rst
```diff
@@ -33,7 +33,7 @@ You could also remove orphan pages with the ``--clean`` option.
 
 .. code-block:: bash
 
-    bin/console sonata:page:create-snapshots --site=all
+    bin/console sonata:page:create-snapshots
```
wbloszyk: IMO we should keep it. Command without argument(s) should not work.

wbloszyk (commented):
Im consider about tow things. 

1. Now you can change Page and you will see changes in site when you logged with `SUPER_ADMIN_ROLE` (preview before create new snapshots). This functionality must be keep. You don't change it but I want be sure.
2. What with performance and crash possibility? IMHO will be nice to move `BackendInteface` to `SonataPageBundle` with new `symfony/messanger` option only and keep `async`. Restore `async` can be done in `4.x` anyway.

docs/reference/command_line.rst
eerison: Hi @wbloszyk good point

> 1. Now you can change Page and you will see changes in site when you logged with `SUPER_ADMIN_ROLE` (preview before create new snapshots). This functionality must be keep. You don't change it but I want be sure.

~~I tested now, change the page and check the page before create a new publication, and no, it does not appear!
I just can see after create a new publication, But I remember this functionality, do you know where is the code that is checking this?~~

Edit 1: I didn't tested with SUPER_ADMIN, because the project that I'm testing do not have user, but I can test it on Monday :)

> 2. What with performance and crash possibility? IMHO will be nice to move `BackendInteface` to `SonataPageBundle` with new `symfony/messanger` option only and keep `async`. Restore `async` can be done in `4.x` anyway.

About `async` functionality, it was requested/agreed  in the begging of the issue: #1415 for @core23 

> IMO we should keep it. Command without argument(s) should not work.

what do you mean should not work?

should we not allow run commands without arguments? or this command won't work without this `--site=all`?

wbloszyk: > I tested now, change the page and check the page before create a new publication, and no, it does not appear!
I just can see after create a new publication, But I remember this functionality, do you know where is the code that is checking this?

I do not know. Probably preview is generated from `Page`. `Snapshot` is  `final cache` version. Can you debug with toolbar and `dump` function  starting from preview page?

> IMO we should keep it. Command without argument(s) should not work.
what do you mean should not work?
should we not allow run commands without arguments? or this command won't work without this --site=all?

`--site=` should be required. Simmilar like `--force` in 'doctrine:schema:update`. 

> About async functionality, it was requested/agreed in the begging of the issue: https://github.com/sonata-project/SonataPageBundle/issues/1415 for @core23

`Async` can be deprecated. We can marge this PR when preview will be fixed.

src/Controller/PageAdminController.php
```diff
@@ -265,4 +289,13 @@ private function setFormTheme(FormView $formView, $theme)
 
         $twig->getRuntime(FormRenderer::class)->setTheme($formView, $theme);
     }
+
+    /**
+     * NEXT_MAJOR: Check if it was added in SonataAdminBundle, if yes remove this method!
+     * @internal
+     */
+    protected function getAdmin(): AdminInterface
```
eerison: @VincentLanglet is it fine to add this method into https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Controller/CRUDController.php ? it can help to mock the admin.

VincentLanglet: I don't see a real benefit.

Adding code for tests is not a good pattern. And you can mock the adminFetcher
```
$adminFetcher = $this->container->get('sonata.admin.request.fetcher');
        \assert($adminFetcher instanceof AdminFetcherInterface);

        /** @var AdminInterface<T> $admin */
        $admin = $adminFetcher->get($request);
        $this->admin = $admin;
```

docs/reference/command_line.rst
eerison: just to update!

~~I can't reproduce this preview feature even using the 3.x branch! what I saw the different was~~

~~- `in release `3.x` using the env=dev`  :  when I update the block content, generate a new publication~~
~~- `in release `3.x` using the env=prod`  :  when I update the block, any publication is generated, But I can't see the changes when I click in the preview~~

~~in my branch, the first behaviour doesn't happen, when I update the block content any publication is created, But I looked into the code searching by `dev`, `prod`, listeners and I didn't find anything :(~~

Edit 1: OK, I got the same behaviour of `3.x` branch, But in any branch I can't make this `preview feature` works, I guess it's missing something, locally to test, can you reproduce this locally @wbloszyk ? or just in `prod`

Edit 2: it's my locally user (I'm using `memory users`)
Screenshot 2022-07-05 at 19 48 03

src/Controller/PageAdminController.php
eerison: @VincentLanglet , But I can not mock `$this->admin` using your example! it does not work for me!

Note: if you look into this test: https://github.com/sonata-project/SonataPageBundle/pull/1418/files#diff-2486e8a2bb13c5d093018e6058f6d13246f449650d15ea1bc41d2f3b7136039f, I'm extending from `TestCase`

**Edit 1:** look I'm trying to mock this line: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Controller/PageAdminController.php#L52

**Edit 2:** What I have found was this: [configureAdmin](https://github.com/sonata-project/SonataAdminBundle/blob/4.x/tests/Controller/CRUDControllerTest.php#L278) or create a reflection, But I guess it's too much to just mock one method/property

docs/reference/command_line.rst
eerison: ✅ Added the `--site=all` as required
✔️ I tested the preview page, and it has exactly the same behaviour then `3.x` branch, But I can't confirm that you will see the changes before de publication, because I can't reproduce this in any branch, even in other project that I'm using.

src/Controller/PageAdminController.php
VincentLanglet: You can mock the adminFetcher return value, like we did in SonataAdmin https://github.com/sonata-project/SonataAdminBundle/blob/4.x/tests/Controller/CRUDControllerTest.php#L154-L157

eerison: > You can mock the adminFetcher return value, like we did in SonataAdmin https://github.com/sonata-project/SonataAdminBundle/blob/4.x/tests/Controller/CRUDControllerTest.php#L154-L157

I can create the mock like this, but the question is, how can I put this mock in `$this->admin` in PageAdminController


Maybe it's really clear and I can't see how to do this 😄

eerison: Ok, tomorrow I'll try again follow your example :)

jordisala1991 (dismissed):

src/Controller/PageAdminController.php
eerison: Hey @VincentLanglet 

I tried to use the setUp that you mentioned, But I'm getting this error

```
RuntimeException : There is no `_sonata_admin` defined for the controller `Sonata\PageBundle\Controller\PageAdminController` and the current route ``.
```

any idea why?

eerison: I removed the `getAdmin` method, But I used Reflection, is it fine for you?

https://github.com/sonata-project/SonataPageBundle/pull/1418/files#diff-2486e8a2bb13c5d093018e6058f6d13246f449650d15ea1bc41d2f3b7136039fR95-R102

VincentLanglet:
@eerison I added a commit to show you how this could be done, if you want to take a look.
I'm pretty sure I mocked too many things though.

eerison:
> @eerison I added a commit to show you how this could be done, if you want to take a look. I'm pretty sure I mocked too many things though.

Ok I'll check ;) 

Edit 1: Hey @VincentLanglet I checked your changes and almost the whole code is necessary, I just removed 1 mock and added some `NEXT_MAJOR` comments. and it's covering  the changes that I made in `batchActionSnapshot`.

VincentLanglet (approved):

VincentLanglet:
> > @eerison I added a commit to show you how this could be done, if you want to take a look. I'm pretty sure I mocked too many things though.
> 
> Ok I'll check ;)
> 
> Edit 1: Hey @VincentLanglet I checked your changes and almost the whole code is necessary, I just removed 1 mock and added some `NEXT_MAJOR` comments. and it's covering the changes that I made in `batchActionSnapshot`.

Then, Am I wrong or is it ready to be merged ?

eerison:
> > > @eerison I added a commit to show you how this could be done, if you want to take a look. I'm pretty sure I mocked too many things though.
> > 
> > 
> > Ok I'll check ;)
> > Edit 1: Hey @VincentLanglet I checked your changes and almost the whole code is necessary, I just removed 1 mock and added some `NEXT_MAJOR` comments. and it's covering the changes that I made in `batchActionSnapshot`.
> 
> Then, Am I wrong or is it ready to be merged ?

Yes it's ready to merge ;)

VincentLanglet:
Ok, I'll wait for @jordisala1991 review

jordisala1991 (approved):

jordisala1991:
Thanks @eerison , really good work!

jordisala1991:
We missed something here, this other command also uses notificationBundle to work: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/CleanupSnapshotsCommand.php

eerison:
> We missed something here, this other command also uses notificationBundle to work: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/CleanupSnapshotsCommand.php

Hey @jordisala1991 no I didn't missed this, I will create a PR for this cleanup too, I didn't do this here, because it will be too big

Issue #1420, opened 2022-05-10
Remove dependency of ContainerAwareCommand


eerison:
## Remove dependency of ContainerAwareCommand

As `ContainerAwareCommand` is deprecated since [Symfony 4.2](https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-containerawarecommand), we should remove code that extends from that Class! like [`src/Command/BaseCommand.php`](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Command/BaseCommand.php)

eerison:
It will be solved after this PR #1460 be merged

eerison:
It was solved on #1460

PR #1422, opened 2022-06-04
Remove deprecated usage of ContainerAwareCommand

ValentinMerlet:
## Subject

I am targeting this branch, because `4.x` will be compatible with Symfony 5.

Closes https://github.com/sonata-project/SonataPageBundle/issues/1420

## Changelog

```markdown
### Removed

- Removed usage of deprecated `ContainerAwareCommand` class
```

### DIFF
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -15,7 +15,7 @@ namespace Sonata\PageBundle\Command;
 
 use Sonata\Doctrine\Model\ManagerInterface;
 use Sonata\NotificationBundle\Backend\BackendInterface;
-use Sonata\PageBundle\CmsManager\CmsPageManager;
+use Sonata\PageBundle\CmsManager\CmsManagerInterface;
 use Sonata\PageBundle\Listener\ExceptionListener;
 use Sonata\PageBundle\Model\PageManagerInterface;
 use Sonata\PageBundle\Model\SiteManagerInterface;
@@ -42,7 +42,7 @@ abstract class BaseCommand extends Command
     /** @var ManagerInterface */
     protected $blockManager;
 
-    /** @var CmsPageManager */
+    /** @var CmsManagerInterface */
     protected $cmsPageManager;
 
     /** @var ExceptionListener */
@@ -59,7 +59,7 @@ abstract class BaseCommand extends Command
         PageManagerInterface $pageManager,
         SnapshotManagerInterface $snapshotManager,
         ManagerInterface $blockManager,
-        CmsPageManager $cmsPageManager,
+        CmsManagerInterface $cmsPageManager,
         ExceptionListener $exceptionListener,
         BackendInterface $backend,
         BackendInterface $backendRuntime
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Command;
 
+use Psr\Container\ContainerInterface;
 use Sonata\Doctrine\Model\ManagerInterface;
 use Sonata\NotificationBundle\Backend\BackendInterface;
 use Sonata\PageBundle\CmsManager\CmsManagerInterface;
@@ -22,58 +23,37 @@ use Sonata\PageBundle\Model\SiteManagerInterface;
 use Sonata\PageBundle\Model\SnapshotManagerInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
 
 /**
  * BaseCommand.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-abstract class BaseCommand extends Command
+abstract class BaseCommand extends Command implements ServiceSubscriberInterface
 {
-    /** @var SiteManagerInterface */
-    protected $siteManager;
+    /** @var ContainerInterface */
+    protected $locator;
 
-    /** @var PageManagerInterface */
-    protected $pageManager;
-
-    /** @var SnapshotManagerInterface */
-    protected $snapshotManager;
-
-    /** @var ManagerInterface */
-    protected $blockManager;
-
-    /** @var CmsManagerInterface */
-    protected $cmsPageManager;
-
-    /** @var ExceptionListener */
-    protected $exceptionListener;
-
-    /** @var BackendInterface */
-    protected $backend;
-
-    /** @var BackendInterface */
-    protected $backendRuntime;
-
-    public function __construct(
-        SiteManagerInterface $siteManager,
-        PageManagerInterface $pageManager,
-        SnapshotManagerInterface $snapshotManager,
-        ManagerInterface $blockManager,
-        CmsManagerInterface $cmsPageManager,
-        ExceptionListener $exceptionListener,
-        BackendInterface $backend,
-        BackendInterface $backendRuntime
-    ) {
+    public function __construct(ContainerInterface $locator)
+    {
         parent::__construct();
 
-        $this->siteManager = $siteManager;
-        $this->pageManager = $pageManager;
-        $this->snapshotManager = $snapshotManager;
-        $this->blockManager = $blockManager;
-        $this->cmsPageManager = $cmsPageManager;
-        $this->exceptionListener = $exceptionListener;
-        $this->backend = $backend;
-        $this->backendRuntime = $backendRuntime;
+        $this->locator = $locator;
+    }
+
+    public static function getSubscribedServices(): array
+    {
+        return [
+            'sonata.page.manager.site' => SiteManagerInterface::class,
+            'sonata.page.manager.page' => PageManagerInterface::class,
+            'sonata.page.manager.snapshot' => SnapshotManagerInterface::class,
+            'sonata.page.manager.block' => ManagerInterface::class,
+            'sonata.page.cms.page' => CmsManagerInterface::class,
+            'sonata.page.kernel.exception_listener' => ExceptionListener::class,
+            'sonata.notification.backend' => BackendInterface::class,
+            'sonata.notification.backend.runtime' => BackendInterface::class,
+        ];
     }
 
     /**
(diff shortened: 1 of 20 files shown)


src/Command/RenderBlockCommand.php
```diff
@@ -78,13 +125,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $output->writeln("\n<info>Response Output</info>");
 
         // fake request
-        $request = new Request();
-        $this->getContainer()->enterScope('request');
-        $this->getContainer()->set('request', $request, 'request');
-
-        $output->writeln($this->getContainer()->get('sonata.block.renderer')->render($context));
-
-        $this->getContainer()->leaveScope('request');
```
ValentinMerlet, line 87: `enterScope`/`leaveScope` functions do not exist anymore, I don't know how to reproduce this behaviour and I'm not sure what the goal was.

src/Command/BaseCommand.php
```diff
@@ -16,75 +16,112 @@
...
+        $this->cmsPageManager = $cmsPageManager;
+        $this->exceptionListener = $exceptionListener;
+        $this->backend = $backend;
+        $this->backendRuntime = $backendRuntime;
+    }
+
     /**
      * @return SiteManagerInterface
      */
     public function getSiteManager()
```
VincentLanglet: I don't see a reason to keep those public function when we can access by the property ?

src/Command/RenderBlockCommand.php
```diff
@@ -41,21 +87,22 @@ public function configure(): void
...
-        if (!$manager instanceof CmsManagerInterface) {
-            throw new \RuntimeException('The service does not implement the CmsManagerInterface');
+        if (!\in_array($manager, ['sonata.page.cms.snapshot', 'sonata.page.cms.page'], true)) {
+            throw new \RuntimeException(
+                'Available managers are "sonata.page.cms.snapshot" and "sonata.page.cms.page"'
+            );
         }
 
-        $page = $manager->getPageById($input->getArgument('page_id'));
+        $managerService = 'sonata.page.cms.snapshot' === $manager ? $this->cmsSnapshotManager : $this->cmsPageManager;
```
VincentLanglet, line 100: What if someone created his own CmsManagerInterface ?
It cannot use it anymore. We should find another solution.

Maybe injecting a service locator with all the CmsManagerInterface instead of directly injecting the CmsManagerInterface.

src/Command/RenderBlockCommand.php
VincentLanglet: I have no idea ^^

VincentLanglet (commented):
There are a lot of dependency in those commands ; any idea how to reduce this @jordisala1991  @sonata-project/contributors ?

VincentLanglet:
https://github.com/sonata-project/SonataPageBundle/runs/6769778295?check_suite_focus=true is failing

src/Command/RenderBlockCommand.php
jordisala1991: This is for when request was an object on the container, We should ensure no one tries to access request directly from the container and use requestStack instead.

src/Command/RenderBlockCommand.php
ValentinMerlet: You're right, I did not think about it. I'm trying to inject a service locator containing all the services implementing `CmsManagerInterface` as you suggest and I let you know.

mareckigit:
@ValentinMerlet Consider to implement Service Subscriber in the BaseCommand and to implement [Lazy Loading](https://symfony.com/doc/5.4/console/commands_as_services.html#lazy-loading) in all commands. I'm working with my own [fork](https://github.com/treehousedevelopers/SonataPageBundle) that contains your changes and [Haniki:sf4_bump](https://github.com/Haniki/SonataPageBundle/tree/sf4_bump). Maybe some changes will be useful for you.

src/Command/RenderBlockCommand.php
ValentinMerlet: Hello @VincentLanglet, I tried, as you suggest, to add a service locator for cms manager services but I'm struggling while trying to add services as argument of the service locator in a compiler pass. See the following commit: https://github.com/sonata-project/SonataPageBundle/pull/1422/commits/59edbd5b067a2694565f2d14d7ce64098e8f83ca.

I have the following error:
error_during_compiler_pass_execution. I guess using reflection in a compiler pass is not a good idea, any other idea how to add services implementing [this interface](https://github.com/sonata-project/SonataPageBundle/blob/f56de820e5d14c008238dbf7bcb077a4c963f717/src/CmsManager/CmsManagerInterface.php), maybe using tags? But it should be documented somewhere. What do you think?

VincentLanglet: I know @franmomu already used service locator multiple times, maybe he can help.

If you look at https://github.com/sonata-project/SonataAdminBundle/pull/7056/files, reflection was used so this might not be the origin of the problem.

I think the purpose of the compilerPass is to dynamically call `$definition->replaceArgument` in order to set the dependency. I don't think this is what you're doing here.

ValentinMerlet: > I think the purpose of the compilerPass is to dynamically call `$definition->replaceArgument` in order to set the dependency. I don't think this is what you're doing here.

I think I should use `addArgument()` as I did and not `replaceArgument()` because I guess we want to support both "natives" services that implements the [CmsManagerInterface](https://github.com/sonata-project/SonataPageBundle/blob/f56de820e5d14c008238dbf7bcb077a4c963f717/src/CmsManager/CmsManagerInterface.php) and services that implements the same interface but defined by the users of SonataPageBundle.

I wonder if the fact that I don't loop on services with a specific tag but onto [all the services](https://github.com/sonata-project/SonataPageBundle/commit/59edbd5b067a2694565f2d14d7ce64098e8f83ca#diff-91f89bb87705458d4026c45c68356cda2fd39acdb1b7cd8d73a6d94bc94b6038R28) can cause performance issues :thinking: 

**Edit**: I figured out that my issue with the fact that `Psr\SimpleCache\CacheInterface` does not exist comes from that the dependency is not installed in the vendor but I can't figure out why. Do you have an idea?

ValentinMerlet:
> @ValentinMerlet Consider to implement Service Subscriber in the BaseCommand and to implement [Lazy Loading](https://symfony.com/doc/5.4/console/commands_as_services.html#lazy-loading) in all commands. I'm working with my own [fork](https://github.com/treehousedevelopers/SonataPageBundle) that contains your changes and [Haniki:sf4_bump](https://github.com/Haniki/SonataPageBundle/tree/sf4_bump). Maybe some changes will be useful for you.

Done :)

src/Command/RenderBlockCommand.php
ValentinMerlet: [Scopes are deprecated since Symfony 2.8](https://symfony.com/blog/new-in-symfony-2-8-deprecating-scopes-and-introducing-shared-services), is there anything to add here to keep the same behaviour as before?

VincentLanglet:
WDYT about the solution @jordisala1991 ? Is it something "classic" to implement service subscriber ? 
I was ok with the injection in the constructor IMHO

treehousedevelopers:
I have implemented Service Subscriber in my fork https://github.com/treehousedevelopers/SonataPageBundle/commit/066f3ea1dfecb401a9fb6123a9fc7c11ee2f2bf9

mareckigit:
I have implemented Service Subscriber in my fork treehousedevelopers@066f3ea

src/Command/BaseCommand.php
```diff
@@ -13,78 +13,47 @@
...
-        return $this->getContainer()->get('sonata.page.manager.site');
-    }
+    /** @var ContainerInterface */
+    protected $locator;
 
-    /**
-     * @return PageManagerInterface
-     */
-    public function getPageManager()
+    public function __construct(ContainerInterface $locator)
```
eerison, line 38: we need to avoid to inject the whole ContainerInterface, as far as I know it will be removed for the future symfony releases, and I guess to this work the service need to be public!

mareckigit: This is not whole container. The interface is the same, but it is small container that includes only declared services and provides lazy. Symfony do the same job in controllers, You can read more about it in doc https://symfony.com/doc/5.4/service_container/service_subscribers_locators.html

eerison: hmmm interesting, I didn't know about it!
well I need to research a bit more about this, But I would inject all dependencies explicit in my service!

But I do not have any argument to say that it's right or wrong, I just like to look at the class and see what is necessary for this work!

mareckigit: I have suggested to use Service Subscriber because there is set of services here and commands use only some from the set. The benefit is that only used services are initialized not all from the set. Commands should be also lazy and private services.

eerison: well but as it's supported by symfony 5.x and 6.x, I'm fine with that, in the future if someone could provide a better solution we can do :)

VincentLanglet: I don't know this a lot too, so I pinged @jordisala1991 to give his opinion on this strategy.

src/Command/RenderBlockCommand.php
franmomu: I don't know this bundle well, but reading the comments we could maybe autoconfigure services that implement `CmsManagerInterface` with a tag, that would allow to loop them instead of all services.

src/Command/RenderBlockCommand.php
jordisala1991: I would say not

VincentLanglet:
WDYT here @jordisala1991, should we inject services in the constructor or implements ServiceSubscriberInterface ?

jordisala1991:
IMO inject deps on construct, and avoid a base class.

VincentLanglet:
Indeed the baseCommand is not useful at all, and one of the method `getNotificationBackend` is now deprecated

Without this base command, each command will have his own dependencies in their constructor.

eerison:
> 

I guess we won't need this baseCommand too, after refactor the cleanup command I can see what is really necessary or not!

eerison:
I checked the BaseCommand and as we thought, we won't need this class anymore, In my Pull request: https://github.com/sonata-project/SonataPageBundle/pull/1434/files#diff-a351a99354f8a5971094d057717392bb08fcfa1b8fa6310d2b8d54d219068552, I'm deprecating the baseCommand, then I guess this PR is not necessary anymore.

jordisala1991:
This is already fixed on 4.x

PR #1427, opened 2022-07-04
Make Cache dependency optional

Hanmac:
## Changelog

```markdown
### Changed
- Make Cache dependency optional
```


src/Admin/BaseBlockAdmin.php
```diff
@@ -133,14 +134,16 @@ public function postPersist($object)
...
 
             @trigger_error(
-                'The '.__METHOD__.'() method is deprecated since sonata-project/block-bundle 3.12.0 and will be removed in version 4.0.',
+                'The ' . __METHOD__ . '() method is deprecated since sonata-project/block-bundle 3.12.0 and will be removed in version 4.0.',
                 \E_USER_DEPRECATED
             );
         }
 
-        $service = $this->blockManager->get($object);
+        if ($this->cacheManager) {
```
VincentLanglet: Prefer a check `!== null`.

Also, the phpdoc of the property need to be updated to `|null`.

src/Admin/BaseBlockAdmin.php
```diff
@@ -133,14 +134,16 @@ public function postPersist($object)
             $block->postPersist($object);
 
             @trigger_error(
-                'The '.__METHOD__.'() method is deprecated since sonata-project/block-bundle 3.12.0 and will be removed in version 4.0.',
+                'The ' . __METHOD__ . '() method is deprecated since sonata-project/block-bundle 3.12.0 and will be removed in version 4.0.',
```
VincentLanglet: Revert this change

VincentLanglet (changes requested):

src/Admin/PageAdmin.php
```diff
@@ -58,7 +58,7 @@ class PageAdmin extends AbstractAdmin
     protected $siteManager;
 
     /**
-     * @var CacheManagerInterface
+     * @var CacheManagerInterface|null
```
eerison, line 61: why it's possible to be null for now?

composer.json
```diff
@@ -27,8 +27,6 @@
         "doctrine/persistence": "^1.3.4 || ^2.0",
         "sonata-project/admin-bundle": "^3.91.1",
         "sonata-project/block-bundle": "^3.20",
-        "sonata-project/cache": "^2.0",
```
eerison, line 30: ohh is it possible to remove cache and cache-bundle in `3.x` without any BC 😮

eerison:
@Hanmac if it's not ready to review put as `draft` ;)

Hanmac:
> @Hanmac if it's not ready to review put as `draft` ;)

it should have been ready i think, just the lint stuff i couldn't get working in my IDE

src/Admin/PageAdmin.php
Hanmac: the plan is to remove it in next major version

VincentLanglet (approved):

VincentLanglet (changes requested):
Although this seems ok, the cache is used as soon as it's installed.
But since SonataCacheBundle is installed by blockBundle, it's always used.

So we need to add a config option to disable the cache, instead of relying on `isset($bundles['SonataCacheBundle'])`. And we'll be able to add a deprecation is the cache is not disabled.

src/Admin/PageAdmin.php
eerison: then this bundle doesn't not require cache to works right?

eerison:
> Although this seems ok, the cache is used as soon as it's installed. But since SonataCacheBundle is installed by blockBundle, it's always used.
> 
> So we need to add a config option to disable the cache, instead of relying on `isset($bundles['SonataCacheBundle'])`. And we'll be able to add a deprecation is the cache is not disabled.



hmmm then in case this option is "true", we will throw a deprecation message and remove cache completely in the next major?

src/Admin/PageAdmin.php
Hanmac: not anymore, that's why i moved it from require to require-dev for now

VincentLanglet:
> hmmm then in case this option is "true", we will throw a deprecation message and remove cache completely in the next major?

Depends if the option is enableCache or disableCache, but yes that's the idea.

`cache: true` should be the default value to not change any behavior, but trigger a deprecation.
`cache: false` will stop using the cache and send no deprecation.

4.x we remove the cache feature.

Hanmac:
> > hmmm then in case this option is "true", we will throw a deprecation message and remove cache completely in the next major?
> 
> Depends if the option is enableCache or disableCache, but yes that's the idea.
> 
> `cache: true` should be the default value to not change any behavior, but trigger a deprecation. `cache: false` will stop using the cache and send no deprecation.
> 
> 4.x we remove the cache feature.

should cache still be removed from the required or not yet?

eerison:
> > hmmm then in case this option is "true", we will throw a deprecation message and remove cache completely in the next major?
> 
> Depends if the option is enableCache or disableCache, but yes that's the idea.
> 
> `cache: true` should be the default value to not change any behavior, but trigger a deprecation. `cache: false` will stop using the cache and send no deprecation.
> 
> 4.x we remove the cache feature.

Ohhh nice, then I guess we do not need to Touch in this composer depencies, only in 4.x

VincentLanglet:
> should cache still be removed from the required or not yet?

I just took a look at blockBundle and
https://github.com/sonata-project/SonataBlockBundle/blob/3.x/composer.json#L27
seems like cache was required but not CacheBundle.

The isset strategy could work then. WDYT @jordisala1991

VincentLanglet (dismissed):

VincentLanglet:
But if we move the dependency from require to require-dev, it also need to be in the conflict section to restrict the range of the versions

composer.json
```diff
@@ -68,6 +66,8 @@
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.16",
         "psalm/plugin-symfony": "^3.0",
+        "sonata-project/cache": "^2.0",
+        "sonata-project/cache-bundle": "^3.3",
```
jordisala1991, line 70: We should add this same range on the conflict section.

jordisala1991 (commented):
Should we mark something to removal on 4.x? We can make on another PR if needed.

The idea of moving this to optional dependencies is to remove it on NEXT_MAJOR.

composer.json
Hanmac: added the lower versions to conflict

VincentLanglet (changes requested):
Tests need to be fixed

VincentLanglet (approved):

VincentLanglet (changes requested):
You'll need to add `@group legacy` above the test 
```
Issue1134Test::setUp from Sonata\PageBundle\Tests\Functional\Ticket
```
to avoid an error with the deprecation.

And we should also add a test with cache:false

VincentLanglet (commented):
The goal is to solve
```
Remaining self deprecation notices (1)
  1x: Using SonataCacheBundle is deprecated since sonata-project/page-bundle 3.27 and will be removed in 4.x
    1x in Issue1134Test::setUp from Sonata\PageBundle\Tests\Functional\Ticket
```

Maybe the group legacy should be added to the setUp of ResetableDBWebTestTest instead.

But you can also try to modify the test config
https://github.com/sonata-project/SonataPageBundle/blob/fed26c06856de8d6b1161d0f23001db7b8377e2b/tests/App/config/sonata.yaml
To add cache: false, instead.

VincentLanglet:
You don't need anymore the `group legacy` since you use `cache: false`

VincentLanglet (approved):

jordisala1991 (approved):

VincentLanglet:
Thanks

PR #1429, opened 2022-07-07
remove symfony/debug dependency

Hanmac:
## Subject

This should fix this deprecation:
> The "Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent::getException()" method is deprecated

I am targeting this branch, because it may be BC even if it is deprecation removal.

## Changelog

```markdown
### Removed
- Remove `symfony/debug` dependency as it is deprecated.
```

### DIFF
--- a/src/Listener/ExceptionListener.php
+++ b/src/Listener/ExceptionListener.php
@@ -20,7 +20,7 @@ use Sonata\PageBundle\Exception\InternalErrorException;
 use Sonata\PageBundle\Page\PageServiceManagerInterface;
 use Sonata\PageBundle\Site\SiteSelectorInterface;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
+use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Templating\EngineInterface;
@@ -154,9 +154,9 @@ class ExceptionListener
      *
      * @throws \Exception
      */
-    public function onKernelException(GetResponseForExceptionEvent $event)
+    public function onKernelException(ExceptionEvent $event): void
     {
-        if ($event->getException() instanceof NotFoundHttpException && $this->cmsManagerSelector->isEditor()) {
+        if ($event->getThrowable() instanceof NotFoundHttpException && $this->cmsManagerSelector->isEditor()) {
             $pathInfo = $event->getRequest()->getPathInfo();
 
             // can only create a CMS page, so the '_route' must be null
@@ -176,7 +176,7 @@ class ExceptionListener
             }
         }
 
-        if ($event->getException() instanceof InternalErrorException) {
+        if ($event->getThrowable() instanceof InternalErrorException) {
             $this->handleInternalError($event);
         } else {
             $this->handleNativeError($event);
@@ -186,18 +186,18 @@ class ExceptionListener
     /**
      * Handles an internal error.
      */
-    private function handleInternalError(GetResponseForExceptionEvent $event)
+    private function handleInternalError(ExceptionEvent $event): void
     {
         if (false === $this->debug) {
-            $this->logger->error($event->getException()->getMessage(), [
-                'exception' => $event->getException(),
+            $this->logger->error($event->getThrowable()->getMessage(), [
+                'exception' => $event->getThrowable(),
             ]);
 
             return;
         }
 
         $content = $this->templating->render('@SonataPage/internal_error.html.twig', [
-            'exception' => $event->getException(),
+            'exception' => $event->getThrowable(),
         ]);
 
         $event->setResponse(new Response($content, 500));
@@ -206,7 +206,7 @@ class ExceptionListener
     /**
      * Handles a native error.
      */
-    private function handleNativeError(GetResponseForExceptionEvent $event)
+    private function handleNativeError(ExceptionEvent $event): void
     {
         if (true === $this->debug) {
             return;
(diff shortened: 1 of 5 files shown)


src/Listener/ExceptionListener.php
```diff
@@ -154,9 +154,9 @@ public function getErrorCodePage($statusCode)
      *
      * @throws \Exception
      */
-    public function onKernelException(GetResponseForExceptionEvent $event): void
```
eerison: @Hanmac you should provide some support to BC or should he just add a comment in UPGRADE.md @VincentLanglet ?

Note: the same for the others methods that you changed the attribute's type

Hanmac: i made that MR extra for 4.x because it is a BC break
and it is done to make 4.x work with newer symfony

or do you mean i should make another PR for 3.x?
The ExceptionListener should be internal stuff imo?

eerison: I don't know, if we should provide BC support for 3.x or just add a comment in Upgrade.md

as this class is not final, if someone extend this class and override this method, it'll break in `4.x`, I guess you need to provide some support from 3.x and add deprecation messages for whom extends, have noticed that it'll be changed, for example

```php
/**
 * NEXT_MAJOR: Restrict the $event param with  "ExceptionEvent" type
 *
 * @param GetResponseForExceptionEvent|ExceptionEvent $event
 */
public function onKernelException($event): void
{
    //NEXT_MAJOR: Remove this "if"
    if ($event instanceof GetResponseForExceptionEvent) {
        //trigger a deprecation message
    }
}
```

and you open this for 3.x branch and remove the old code in `4.x`

Hanmac: in symfony 4.4 ExceptionEvent extends from GetResponseForExceptionEvent

so even if the user didn't change the code it should work with symfony 4.4

technically i might even so this same PR in the 3.x branch because it should be BC (for now)

eerison: well if I understand correct, we should provide something to support a smooth upgrade from `3.x` to `4.x`
https://symfony.com/doc/current/contributing/code/bc.html

then you need to do something like the example that I added above! and open the PR to `3.x` and not to `4.x`

Hanmac: it might be a warning in PHP 7.3 but works without problems in PHP 7.4

Example:
https://onlinephp.io/c/d9e66

eerison: in this case, I guess we should required 7.4 in `3.x` branch @VincentLanglet ?

Edit 1: I guess do not have problem upgrade php minor version into the current branch (3.x), I saw it's happening  in others bundles, But I guess you need to remove 7.3 tests from the pipeline :/

Hanmac: i also could do a small PR for 3.x too, but that is for @VincentLanglet do decide what to do

eerison: > I don't know, if we should provide BC support for 3.x or just add a comment in Upgrade.md
> 
> as this class is not final, if someone extend this class and override this method, it'll break in `4.x`, I guess you need to provide some support from 3.x and add deprecation messages for whom extends, have noticed that it'll be changed, for example
> 
> ```
> /**
>  * NEXT_MAJOR: Restrict the $event param with  "ExceptionEvent" type
>  *
>  * @param GetResponseForExceptionEvent|ExceptionEvent $event
>  */
> public function onKernelException($event): void
> {
>     //NEXT_MAJOR: Remove this "if"
>     if ($event instanceof GetResponseForExceptionEvent) {
>         //trigger a deprecation message
>     }
> }
> ```
> 
> and you open this for 3.x branch and remove the old code in `4.x`

it won't work too, it'll leads warnings 😄

Hanmac: > > I don't know, if we should provide BC support for 3.x or just add a comment in Upgrade.md
> > as this class is not final, if someone extend this class and override this method, it'll break in `4.x`, I guess you need to provide some support from 3.x and add deprecation messages for whom extends, have noticed that it'll be changed, for example
> > ```
> > /**
> >  * NEXT_MAJOR: Restrict the $event param with  "ExceptionEvent" type
> >  *
> >  * @param GetResponseForExceptionEvent|ExceptionEvent $event
> >  */
> > public function onKernelException($event): void
> > {
> >     //NEXT_MAJOR: Remove this "if"
> >     if ($event instanceof GetResponseForExceptionEvent) {
> >         //trigger a deprecation message
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and you open this for 3.x branch and remove the old code in `4.x`
> 
> it won't work too, it'll leads warnings 😄

thats because the new ExceptionEvent extends from GetResponseForExceptionEvent

VincentLanglet: If I understand correctly, if we drop 7.3 support this is BC. isn't it ?
Let's drop 7.3 then.

Can you bump the php version to 7.4 ?

VincentLanglet: I made the dev-kit PR https://github.com/sonata-project/dev-kit/pull/2104

VincentLanglet (dismissed):

eerison:
you need to remove the `7.3` pipelines

Hanmac:
> you need to remove the `7.3` pipelines

i could do that, but isn't that the work of the dev-kit?

VincentLanglet (approved):

VincentLanglet:
@eerison @Hanmac, are you able to try 3.x-dev on your project before I create a new release ?

eerison:
> @eerison @Hanmac, are you able to try 3.x-dev on your project before I create a new release ?

@VincentLanglet maybe we should do this after I finish the cleanup snapshots, what do you think?

Edit 1: Or do you prefer create a new release now and other after I finish the cleanup snapshots?

Hanmac:
i don't have a project that uses page bundle, i was just there for help ;P

VincentLanglet:
> > @eerison @Hanmac, are you able to try 3.x-dev on your project before I create a new release ?
> 
> @VincentLanglet maybe we should do this after I finish the cleanup snapshots, what do you think?
> 
> Edit 1: Or do you prefer create a new release now and other after I finish the cleanup snapshots?

Yes of course, I can wait if there is more to change in 3.x.

jordisala1991 (approved):

jordisala1991:
Thanks @Hanmac

PR #1433, opened 2022-07-07
Solve most NEXT_MAJOR comments

jordisala1991:
## Subject

I am targeting this branch, because this is for NEXT MAJOR.

PR #1434, opened 2022-07-08
Feature/1415/Decouple notification-bundle from cleanup snapshot

eerison:
## Decouple notification-bundle from cleanup snapshot code

I am targeting this branch, because the cleanup snapshot uses notification bundle and we need to decouple the code to remove on the next major

Issue #1415

## Changelog

```markdown
### Deprecated
- Deprecated `CleanupSnapshotConsumer` class
- Deprecated `CleanupSnapshotsConsumer` class
- Deprecated `--mode` option for `sonata:page:cleanup-snapshots`
- Deprecated `--base-console` option for `sonata:page:cleanup-snapshots`
- Deprecated `--site=all` option  for `sonata:page:cleanup-snapshots`
- Deprecated `--base-console` option  for `sonata:page:create-snapshots`
- Deprecated `--site=all` option  for `sonata:page:create-snapshots`
- Deprecated `--site=all` option  for `sonata:page:update-core-routes` 
- Deprecated `BaseCommand::class`
```

## To do

- [x] Cover the cleanup command before use the new code.
- [x] Prepare the cleanup command to use the new code and deprecate the old one.
- [x] Created test for clean snapshot
- [x] Add BC in clean snapshot command
- [x] Decouple NotificationBundle for clean snapshot
- [x] Apply the deprecate message only in [BaseCommand](https://github.com/sonata-project/SonataPageBundle/pull/1434#discussion_r916534159) `__construct`
    - [x] Not use `getSites`
    - [x] Not use `getNotificationBackend`
    - [x] Not use `getErrorListener`
    - [x] Not use `getDecoratorStrategy`
    - [x] Not use `getCmsPageManager`
    - [x] Not use `getSnapshotManager`
    - [x] Not use `getBlockManager`
    - [x] Not use `getPageManager`
    - [x] Not use `getSiteManager`
    - [x] Not use the code from BaseCommand
- [x] Create a new service to cleanup snapshots
- [x] Use the new service into the command
- [x] Fix the pipelines
- [x] Test clean snapshot (manual test :D )
- [x] Check if it's getting sites from `create snapshot`.
- [x] Check if it's getting sites for `sonata:page:update-core-routes` (There is  an issue here)


src/Consumer/CleanupSnapshotConsumer.php
```diff
@@ -41,6 +47,15 @@ class CleanupSnapshotConsumer implements ConsumerInterface
      */
     public function __construct(SnapshotManagerInterface $snapshotManager, PageManagerInterface $pageManager)
     {
+        @trigger_error(
```
eerison, line 48: @VincentLanglet should I add this deprecate message here or just the [`@deprecated`](https://github.com/sonata-project/SonataPageBundle/pull/1434/files#diff-572a33f9239c5fe1ab83911ea8986a3a6d7961771d8edc182a70c2efea8d3f6eR30) on top of class is enough?

VincentLanglet: It's better with the deprecation in the constructor like you did i would say.

VincentLanglet:
When do you plan/think you'll have time to do it ? :)

This will unblock https://github.com/sonata-project/SonataPageBundle/issues/1420 and https://github.com/sonata-project/SonataPageBundle/pull/1433.

src/Consumer/CleanupSnapshotConsumer.php
eerison: I thought the same, but it will broken the all tests that are not legacy

What do you think I add the `legacy group` for test that use this baseCommand and in the next major remove the group?

eerison:
> When do you plan/think you'll have time to do it ? :)
> 
> This will unblock #1420 and #1433.

Well I'm on that, I'm working few hours in the week.

But I'm not sure when I will finish this, because I need to cover it with test before start to change, there isn't any test for this command :/

But it's not too much

src/Consumer/CleanupSnapshotConsumer.php
VincentLanglet: Sure, you can add 
```
/**
 * NEXT_MAJOR: Remove the legacy group. 
 * @group legacy
 */
```

eerison:
Hey @VincentLanglet any idea why the XML lint is failing?

VincentLanglet:
> Hey @VincentLanglet any idea why the XML lint is failing?

No but I think you can run make lint-xml locally

The only file changed is https://github.com/sonata-project/SonataPageBundle/pull/1434/files#diff-0e4dafe909dac19649908b663e4b42f98822463c6b198483187a692e8fa15506

So maybe something is wrong, but i dont see it.

eerison:
> > Hey @VincentLanglet any idea why the XML lint is failing?
> 
> No but I think you can run make lint-xml locally
> 
> The only file changed is https://github.com/sonata-project/SonataPageBundle/pull/1434/files#diff-0e4dafe909dac19649908b663e4b42f98822463c6b198483187a692e8fa15506
> 
> So maybe something is wrong, but i dont see it.

Ok this XML problem I can check what is wrong, But this error here: https://github.com/sonata-project/SonataPageBundle/runs/7305503718?check_suite_focus=true#step:8:34

I can't add a `@group legacy` for https://github.com/sonata-project/SonataPageBundle/blob/3.x/tests/custom_bootstrap.php, any idea? 👀 

Edit 1: do you think, it's possible to ignore `custom_bootstrap.php`?

VincentLanglet:
Dunno if 
- we have a way to avoid this deprecation (sometimes we add an optional param to not trigger the deprecation)
- we should add some lines in phpunit.xml to accept those deprecations
- we should remove the trigger error

Any idea how to solve this @jordisala1991 ?

src/Command/BaseCommand.php
```diff
@@ -29,9 +29,28 @@
...
+ *
+ * @deprecated since 3.27, and it will be removed in 4.0.
  */
 abstract class BaseCommand extends ContainerAwareCommand
 {
+    public function __construct(?string $name = null)
+    {
+        parent::__construct($name);
+
+        if (false === strpos($this->getName(), 'doctrine:')) {
```
VincentLanglet: Sonata you meant ?

VincentLanglet: But this seems like a nice solution

eerison: Hey @VincentLanglet 

the commands from `custom_bootstrap` doesn't use the BaseCommand constructor, for this reason it was not working there!

I changed to `sonata:` then if found the command with sonata into the name the deprecation will show, But if someone extends this class and put other name that does't contain sonata it won't show :/

I thought to check by `doctrine:` but in theory any command different of `doctrine:` will show this deprecation, then I guess check with sonata is better :)

eerison: WDYT? is it fine?

VincentLanglet: Seems ok to me

eerison:
Ok, I changed the PR to ready to review, But just merge this after I mark all checkbox in my TODO list :) But it's just missing do some manual test in some project ;)

src/Command/BaseCommand.php
```diff
@@ -106,6 +111,21 @@ public function getNotificationBackend($mode)
...
+            @trigger_error(
+                sprintf(
+                    'The %s class is deprecate since sonata-project/page-bundle 3.27.0 and it will be remove in 4.0',
+                    __CLASS__
+                ),
+                \E_USER_DEPRECATED
+            );
+        }
+
+        return parent::run($input, $output); // TODO: Change the autogenerated stub
```
VincentLanglet:
```suggestion
        return parent::run($input, $output);
```

src/Command/CleanupSnapshotsCommand.php
```diff
@@ -62,36 +65,37 @@ protected function initialize(InputInterface $input, OutputInterface $output)
...
-                    $output->write(sprintf('<info>%s</info> - Cleaning up snapshots ...', $site->getName()));
-                }
+        $siteOption = $input->getOption('site');
+        $keepSnapshots = $input->getOption('keep-snapshots');
+
+        //NEXT_MAJOR: Inject GetSitesFromCommand $getSites
+        $getSites = $this->getContainer()->get('sonata.page.service.get_sites');
+
+        foreach ($getSites->findSitesById($siteOption) as $site) {
+            if ('async' === $input->getOption('mode')) {
```
VincentLanglet, line 92: NEXT MAJOR remove this ?

src/Command/CreateSnapshotsCommand.php
```diff
@@ -59,12 +59,22 @@ public function configure()
...
+        if ('php app/console' !== $input->getOption('base-console')) {
+            @trigger_error(
+                'The "base-console" is deprecated since sonata-project/page-bundle 3.27.0 and will be removed in 4.0',
+                \E_USER_DEPRECATED
+            );
+        }
+
         $siteOption = $input->getOption('site');
 
         if ([] === $siteOption) {
```
VincentLanglet, line 82: I think it could be interesting to consider that not providing a site means doing it on every site.

This way
`[] => findAll`
`[1, 2, 3] => findByIds`

This way it reduce the logic to find the sites, and we can use directly the siteManager rather than introducing a new service.

src/Service/Contract/GetSitesFromCommandInterface.php
```diff
@@ -0,0 +1,28 @@
...
+interface GetSitesFromCommandInterface
+{
+    public const ALL = 'all';
+
+    /**
+     * @param array<int>|array<self::ALL> $ids
+     *
+     * @return array<Site>
+     */
+    public function findSitesById(array $ids): array;
```
VincentLanglet: I don't think we need this interface.

Let's duplicate the logic in the two commands, and try to simplify the logic instead

cf comment in src/Command/CreateSnapshotsCommand.php

src/Service/GetSitesService.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace Sonata\PageBundle\Service;
+
+use Sonata\PageBundle\Exception\ParameterNotAllowedException;
+use Sonata\PageBundle\Model\Site;
+use Sonata\PageBundle\Model\SiteManagerInterface;
+use Sonata\PageBundle\Service\Contract\GetSitesFromCommandInterface;
+
+final class GetSitesService implements GetSitesFromCommandInterface
```
VincentLanglet: We don't need this

src/Exception/SonataPageException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Exception;
+
+class SonataPageException extends \Exception
```
VincentLanglet: We don't need this

src/Exception/ParameterNotAllowedException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Exception;
+
+class ParameterNotAllowedException extends SonataPageException
```
VincentLanglet: We don't need this

VincentLanglet (changes requested):

src/Exception/ParameterNotAllowedException.php
eerison: Why not? how do you group the exceptions? fro example

if I want to catch all exceptions from SonataPage like

```php
catch(SonataPageException $exception) ...
```

src/Service/GetSitesService.php
eerison: Why not? I moved the logical from getSites to here.

other thing that I was thinking is, we not access repository/manager directly from commands(WDYT?)

VincentLanglet: I seems fine to me to access to the siteManager in the command.

Currently we have the behavior
```
if (in_array('all', $ids)) {
     $sites = $this->siteManager->findAll();
} else {
     $sites = $this->siteManager->findByIds($ids);
}
```
I'd like to have in the next major
```
if ([] === $ids) {
     $sites = $this->siteManager->findAll();
} else {
     $sites = $this->siteManager->findByIds($ids);
}
```

And the logic seems simpler to me to not need a new service ; even if those 5 lines are duplicated in the two commands. This way
- We don't need the GetSite service
- We don't need the GetSite interface
- We don't need the ParamException
- We don't need the SonataException

src/Exception/ParameterNotAllowedException.php
VincentLanglet: It's just that if we don't need https://github.com/sonata-project/SonataPageBundle/pull/1434#discussion_r919977982 we then don't need any custom exception.

src/Service/Contract/GetSitesFromCommandInterface.php
eerison: well I prefer this in separated class instead of copy past around, But it's fine for me 👍🏼

src/Command/CreateSnapshotsCommand.php
eerison: Hey @VincentLanglet 
Well it was the first approach that I had implemented, But it was the argument to keep the `--site=all` argument

https://github.com/sonata-project/SonataPageBundle/pull/1418#discussion_r912350492

src/Command/CleanupSnapshotsCommand.php
eerison: true ;)

src/Command/CreateSnapshotsCommand.php
VincentLanglet: Didn't see this.

IMHO a command without argument can work.
And here the site is not an argument but an option, so it doesn't need to be mandatory.

But I don't know the impact of this command
- If running this command is risky, I can understand we prefer to force the list of sites.
- If running this command on all the sites, is not something risky, let's just allow it.

In the example of `doctrine:schema:update`, updating the schema is risky, so the `--force` argument makes sens.

src/Exception/ParameterNotAllowedException.php
eerison: Ok I got it :)

src/Command/CreateSnapshotsCommand.php
eerison: After we decide the final solution I will remove `--site='all'` from create snapshot too ;) and the docs again ;)

eerison: > Didn't see this.
> 
> IMHO a command without argument can work. And here the site is not an argument but an option, so it doesn't need to be mandatory.
> 
> But I don't know the impact of this command
> 
> * If running this command is risky, I can understand we prefer to force the list of sites.
> * If running this command on all the sites, is not something risky, let's just allow it.
> 
> In the example of `doctrine:schema:update`, updating the schema is risky, so the `--force` argument makes sense.

Well the `doctrine:schema:update` should be a command to run locally only, not in prod!
the `create snapshot` or `cleanup snapshot` are commands to run everywhere, then if you go in prod and run `create:snapshot`, it's not an accident, you went there and run this manually, there isn't other command or solution to do this in production ....

well I vote to run `create:snapshot` without `--site=all` (only numbers are possible to pass as optional --site=1 --site=2)✋🏼 

Note: 
`create snapshot:` in theory this command will be used only once in begging of the site, after the site is on fly, just the create snapshot by admin will be used, it was the argument to remove the `async` functionality by @core23, to remove the functionality that is not useful.

`cleanup snapshot`: I believe it will used rarely, after the screenshot table has too many row, it'll be used, and probably the guy will check the doc to see how it works or test locally before run in production, and even if he run `--site=all`, it won't work because we will try to find id in ('all'), and it will return empty, and nothing will be executed!

VincentLanglet: I meant, is it risky if you write `create:snapshot` and hit the `enter` button before adding `--site=...`.
or you can run this command 10 times on all or 1 specific site, without any issues.

eerison: I can check the behaviour if you want @VincentLanglet , But I guess nothing will be executed because in the worst case I will find ids by "all", like `id in ('all')`, and it will return a empty list of sites.

VincentLanglet: I meant if we decide that
```
create:snapshot
```
will run on every sites.

It will run https://github.com/sonata-project/SonataPageBundle/blob/89013450f4e78cc79232d386ed27b9d0870c3030/src/Service/CreateSnapshotService.php#L59-L73 on every page of every site, and dunno if it can lead to weird/unwanted behavior, or if we don't care if snapshot are created/enabled and the purpose of the site id was just to have faster commands.

eerison: Well IMO it won't have a `weird/unwanted behaviour`, because it'll have the same behaviour of `create:snapshot --site=all`, we can add the default parameter of site = `['all']`, and it will works like before.

Edit 1: well It'll be less code and checks to keep in the code 😃

VincentLanglet: Ok, so it seems natural to make optional the site option

so
```
create:snapshot
```
will be the same than
```
create:snapshot --site=all
```

And we deprecate using `--site=all`.

In next major it will be simpler
if ([] === $siteIds) findAll else findByIds

eerison: > Ok, so it seems natural to make optional the site option
> 
> so
> 
> ```
> create:snapshot
> ```
> 
> will be the same than
> 
> ```
> create:snapshot --site=all
> ```
> 
> And we deprecate using `--site=all`.
> 
> In next major it will be simpler if ([] === $siteIds) findAll else findByIds

agreed ;)

eerison:
Done @VincentLanglet, after this PR get the approves, it's able to merge ;)

VincentLanglet (approved):

jordisala1991 (approved):

PR #1436, opened 2022-07-08
PageExtension without deprecated InitRuntimeInterface

Hanmac:
## Subject

I am targeting this branch, because i don't know if that would count as BC.

it fixes:
> The "Sonata\PageBundle\Twig\Extension\PageExtension" class implements "Twig\Extension\InitRuntimeInterface" that is deprecated since Twig 2.7, to be removed in 3.0.

The Thing about `$env->getGlobals()['app']->getRequest()->getPathInfo();`, my gut is telling me that RequestStack would be cleaner, but that wouldn't be BC anymore if this one might be.

## Changelog

```markdown
### Changed
- PageExtension doesn't need InitRuntimeInterface anymore
```

### DIFF
--- a/src/Twig/Extension/PageExtension.php
+++ b/src/Twig/Extension/PageExtension.php
@@ -33,10 +33,8 @@ use Twig\TwigFunction;
  * PageExtension.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
- *
- * @final since sonata-project/page-bundle 3.26
  */
-class PageExtension extends AbstractExtension
+final class PageExtension extends AbstractExtension
 {
     /**
      * @var CmsManagerSelectorInterface
--- a/src/Twig/Extension/PageExtension.php
+++ b/src/Twig/Extension/PageExtension.php
@@ -26,6 +26,9 @@ use Symfony\Component\HttpKernel\Controller\ControllerReference;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
 use Twig\Environment;
+use Twig\Error\LoaderError;
+use Twig\Error\RuntimeError;
+use Twig\Error\SyntaxError;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
@@ -45,11 +48,6 @@ final class PageExtension extends AbstractExtension
      * @var SiteSelectorInterface
      */
     private $siteSelector;
-
-    /**
-     * @var array
-     */
-    private $resources;
 
     /**
      * @var RouterInterface
@@ -100,9 +98,11 @@ final class PageExtension extends AbstractExtension
     }
 
     /**
-     * @return string
+     * @throws LoaderError
+     * @throws RuntimeError
+     * @throws SyntaxError
      */
-    public function breadcrumb(Environment $twig, ?PageInterface $page = null, array $options = [])
+    public function breadcrumb(Environment $twig, ?PageInterface $page = null, array $options = []): string
     {
         if (!$page) {
             $page = $this->cmsManagerSelector->retrieve()->getCurrentPage();
@@ -140,7 +140,7 @@ final class PageExtension extends AbstractExtension
             }
         }
 
-        return $this->render($twig, $options['template'], [
+        return $twig->render($options['template'], [
             'page' => $page,
             'breadcrumbs' => $breadcrumbs,
             'options' => $options,
@@ -255,19 +255,5 @@ final class PageExtension extends AbstractExtension
         }
 
         return HttpKernelExtension::controller($controller, $attributes, $query);
-    }
-
-    /**
-     * @param string $template
-     *
-     * @return string
-     */
-    private function render(Environment $twig, $template, array $parameters = [])
-    {
-        if (!isset($this->resources[$template])) {
-            $this->resources[$template] = $twig->load($template);
-        }
-
-        return $this->resources[$template]->render($parameters);
     }
 }
--- a/src/Twig/Extension/PageExtension.php
+++ b/src/Twig/Extension/PageExtension.php
@@ -102,19 +102,9 @@ class PageExtension extends AbstractExtension
     }
 
     /**
-     * NEXT_MAJOR: remove this method.
-     *
-     * @deprecated since sonata-project/page-bundle 3.14, to be removed in version 4.0.
-     */
-    public function getName()
-    {
-        return 'sonata_page';
-    }
-
-    /**
      * @return string
      */
-    public function breadcrumb(Environment $env, ?PageInterface $page = null, array $options = [])
+    public function breadcrumb(Environment $twig, ?PageInterface $page = null, array $options = [])
     {
         if (!$page) {
             $page = $this->cmsManagerSelector->retrieve()->getCurrentPage();
@@ -152,7 +142,7 @@ class PageExtension extends AbstractExtension
             }
         }
 
-        return $this->render($env, $options['template'], [
+        return $this->render($twig, $options['template'], [
             'page' => $page,
             'breadcrumbs' => $breadcrumbs,
             'options' => $options,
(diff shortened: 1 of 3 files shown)


eerison:
@Hanmac can you open to 3.x? we should remove the deprecation from 3.x and merge into 4.x

Hanmac:
> @Hanmac can you open to 3.x? we should remove the deprecation from 3.x and merge into 4.x

@eerison you mean this changes should be BC for 3.x? or do you mean twig 3.x?

eerison:
> > @Hanmac can you open to 3.x? we should remove the deprecation from 3.x and merge into 4.x
> 
> @eerison you mean this changes should be BC for 3.x? or do you mean twig 3.x?

well because this deprecations are in 3.x, and to upgrade from 3.x to 4.x, the guy need to solve all deprecations and after that he is safe to migrate to next major.

for this reason we should solve the deprecations into the 3.x branch and after that merge into 4.x

src/Twig/Extension/PageExtension.php
```diff
@@ -267,13 +253,13 @@ public function renderBlock(PageBlockInterface $block, array $options = [])
      *
      * @return ControllerReference
      */
-    public function controller($controller, $attributes = [], $query = [])
+    public function controller(Environment $env, $controller, $attributes = [], $query = [])
```
eerison: why don't you inject this in `__construct`?

Hanmac: because https://www.lrotherfield.com/blog/symfony2-get-twig-environment-twig-extension

but what i would inject in construct is RequestStack

eerison: Well I would inject this in constructor and not in the method, because it's a dependency that your class needs to work, then IMO it should be injected on constructor.

eerison: https://symfony.com/doc/current/templates.html#rendering-a-template-in-services

VincentLanglet:
I disagree with @eerison, this should be done in 4.x.

In 3.x, we only to add deprecation about what will be changed/removed in 4.x about the PageBundle.
But there is no need to solve all the symfony/twig deprecation directly in 3.x. Especially because some of them require BC-break.

Here, not extending/implementing a class/interface is a BC-break. Sure there is almost no chance it will break someone code ; but we should stay with our BC-policy.

The strategy for the user will be the following
- Bump to SonataPage 4.x which will be compatible with Symfony 4 and 5 and Twig 2 and 3.
- Bump to Symfony 5.
- Bump to Twig 3.

src/Twig/Extension/PageExtension.php
eerison: and could you use `$twig` instead of `$env`, like symfony example, it is quite clear that you are inject twig class

eerison:
> 

hmmm I see, But in this case who uses `sonata page bundle` 3.x never will solve all deprecations, because some of them are inside of `sonata page code`.

Then what do you say, is we do not need to care about deprecations in `3.x` that was introduced from symfony, just solve them in the next major.

src/Twig/Extension/PageExtension.php
Hanmac: the problem is that this is a twig extension, not some general service

defining Twig Env in Constructor causes circular dependency problems

src/Twig/Extension/PageExtension.php
```diff
@@ -37,7 +36,7 @@
  *
  * @final since sonata-project/page-bundle 3.26
  */
-class PageExtension extends AbstractExtension implements InitRuntimeInterface
+class PageExtension extends AbstractExtension
```
eerison: Let this class as final , and remove the comment above

src/Twig/Extension/PageExtension.php
eerison: Hmmmm

eerison:
~~For me it's fine :)~~

There is a comment that you need to solve 👀

src/Twig/Extension/PageExtension.php
```diff
@@ -288,10 +262,10 @@ public function controller($controller, $attributes = [], $query = [])
      *
      * @return string
      */
-    private function render($template, array $parameters = [])
+    private function render(Environment $twig, $template, array $parameters = [])
```
eerison: but @Hanmac I was thinking we do not use render like this

usually we just call 

```php
$this->render('template.html.twig', []);
```

Hanmac: this function currently is only called by the `breadcrumb` function.
such breadcrumbs should only just be called once per page load?

if we don't need resources stored, then we could remove the render function and just call:
`$twig->render($template, $parameters)` directly

VincentLanglet:
> But in this case who uses sonata page bundle 3.x never will solve all deprecations, because some of them

It's not an issue.
People should just solve direct deprecations, not indirect ones.

src/Twig/Extension/PageExtension.php
Hanmac: @VincentLanglet your opinion? do we really need to store the twig templates when it most likely only render it once for breadcrumb?

VincentLanglet: I would say that since it was stored before, we should still doing it, in order to avoid to many changes.

VincentLanglet (dismissed):

src/Twig/Extension/PageExtension.php
Hanmac: because we are BC breaking anyway we might remove this part too

because breadcrumbs are only called once most likely, storing the loaded /not rendered template shouldn't even have any effect at all

src/Twig/Extension/PageExtension.php
```diff
@@ -86,48 +78,31 @@ class PageExtension extends AbstractExtension implements InitRuntimeInterface
...
             new TwigFunction('sonata_page_ajax_url', [$this, 'ajaxUrl']),
-            new TwigFunction('sonata_page_breadcrumb', [$this, 'breadcrumb'], ['is_safe' => ['html']]),
+            new TwigFunction('sonata_page_breadcrumb', [$this, 'breadcrumb'], ['is_safe' => ['html'], 'needs_environment' => true]),
             new TwigFunction('sonata_page_render_container', [$this, 'renderContainer'], ['is_safe' => ['html']]),
             new TwigFunction('sonata_page_render_block', [$this, 'renderBlock'], ['is_safe' => ['html']]),
             new TwigFunction('controller', [$this, 'controller']),
         ];
     }
 
-    public function initRuntime(Environment $environment): void
```
eerison, line 110: @VincentLanglet just to clarify

in this case, ins't it a BC broken?

if someone is using this  `initRuntime` method, and the people didn't get any deprecation message, it's a BC broken, isn't it?

Edit 1: the same for `breadcrumb` if add a new parameter, it will broken, won't it?

Hanmac: that's why i did it for the 4.x branch so i can do BC breaks

eerison: yes, But it wasn't deprecated in `3.x` then who uses `3.x` doesn't know that it was removed in `4.x`.

Hanmac: The function was already deprecated by twig

eerison: ahh Ok ;)

Hanmac: Also page Extension was ``@final`` so it shouldn't have been touched by users anyway

src/Twig/Extension/PageExtension.php
jordisala1991: It could be worth to simplify this code.

jordisala1991 (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

src/Twig/Extension/PageExtension.php
VincentLanglet: Can be done in another PR if needed :)

PR #1437, opened 2022-07-09
Fixes Validator Testcase

Hanmac:
## Subject

I am targeting this branch, because i don't know if BC or not.

It fixes this:
> 2x: Passing a null message when instantiating a "Symfony\Component\Validator\ConstraintViolation" is deprecated since Symfony 4.4.
1x in UniqueUrlValidatorTest::testValidateWithPageFound from Sonata\PageBundle\Tests\Validator
1x in UniqueUrlValidatorTest::testValidateWithRootUrlAndNoParent from Sonata\PageBundle\Tests\Validator

The main reason for this was the faulty Translator Mock used.
I updated the tests to check for the Violations.

## Changelog

```markdown
### Changed
- Updated ValidatorTest to use ConstraintValidatorTestCase
```

### DIFF
--- a/src/Resources/translations/validators.en.xliff
+++ b/src/Resources/translations/validators.en.xliff
@@ -12,6 +12,10 @@
         <target>The root URL '/' is already associated with another page of this site, please select a parent.
                 </target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>The Page needs Site to be set.</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.en.xliff
+++ b/src/Resources/translations/validators.en.xliff
@@ -14,7 +14,7 @@
       </trans-unit>
       <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
         <source>error.uniq_url.no_site</source>
-        <target>The Page needs Site to be set.</target>
+        <target>A website must be assigned to the page.</target>
       </trans-unit>
     </body>
   </file>
--- a/src/Resources/translations/validators.es.xliff
+++ b/src/Resources/translations/validators.es.xliff
@@ -12,10 +12,6 @@
         <target>La URL raíz '/' ya está asociada a otra página de este sitio, por favor selecciona un padre.
                 </target>
       </trans-unit>
-      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
-        <source>error.uniq_url.no_site</source>
-        <target>error.uniq_url.no_site</target>
-      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.es.xliff
+++ b/src/Resources/translations/validators.es.xliff
@@ -12,6 +12,10 @@
         <target>La URL raíz '/' ya está asociada a otra página de este sitio, por favor selecciona un padre.
                 </target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.fr.xliff
+++ b/src/Resources/translations/validators.fr.xliff
@@ -12,6 +12,10 @@
         <target>L'URL racine '/' est déjà associée à une autre page de ce site, veuillez sélectionner
                 un parent.</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.fr.xliff
+++ b/src/Resources/translations/validators.fr.xliff
@@ -14,7 +14,7 @@
       </trans-unit>
       <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
         <source>error.uniq_url.no_site</source>
-        <target>error.uniq_url.no_site</target>
+        <target>Un site Web doit être attribué à la page.</target>
       </trans-unit>
     </body>
   </file>
--- a/src/Resources/translations/validators.hu.xliff
+++ b/src/Resources/translations/validators.hu.xliff
@@ -11,10 +11,6 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
-      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
-        <source>error.uniq_url.no_site</source>
-        <target>error.uniq_url.no_site</target>
-      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.hu.xliff
+++ b/src/Resources/translations/validators.hu.xliff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.it.xliff
+++ b/src/Resources/translations/validators.it.xliff
@@ -12,10 +12,6 @@
         <target>L'URL radice '/' è già associato ad un'altra pagina di questo sito, selezionare un genitore.
                 </target>
       </trans-unit>
-      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
-        <source>error.uniq_url.no_site</source>
-        <target>error.uniq_url.no_site</target>
-      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.it.xliff
+++ b/src/Resources/translations/validators.it.xliff
@@ -12,6 +12,10 @@
         <target>L'URL radice '/' è già associato ad un'altra pagina di questo sito, selezionare un genitore.
                 </target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/src/Resources/translations/validators.ja.xliff
+++ b/src/Resources/translations/validators.ja.xliff
@@ -11,10 +11,6 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
-      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
-        <source>error.uniq_url.no_site</source>
-        <target>error.uniq_url.no_site</target>
-      </trans-unit>
     </body>
   </file>
 </xliff>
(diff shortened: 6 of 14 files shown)


src/Validator/UniqueUrlValidator.php
```diff
@@ -36,16 +39,20 @@ public function __construct(PageManagerInterface $manager)
...
+        if (!$value instanceof PageInterface) {
+            throw new UnexpectedValueException($value, PageInterface::class);
+        }
 
+        // at this it can only be SiteInterface or null
         if (!$value->getSite() instanceof SiteInterface) {
-            $this->context->addViolation('The page is not linked to a Site');
-
-            return;
+            throw new UnexpectedValueException($value->getSite(), SiteInterface::class);
```
VincentLanglet: Shouldn't it be a addViolation with a message ?

VincentLanglet:
> I am targeting this branch, because i don't know if BC or not.

Since all the deprecations will only be solved on 4.x. You can do most of all the PR on 4.x, it's easier.
The only needed PR on 3.x are the one which deprecate some PageBundle code.

src/Validator/UniqueUrlValidator.php
Hanmac: i'm following the route of other Validators there, for example the Language Validator throws this when it can't be turned into a String

https://github.com/symfony/symfony/blob/870eeb975feb1abb4b8a1722e1fd57beeab2b230/src/Symfony/Component/Validator/Constraints/LanguageValidator.php#L40-L42

VincentLanglet: I agree it should be done for `if (!$value instanceof PageInterface) {` because it means that the constraint is used on a wrong value.

But for the check `if (!$value->getSite() instanceof SiteInterface) {` I think it's better to have a message violation to say the user didn't selected a Site. (Like we had before)

Hanmac: for the Page -> getSite thing i wasn't sure when directly getSite returns something else than SiteInterface
imo that can only happen for null right?

Hanmac: otherwise if it is a message, it should be translated like `error.uniq_url` is?

VincentLanglet: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Model/PageInterface.php#L330-L332

So yes, either getSite is a site, either it's null.
And when it's null, it would mean that the user should select a site. (which is supposed to be always true https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Admin/PageAdmin.php#L298)

VincentLanglet: > otherwise if it is a message, it should be translated like `error.uniq_url` is?

Yes, using a translation key would be waaaaaay better indeed :)

Hanmac: I added better translatable Violation message

VincentLanglet (approved):

src/Validator/UniqueUrlValidator.php
```diff
@@ -36,14 +38,21 @@ public function __construct(PageManagerInterface $manager)
...
             return;
         }
+        if (!$value instanceof PageInterface) {
+            throw new UnexpectedValueException($value, PageInterface::class);
+        }
 
-        if (!$value->getSite() instanceof SiteInterface) {
-            $this->context->addViolation('The page is not linked to a Site');
+        if (null === $value->getSite()) {
+            $this->context->buildViolation('error.uniq_url.no_site')
```
VincentLanglet, line 50: We need to add translations for this key:
In https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Resources/translations/validators.en.xliff and other files.

VincentLanglet (changes requested):

src/Validator/UniqueUrlValidator.php
```diff
@@ -36,14 +38,21 @@ public function __construct(PageManagerInterface $manager)
 
     public function validate($value, Constraint $constraint): void
     {
-        if (!$value instanceof PageInterface) {
-            $this->context->addViolation('The page is not valid, expected a PageInterface');
-
+        if (!$constraint instanceof UniqueUrl) {
+            throw new UnexpectedTypeException($constraint, UniqueUrl::class);
```
jordisala1991, line 39: I think this is not correct, you cant receive another Constraint, otherwise this code wont be called.

This exception class is used when the value is not the type ypu expect to be (ex: I expect the value to be a Page and I am getting string).

Hanmac: All the other symfony validators do it

https://github.com/symfony/symfony/blob/870eeb975feb1abb4b8a1722e1fd57beeab2b230/src/Symfony/Component/Validator/Constraints/LanguageValidator.php#L32-L34

src/Validator/UniqueUrlValidator.php
Hanmac: I can try to add them, but im probably not good with translate them

VincentLanglet: You can add the one you know (I can give you the french one).

For others, either we use google translate, either we do nothing and rely on contribution when someone need it ;)

src/Validator/UniqueUrlValidator.php
VincentLanglet: Yeah, I'm doing the same in my project. To me it seems correct @jordisala1991

jordisala1991 (dismissed):
Except for the translations all good!

src/Validator/UniqueUrlValidator.php
Hanmac: added English and German translation

src/Resources/translations/validators.fr.xliff
```diff
@@ -12,6 +12,10 @@
         <target>L'URL racine '/' est déjà associée à une autre page de ce site, veuillez sélectionner
                 un parent.</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
```
VincentLanglet:
```suggestion
        <target>Un site Web doit être attribué à la page.</target>
```

src/Resources/translations/validators.en.xliff
```diff
@@ -12,6 +12,10 @@
         <target>The root URL '/' is already associated with another page of this site, please select a parent.
                 </target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>The Page needs Site to be set.</target>
```
VincentLanglet:
```suggestion
        <target>A website must be assigned to the page.</target>
```

VincentLanglet (changes requested):

VincentLanglet (approved):

src/Resources/translations/validators.es.xliff
```diff
@@ -12,6 +12,10 @@
         <target>La URL raíz '/' ya está asociada a otra página de este sitio, por favor selecciona un padre.
                 </target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
```
jordisala1991: Please dont add this copy unless you have the translation for it.

src/Resources/translations/validators.es.xliff
```diff
@@ -12,6 +12,10 @@
...
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.hu.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.it.xliff
```diff
@@ -12,6 +12,10 @@
         <target>L'URL radice '/' è già associato ad un'altra pagina di questo sito, selezionare un genitore.
                 </target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.ja.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.nl.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.pt_BR.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.ru.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>Корневой URL '/' уже связан с другой страницей этого сайта, пожалуйста, выберите родителя.</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.sk.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

src/Resources/translations/validators.sl.xliff
```diff
@@ -11,6 +11,10 @@
         <source>error.uniq_url.parent_unselect</source>
         <target>error.uniq_url.parent_unselect</target>
       </trans-unit>
+      <trans-unit id="error.uniq_url.no_site" resname="error.uniq_url.no_site">
+        <source>error.uniq_url.no_site</source>
+        <target>error.uniq_url.no_site</target>
+      </trans-unit>
```
VincentLanglet:
```suggestion
```

VincentLanglet (changes requested):

src/Resources/translations/validators.es.xliff
Hanmac: In the other files, the other translations do it too

VincentLanglet: We try to stop this because it's not a good practice.

With a placeholder, you get the placeholder
Without the translation, it default to the english one, which is better

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #1440, opened 2022-07-10
Solve multiple deprecations

VincentLanglet:


src/Admin/PageAdmin.php
```diff
@@ -167,6 +154,19 @@ public function setCacheManager(CacheManagerInterface $cacheManager)
         $this->cacheManager = $cacheManager;
     }
 
+    protected function configureBatchActions($actions): array
+    {
+        $actions = parent::configureBatchActions($actions);
+
+        $actions['snapshot'] = [
+            'label' => 'create_snapshot',
+            'translation_domain' => $this->getTranslationDomain(),
```
Hanmac: isn't that unnecessary? the translation domain defaults to the Admins Translation domain anyway

src/Admin/PageAdmin.php
```diff
@@ -266,8 +266,8 @@ protected function configureDatagridFilters(DatagridMapper $filter)
                 'field_options' => [
                     'required' => false,
                     'choices' => [
-                        'hybrid' => $this->trans('hybrid'),
-                        'cms' => $this->trans('cms'),
+                        'hybrid' => $this->getTranslator()->trans('hybrid', [], 'SonataPageBundle'),
```
Hanmac: wouldn't it just be enough to remove the 'choice_translation_domain' part?

then the translator would translate it anyway?

src/Admin/SnapshotAdmin.php
```diff
@@ -68,7 +68,7 @@ public function getBatchActions()
         $actions = parent::getBatchActions();
 
         $actions['toggle_enabled'] = [
-            'label' => $this->trans('toggle_enabled'),
+            'label' => $this->getTranslator()->trans('toggle_enabled', [], 'SonataPageBundle'),
```
Hanmac: wouldn't this be automatic translated by the translator anyway?

src/Admin/PageAdmin.php
VincentLanglet: I just changed getBatchActions to configureBatchActions without changing the code inside, but I can do this indeed.

src/Admin/PageAdmin.php
VincentLanglet: Indeed, should be better knwo

src/Admin/SnapshotAdmin.php
VincentLanglet: Indeed

jordisala1991 (approved):

src/Admin/PageAdmin.php
```diff
@@ -266,10 +265,10 @@ protected function configureDatagridFilters(DatagridMapper $filter)
                 'field_options' => [
                     'required' => false,
                     'choices' => [
-                        'hybrid' => $this->trans('hybrid'),
-                        'cms' => $this->trans('cms'),
+                        'hybrid' => 'hybrid',
+                        'cms' => 'cms',
                     ],
-                    'choice_translation_domain' => false,
+                    'choice_translation_domain' => 'SonataPageBundle',
```
Hanmac, line 271: i would have just removed that line and have it get the translation domain from the admin anyway

VincentLanglet: To me, the default value is not the admin translation domain here, but the one from ChoiceType, which is message.

So it's safer this way

Hanmac:
@VincentLanglet i think you missed this part:

https://github.com/sonata-project/SonataPageBundle/blob/c9c1fdce6e4025597c32616f72cc5b555d5ed0d0/src/Admin/BaseBlockAdmin.php#L182-L191

or was that on purpose?

VincentLanglet:
Missed it. Do you want to do the pr ?

Hanmac:
i will do later if you aren't faster

Hanmac:
can this change be done in 3.x or should that be in 4.x? The function was already marked as ``@final``

VincentLanglet:
> can this change be done in 3.x or should that be in 4.x? The function was already marked as `@final`

This can be done on 3.x because if you implements configurePersistentParameter, the `getPersistentParameter` method will give the same result than before.

PR #1441, opened 2022-07-10
Add rector

VincentLanglet:
https://github.com/sonata-project/dev-kit/pull/2106


jordisala1991:
I dont think it is possible to do on 3.x without breaking everything. It needs to be done on 4.x

VincentLanglet:
> I dont think it is possible to do on 3.x without breaking everything. It needs to be done on 4.x

Why would it break something ? The change made are supposed to be BC.
Do you have any BC-break in mind ?

But, I prefer to finish this PR after we solved https://github.com/sonata-project/SonataPageBundle/issues/1442 to avoid having wrong typehint.

jordisala1991:
Adding typehints to all properties is not Bc break?

VincentLanglet:
> Adding typehints to all properties is not Bc break?

I had no error for
```
class A
{
     protected A $foo;
}

class B extends A
{
	protected $foo;
}
```
But trying with another sandbox there is an error indeed. ^^'

I'll try to disable some rules.
We could add a check for page-bundle 3.x in https://github.com/sonata-project/dev-kit/blob/master/templates/project/rector.php.twig#L37

This would 
- allow to have rector on 4.x
- allow to remove the rector config key in dev-kit.

VincentLanglet:
Rector is not supposed to add BC break. I found the issue, it's the `@final` annotation.
I created an issue https://github.com/rectorphp/rector/issues/7289

eerison:
> > Adding typehints to all properties is not Bc break?
> 
> I had no error for
> 
> ```
> class A
> {
>      protected A $foo;
> }
> 
> class B extends A
> {
> 	protected $foo;
> }
> ```
> 
> But trying with another sandbox there is an error indeed. ^^'
> 
> I'll try to disable some rules. We could add a check for page-bundle 3.x in https://github.com/sonata-project/dev-kit/blob/master/templates/project/rector.php.twig#L37
> 
> This would
> 
> * allow to have rector on 4.x
> * allow to remove the rector config key in dev-kit.

What do you think to use `LevelSetList::UP_TO_PHP_73` and enable 7.4 only in 4.x branch?

VincentLanglet:
> What do you think to use LevelSetList::UP_TO_PHP_73 and enable 7.4 only in 4.x branch?

Dev kit automatically use the lower php version of the branch (which is now 7.4)
So anyway, it require some extra dev on dev-kit

VincentLanglet:
Now https://github.com/rectorphp/rector/issues/7289 is merged, this is BC @jordisala1991 
But I have to target dev-main atm.

src/Request/RequestFactory.php
```diff
@@ -66,17 +66,7 @@ private static function configureFactory($type)
...
-            array $request = [],
-            array $attributes = [],
-            array $cookies = [],
-            array $files = [],
-            array $server = [],
-            $content = null
-        ) {
-            return new SiteRequest($query, $request, $attributes, $cookies, $files, $server, $content);
-        });
+        Request::setFactory(static fn (array $query = [], array $request = [], array $attributes = [], array $cookies = [], array $files = [], array $server = [], $content = null) => new SiteRequest($query, $request, $attributes, $cookies, $files, $server, $content));
```
jordisala1991: This would be nice to do some line breaks.

src/Request/SiteRequestContext.php
```diff
@@ -31,10 +31,7 @@ class SiteRequestContext extends RequestContext implements SiteRequestContextInt
      */
     protected $selector;
 
-    /**
-     * @var SiteInterface
-     */
-    private $site;
+    private ?SiteInterface $site = null;
```
jordisala1991, line 34: is it nullable?

src/Twig/Extension/PageExtension.php
```diff
@@ -39,45 +39,24 @@
...
     /**
      * @var array
      */
     private $resources;
 
-    /**
-     * @var Environment
-     */
-    private $environment;
+    private ?Environment $environment = null;
```
jordisala1991, line 51: is it nullable?

tests/Entity/Snapshot.php
```diff
@@ -17,10 +17,7 @@
 
 final class Snapshot extends BaseSnapshot
 {
-    /**
-     * @var int
-     */
-    protected $id;
+    protected ?int $id = null;
```
jordisala1991, line 20: is it nullable?

tests/Site/HostPathSiteSelectorTest.php
```diff
@@ -155,9 +155,7 @@ final class HostPathSiteSelector extends BaseSiteSelector
      */
     public static function _camelize($property): string
     {
-        return preg_replace_callback('/(^|[_. ])+(.)/', static function ($match) {
-            return ('.' === $match[1] ? '_' : '').strtoupper($match[2]);
-        }, $property);
+        return preg_replace_callback('/(^|[_. ])+(.)/', static fn ($match) => ('.' === $match[1] ? '_' : '').strtoupper($match[2]), $property);
```
jordisala1991: I would break this in one line per argument.

tests/Site/HostSiteSelectorTest.php
```diff
@@ -115,9 +115,7 @@ final class HostSiteSelector extends BaseSiteSelector
      */
     public static function _camelize($property): string
     {
-        return preg_replace_callback('/(^|[_. ])+(.)/', static function ($match) {
-            return ('.' === $match[1] ? '_' : '').strtoupper($match[2]);
-        }, $property);
+        return preg_replace_callback('/(^|[_. ])+(.)/', static fn ($match) => ('.' === $match[1] ? '_' : '').strtoupper($match[2]), $property);
```
jordisala1991: same as before

src/Model/SnapshotPageProxy.php
```diff
@@ -45,12 +36,9 @@ class SnapshotPageProxy implements SnapshotPageProxyInterface
     /**
      * @var PageInterface[]
      */
-    private $parents;
+    private ?array $parents = null;
```
jordisala1991, line 43: is it nullable?

src/Model/SnapshotPageProxy.php
```diff
@@ -22,20 +22,11 @@
...
-     * @var SnapshotInterface
-     */
-    private $snapshot;
+    private SnapshotInterface $snapshot;
 
-    /**
-     * @var PageInterface
-     */
-    private $page;
+    private ?PageInterface $page = null;
```
jordisala1991, line 29: is it nullable?

jordisala1991 (commented):
Nice, it will be hard to merge to 4.x tho

src/Request/SiteRequestContext.php
VincentLanglet: It's not initiliazed.

```
final public function setSite(SiteInterface $site)
    {
        $this->site = $site;
    }

    final public function getSite()
    {
        if (!$this->site instanceof SiteInterface) {
            $this->site = $this->selector->retrieve();
        }

        return $this->site;
    }
```
Coming from a setter.

src/Twig/Extension/PageExtension.php
VincentLanglet: Not initialized in the constructor but in
```
public function initRuntime(Environment $environment)
    {
        $this->environment = $environment;
    }
```

src/Model/SnapshotPageProxy.php
VincentLanglet: Yes when it's not initialized 
```
public function getParents()
    {
        if (!$this->parents) {
```

src/Model/SnapshotPageProxy.php
VincentLanglet: Yes when it's not loaded
```
private function load()
    {
        if (!$this->page && $this->transformer) {
            $this->page = $this->transformer->load($this->snapshot);
        }
    }
```

tests/Entity/Snapshot.php
VincentLanglet: It's not initialized.

jordisala1991 (approved):

composer.json
```diff
@@ -65,6 +65,7 @@
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.16",
         "psalm/plugin-symfony": "^3.0",
+        "rector/rector": "dev-main",
```
jordisala1991, line 68: I think rector just released a patch version. Can you see if the fix is included?

VincentLanglet:
It is invluded

PR #1443, opened 2022-07-12
Remove BaseBlockAdmin getPersistentParameters

Hanmac:
configurePersistentParameters is already below, and it isn't in 3.x

### DIFF
--- a/src/Admin/BaseBlockAdmin.php
+++ b/src/Admin/BaseBlockAdmin.php
@@ -179,17 +179,6 @@ abstract class BaseBlockAdmin extends AbstractAdmin
         $this->containerBlockTypes = $containerBlockTypes;
     }
 
-    public function getPersistentParameters(): array
-    {
-        if (!$this->hasRequest()) {
-            return [];
-        }
-
-        return [
-            'type' => $this->getRequest()->get('type'),
-        ];
-    }
-
     public function preBatchAction($actionName, ProxyQueryInterface $query, array &$idx, $allElements): void
     {
         $parent = $this->getParent();


src/Admin/BaseBlockAdmin.php
```diff
@@ -179,17 +179,6 @@ public function setContainerBlockTypes(array $containerBlockTypes): void
         $this->containerBlockTypes = $containerBlockTypes;
     }
 
-    public function getPersistentParameters(): array
```
eerison, line 182: - is there any reason to remove this?
- have you checked if it has used in the code, for example: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Controller/BlockAdminController.php#L86
- ~~If you want to remove this method in `4.x`, you should deprecate in `3.x` ?!~~ remove method without any deprecate message or annotation is fine too right @VincentLanglet ?

Hanmac: * reason: the parent method is or is going to be final:
https://github.com/sonata-project/SonataAdminBundle/blob/29fc2f711f5eaf8108605f9d09b3c9f44ad374f3/src/Admin/AbstractAdmin.php#L1457
* the method exist in 4.x but for some reason not in 3.x
* the correct method  configurePersistentParameters already exist in 3.x

VincentLanglet: Indeed there is
```
    protected function configurePersistentParameters(): array
    {
        if (!$this->hasRequest()) {
            return [];
        }
        return [
            'type' => $this->getRequest()->get('type'),
        ];
    }
```
doing the job

VincentLanglet (approved):

src/Admin/BaseBlockAdmin.php
eerison: well in theory, if you remove this method, it'll throw an exception, have you tested this?

VincentLanglet: It's already implemented in the AbstractAdmin

Issue #1417, opened 2022-04-22
Remove deprecated code


eerison:
### ⚠️ Deprecations that need to be solved for **Symfony 5** upgrade.

> **Note**
> those deprecations are waiting for this PR https://github.com/sonata-project/SonataAdminBundle/pull/7867

Remaining direct deprecation notices (6)
  - [ ] https://github.com/sonata-project/SonataAdminBundle/pull/7867 1x: The "Sonata\AdminBundle\Admin\AbstractAdmin::getBaseRoutePattern()" method is considered final since sonata-project/admin-bundle 3.102. It may change without further notice as of its next major version. You should not extend it from "Sonata\PageBundle\Admin\SharedBlockAdmin".
    1x in BaseBlockAdminTest::testSettingAsEditedOnPreBatchDeleteAction from Sonata\PageBundle\Tests\Admin
  -  [ ] https://github.com/sonata-project/SonataAdminBundle/pull/7867 1x: The "Sonata\AdminBundle\Admin\AbstractAdmin::getBaseRouteName()" method is considered final since sonata-project/admin-bundle 3.102. It may change without further notice as of its next major version. You should not extend it from "Sonata\PageBundle\Admin\SharedBlockAdmin".
    1x in BaseBlockAdminTest::testSettingAsEditedOnPreBatchDeleteAction from Sonata\PageBundle\Tests\Admin
- [x] #1483 1x: The "Sonata\BlockBundle\Block\BlockContextManager" class is considered final since sonata-project/block-bundle 3.0. It may change without further notice as of its next major version. You should not extend it from "Sonata\PageBundle\Block\BlockContextManager".
    1x in BlockContextManagerTest::testGetWithValidData from Sonata\PageBundle\Tests\Block
- [x] #1483 1x: The "Sonata\BlockBundle\Block\Service\ContainerBlockService" class is considered final since sonata-project/block-bundle 3.0. It may change without further notice as of its next major version. You should not extend it from "Sonata\PageBundle\Block\ContainerBlockService".
    1x in ContainerBlockServiceTest::testExecute from Sonata\PageBundle\Tests\Block
- [x] #1434 1x: The "Sonata\PageBundle\Command\BaseCommand" class extends "Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand" that is deprecated since Symfony 4.2, use {@see Command} instead.
[...]

PR #1444, opened 2022-07-12
Change sonata.templating by twig where it's possible

Daric971:
## Subject

I'm targeting this branch, because [4.x](https://github.com/sonata-project/SonataPageBundle/tree/4.x) will be compatible with Symfony 5

### DIFF
--- a/src/Block/ChildrenPagesBlockService.php
+++ b/src/Block/ChildrenPagesBlockService.php
@@ -15,7 +15,7 @@ namespace Sonata\PageBundle\Block;
 
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\BlockBundle\Block\BlockContextInterface;
-use Sonata\BlockBundle\Block\Service\AbstractAdminBlockService;
+use Sonata\BlockBundle\Block\Service\AbstractBlockService;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\Form\Type\ImmutableArrayType;
 use Sonata\PageBundle\CmsManager\CmsManagerSelectorInterface;
@@ -26,33 +26,22 @@ use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
-use Symfony\Component\Templating\EngineInterface;
+use Twig\Environment;
 
 /**
  * Render children pages.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
- *
- * @final since sonata-project/page-bundle 3.26
  */
-class ChildrenPagesBlockService extends AbstractAdminBlockService
+final class ChildrenPagesBlockService extends AbstractBlockService
 {
-    /**
-     * @var SiteSelectorInterface
-     */
-    protected $siteSelector;
+    private SiteSelectorInterface $siteSelector;
 
-    /**
-     * @var CmsManagerSelectorInterface
-     */
-    protected $cmsManagerSelector;
+    private CmsManagerSelectorInterface $cmsManagerSelector;
 
-    /**
-     * @param string $name
-     */
-    public function __construct($name, EngineInterface $templating, SiteSelectorInterface $siteSelector, CmsManagerSelectorInterface $cmsManagerSelector)
+    public function __construct(Environment $twig, SiteSelectorInterface $siteSelector, CmsManagerSelectorInterface $cmsManagerSelector)
     {
-        parent::__construct($name, $templating);
+        parent::__construct($twig);
 
         $this->siteSelector = $siteSelector;
         $this->cmsManagerSelector = $cmsManagerSelector;
@@ -122,7 +111,7 @@ class ChildrenPagesBlockService extends AbstractAdminBlockService
         ]);
     }
 
-    public function getName()
+    public function getName(): string
     {
         return 'Children Page (core)';
     }
--- a/src/Listener/ExceptionListener.php
+++ b/src/Listener/ExceptionListener.php
@@ -17,108 +17,76 @@ use Psr\Log\LoggerInterface;
 use Sonata\PageBundle\CmsManager\CmsManagerSelectorInterface;
 use Sonata\PageBundle\CmsManager\DecoratorStrategyInterface;
 use Sonata\PageBundle\Exception\InternalErrorException;
+use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Page\PageServiceManagerInterface;
 use Sonata\PageBundle\Site\SiteSelectorInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
-use Symfony\Component\Templating\EngineInterface;
+use Twig\Environment;
 
 /**
  * ExceptionListener.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
- *
- * @final since sonata-project/page-bundle 3.26
  */
-class ExceptionListener
+final class ExceptionListener
 {
-    /**
-     * @var SiteSelectorInterface
-     */
-    protected $siteSelector;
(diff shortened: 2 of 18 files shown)


src/Block/ChildrenPagesBlockService.php
```diff
@@ -26,33 +26,22 @@
...
-
-    /**
-     * @var CmsManagerSelectorInterface
-     */
-    protected $cmsManagerSelector;
-
-    /**
-     * @param string $name
-     */
-    public function __construct($name, EngineInterface $templating, SiteSelectorInterface $siteSelector, CmsManagerSelectorInterface $cmsManagerSelector)
```
eerison, line 53: Sorry to ask this again, But I'm still getting what is BC break or not 😶‍🌫️ 

in this case `ChildrenPagesBlockService` is not `final` in `3.x`, and was removed a parameter from constructor

then in case someone extends this class it'll throw an exception, and as this class is from SonataPageBundle, should we provide BC in `3.x` branch?

src/Listener/ExceptionListener.php
```diff
@@ -17,123 +17,87 @@
...
- *
- * @final since sonata-project/page-bundle 3.26
  */
-class ExceptionListener
+final class ExceptionListener
 {
-    /**
-     * @var SiteSelectorInterface
-     */
-    protected $siteSelector;
```
eerison, line 40: if I remember correct it's a BC break

I saw that was added a comment like thin in rector PR.

src/Block/ChildrenPagesBlockService.php
jordisala1991: This can't be changed in a BC way without a lot of trouble. Probably it can be done, but it is not worth it, the idea is to release 4.0 as soon as possible.

eerison: I will wait for @VincentLanglet confirmation before do the complete review, But in the end what I think that could be a problem is just add `type` in the code and change the property visibility like from `protected` to `private`.

I'm just thinking in case someone extends of those class.

Note: I mean it's a problem to add the type because there ins't any deprecate message in `3.x` saying that, it'll be changed in `4.x`

src/Listener/ExceptionListener.php
VincentLanglet: We're on 4.x branch. BC-break are allowed.

src/Block/ChildrenPagesBlockService.php
VincentLanglet: In SonataAdminBundle we added typehint everywhere, and final as possible.
We'll follow the same strategy here.
Same for restricting visibility.

For everything we declare as final in 4.x ; a `@final` annotation is supposed to be added in 3.x.
For typehint in 4.x ; nothing to do on 3.x
For visibility ; if the class is/will be final, nothing to do. If it's not, we can sometimes add deprecation message, but it's not mandatory.

src/Model/SnapshotChildrenCollection.php
```diff
@@ -13,48 +13,41 @@
...
-     * @var PageInterface
-     */
-    protected $page;
+    private PageInterface $page;
 
-    /**
-     * @var array
-     */
-    protected $collection;
+    private Collection $collection;
```
VincentLanglet:
```suggestion
    private ?Collection $collection = null;
```

src/Model/SnapshotChildrenCollection.php
```diff
@@ -13,48 +13,41 @@
...
 
     public function __construct(TransformerInterface $transformer, PageInterface $page)
     {
         $this->transformer = $transformer;
         $this->page = $page;
+        $this->collection = new ArrayCollection();
```
VincentLanglet:
```suggestion
```

It will be initialized in the load method

src/Block/ChildrenPagesBlockService.php
eerison: - and for remove property without `@deprecation` annotation?
- and for add or remove params in `__construct` or other `method`, is it fine too?

src/Model/SnapshotChildrenCollection.php
VincentLanglet: And it's currently broken because of the line
```
$this->collection = $this->transformer->getChildren($this->page);
```
and getChildren returns an array.

VincentLanglet (changes requested):

src/Block/ChildrenPagesBlockService.php
VincentLanglet: > * and for remove property without `@deprecation` annotation?
> * and for add or remove params in `__construct` or other `method`, is it fine too?

Depends on the situation.
Sometime we can accept `Foo|DeprecatedThings` in the constructor, so we add a deprecation.
And for property, we try to add deprecation indeed, if people can avoid to use it ; and the class is not final.

Here, BlockBundle require to change from templating to twig. There is already a deprecation coming from BlockBundle. And we cant do anything easy to solve the deprecation in 3.x.
We doing a "BC-break without deprecation" here because there is one unsolvable from another dependency.

The same way, we'll do lot of similar things when adding SF5 support.

eerison: Ok, thanks for the explanation :)

Daric971: So, for that review? there is no problem?

VincentLanglet: Except my review on SnapshotChildrenCollection

eerison: for me it's good :)

VincentLanglet (dismissed):

src/Model/SnapshotChildrenCollection.php
```diff
@@ -13,48 +13,40 @@
 
 namespace Sonata\PageBundle\Model;
 
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+
```
VincentLanglet, line 18:
```suggestion
```

src/Model/SnapshotChildrenCollection.php
```diff
@@ -13,48 +13,40 @@
...
-     * @var PageInterface
-     */
-    protected $page;
+    private PageInterface $page;
 
-    /**
-     * @var array
-     */
-    protected $collection;
+    private ?Collection $collection = null;
```
VincentLanglet, line 30:
```suggestion
    /**
     * @var array
     */
    private $collection;
```

VincentLanglet (dismissed):

eerison:
@Daric971 do you know if this ~~ticket~~ PR solve one of those deprecations: #1417?

src/Model/SnapshotChildrenCollection.php
VincentLanglet:
```suggestion
```

src/Model/SnapshotChildrenCollection.php
VincentLanglet:
```suggestion
    /**
     * @var array
     */
    private $collection;
```
Let's this untouched, since it's broken so far.

VincentLanglet (changes requested):

VincentLanglet:
I was fixing your PR @Daric971, but you force pushed my changes :/

src/Model/SnapshotChildrenCollection.php
Daric971: sorry, I just remove ArrayCollection line
private ?Collection $collection = null;

src/Block/ChildrenPagesBlockService.php
Daric971: So, what can I do?

src/Model/SnapshotChildrenCollection.php
Daric971: done

VincentLanglet (approved):

VincentLanglet:
Thanks

src/Listener/ExceptionListener.php
```diff
@@ -17,123 +17,87 @@
...
         $this->siteSelector = $siteSelector;
         $this->cmsManagerSelector = $cmsManagerSelector;
         $this->debug = $debug;
-        $this->templating = $templating;
+        $this->twig = $twig;
         $this->pageServiceManager = $pageServiceManager;
         $this->decoratorStrategy = $decoratorStrategy;
         $this->httpErrorCodes = $httpErrorCodes;
         $this->logger = $logger;
+        $this->status = $status;
```
jordisala1991, line 75: what is this status?

PR #1446, opened 2022-07-12
SnapshotManager: use QueryBuilder

Hanmac:
## Subject

I am targeting this branch, because it should be BC.

Still Draft, i need to update the other queries, and i need to write a functional testcase using KernelTestCase and a Test Database

## Changelog

```markdown
### Changed
- Updated SnapshotManger to use QueryBuilder instead of native SQL
```


src/Entity/SnapshotManager.php
```diff
@@ -99,16 +100,17 @@ public function enableSnapshots(array $snapshots, ?\DateTime $date = null)
...
-        );
 
-        $this->getConnection()->query($sql);
+        $qb = $this->getRepository()->createQueryBuilder('s');
+        $q = $qb->update()
+            ->set('s.publicationDateEnd', ':date_end')
+            ->where($qb->expr()->notIn('s.id', $snapshotIds))
+            ->andWhere($qb->expr()->in('s.page_id', $pageIds))
+            ->andWhere($qb->expr()->isNull('s.publicationDateEnd'))
+            ->setParameter('date_end', $date, Types::DATETIME_MUTABLE)
```
eerison: why this type `Types::DATETIME_MUTABLE`?

Hanmac: in my attempt to do it with Mock tests, it was failing because this:
> Binding entities to query parameters only allowed for entities that have an identifier.
Class "DateTime" does not have an identifier.

thats why i said i need another tests where i can do functional tests

Hanmac: seems it worked without, the mocking was incomplete

Hanmac: Also that why:
>  QueryBuilderSetParameter: To improve performance set explicit type for objects

https://github.com/doctrine/orm/issues/8113
https://github.com/psalm/psalm-plugin-symfony/issues/158

Hanmac:
first query is done, one more to go

Hanmac:
it seems it doesn't crash on newer doctrine orm, i need to debug that more

VincentLanglet:
> it seems it doesn't crash on newer doctrine orm, i need to debug that more

You're allowed to bump the doctrine orm version if needed

src/Entity/SnapshotManager.php
```diff
@@ -244,7 +244,7 @@ public function cleanup(PageInterface $page, $keep)
 
         // Subquery DQL doesn't support Limit
         $innerQb
-            ->select('i.id')
+            ->select('DISTINCT i.id')
```
eerison: I guess distinct just work for MySQL

eerison: Try to use group by

src/Entity/SnapshotManager.php
```diff
@@ -231,60 +233,40 @@ public function cleanup(PageInterface $page, $keep)
...
+
+        // try a better Function expression for this?
+        $ifNullExpr = sprintf(
+            'CASE WHEN %s THEN 1 ELSE 0 END',
+            $expr->isNull('i.publicationDateEnd')
+        );
+
+        // Subquery DQL doesn't support Limit
+        $innerQb
+            ->select('DISTINCT i.id')
```
eerison: Distinct is a MySQL function, I guess you can have the same result using group by

src/Entity/SnapshotManager.php
Hanmac: DISTINCT wasn't needed it seems

VincentLanglet (dismissed):

src/Entity/SnapshotManager.php
```diff
@@ -231,60 +233,40 @@ public function cleanup(PageInterface $page, $keep)
...
+        // Subquery DQL doesn't support Limit
+        $innerQb
+            ->select('i.id')
+            ->where($expr->eq('i.page', $page->getId()))
+            ->orderBy($ifNullExpr, Criteria::DESC)
+            ->addOrderBy('i.publicationDateEnd', Criteria::DESC)
+            ->setMaxResults($keep);
+
+        $query = $innerQb->getQuery();
+        if (method_exists($query, 'getSingleColumnResult')) {
```
eerison: Hey @Hanmac why do you need this?

isn't it possible to get the results using fecthAll(FETCH_ARRAY) or something like this?

Hanmac: getSingleColumnResult is doctrine orm 2.10 or newer

VincentLanglet: It was introduced in 2.10.x. 

@jordisala1991 WDYT ? Should we add a stricter conflict about the version ? 
Or always using the else branch to avoid extra if/else ?

Hanmac: i tried to raise the dependencies as low as possible, but i will follow your decisions

src/Entity/SnapshotManager.php
```diff
@@ -231,60 +233,40 @@ public function cleanup(PageInterface $page, $keep)
...
-                $tableName,
-                $page->getId(),
-                $keep
+        $qb = $this->getRepository()->createQueryBuilder('s');
+        $expr = $qb->expr();
+        $qb->delete()
+            ->where($expr->eq('s.page', $page->getId()))
+            ->andWhere($expr->notIn(
+                's.id',
+                $innerArray
```
eerison, line 262: Have you tried to use `$innerQb->getQuery()` here?

src/Entity/SnapshotManager.php
Hanmac: same with this one:

```
        // for older doctrine dbal
        if (method_exists(Connection::class, 'executeStatement')) {
            $method = 'executeStatement';
        } else {
            $method = 'executeUpdate';
        }
```

src/Entity/SnapshotManager.php
Hanmac: yes i tried
> // Subquery DQL doesn't support Limit

Hanmac: but i might try again inserting the final sql of the subquery as literal

src/Entity/SnapshotManager.php
eerison: ~~why don't you add the minimal doctrine version: `^2.10`? here: https://github.com/sonata-project/SonataPageBundle/pull/1446/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R26~~

Edit 1: sorry I misunderstood the dependencies 😄!

well if we will strict in version less then `2.8` , I guess we can put `2.10` because the difference of installations is not too much.

Screenshot 2022-07-18 at 10 12 59

VincentLanglet: I think bumping the minimum minor version is totally fine

Hanmac: doctrine bundle != doctrine orm != doctrine dbal version

i need to see which one i need to raise more

src/Entity/SnapshotManager.php
eerison: But it's fine insert the literal SQL, because it'll be generated by `doctrine`, the problem to use the sql manually is different sql syntax for different DB, and we need to add some `if/else` like before for `mysql` and `oracle`.

But in this case Doctrine will care about this and generate the correct sql, I guess it's better then go in the database twice!

Hanmac: like i said before, i was using DQL and doctrine was okay with it, but that was ignoring Limit

i will try again with SQL, but i need to mark it so Doctrine isn't trying to parse it again

src/Entity/SnapshotManager.php
eerison: and in theory it should be fine, it's just a minor version without BC breaks.

eerison: > doctrine bundle != doctrine orm != doctrine dbal version
> 
> i need to see which one i need to raise more

yep, I misunderstood between `doctrine bundle` and `doctrine orm` 😅

src/Entity/SnapshotManager.php
eerison: But it's not a blocker for you PR @Hanmac, this command will be used rarely then it's fine we go in the database twice!

if you see that you will invest too much energy on that, please just ignore this :)

Hanmac: doctrine doesn't seems to like to insert the generated SQL (with Limit) into the Query Builder for the outer Query.

a suggestion would be to add a DQL Function that gives the SubSelect the Limit back when turning DQL into SQL,
something like that:
https://stackoverflow.com/questions/24068947/subquery-with-limit-in-doctrine/46663949#46663949

eerison: well I guess it's too much to be used only in one command, I guess if we need more subqueries in others places, we could use this approach :)

src/Entity/SnapshotManager.php
Hanmac: i bumped doctrine dependencies

eerison:
for me it's fine :) , good job @Hanmac 👏🏼

eerison:
@Hanmac can you just update your PR's description please! specially the `Changelog`

VincentLanglet (dismissed):

composer.json
```diff
@@ -23,8 +23,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "cocur/slugify": "^3.0 || ^4.0",
-        "doctrine/doctrine-bundle": "^1.12.3 || ^2.0",
-        "doctrine/persistence": "^1.3.4 || ^2.0",
```
jordisala1991, line 27: you should not remove this requirement. It is used on the code. Maybe you want to drop lower version?

Here you have the requirements for 4.x branch:

```
        "doctrine/doctrine-bundle": "^2.5",
        "doctrine/persistence": "^2.1",
```

jordisala1991 (dismissed):

tests/Functional/Snapshot/SnapshotManagerTest.php
```diff
@@ -0,0 +1,186 @@
...
+        // asking if entity manager contains entity isn't enough
+        $repo = $this->entityManager->getRepository(SonataPageSnapshot::class);
+        $all = $repo->findAll();
+        static::assertCount(1, $all);
+        static::assertContains($snapshot, $all);
+        static::assertNotContains($snapshot2, $all);
+
+        // TODO the entities are still in the repo by ID
+//        static::assertNotNull($repo->find(123));
+//        static::assertNull($repo->find(789));
```
jordisala1991: How should we handle this?

tests/bootstrap.php
```diff
@@ -31,10 +31,5 @@
 }
 
 if (file_exists($file = __DIR__.'/custom_bootstrap.php')) {
-    /**
-     * @psalm-suppress MissingFile,UnusedPsalmSuppress
-     *
-     * @see https://github.com/vimeo/psalm/issues/3886
-     */
```
jordisala1991: You can't touch this file.

Hanmac: my psalm was telling me "unneeded suppression"

> vendor/bin/psalm --no-progress --show-info=false --stats --output-format=github --shepherd --php-version=8.1
> ::error file=tests/bootstrap.php,line=35,col=24::This suppression is never used (see https://psalm.dev/207)
> ::error file=tests/bootstrap.php,line=35,col=36::This suppression is never used (see https://psalm.dev/207)

tests/Functional/Snapshot/SnapshotManagerTest.php
Hanmac: it is more or less a design from ORM: https://github.com/doctrine/orm/discussions/9908#discussioncomment-3169982

i could extend the test case more to see what happens after clearing the entity manager

tests/bootstrap.php
jordisala1991: But this class is manager by dev-kit, if you remove lines here, dev-kit will add them again. Can you please add them to see if the build fails?

tests/Functional/Snapshot/SnapshotManagerTest.php
jordisala1991: I prefer not to let pending todos. If those 2 lines are not needed, please remove them.

tests/bootstrap.php
Hanmac: still seems to pass, curious, maybe i have the wrong psalm version (i did the recent composer tests with lowest)

tests/Functional/Snapshot/SnapshotManagerTest.php
Hanmac: i added some checks that can be done before and after entity manager clear

the entity classes aren't allowed to be final, for the DoctrineProxy to work

jordisala1991 (approved):

PR #1459, opened 2022-07-15
Deprecate Mustache generator

### LINKED ISSUE #1458: Remove Mustache generator
I was checking the code that was not coveraged by test and I saw that, there is this [Mustache](https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Generator/Mustache.php) class is not used anywhere,  it's not mentioned in the documentation and in case it's related with [mustache.php](https://github.com/bobthecow/mustache.php) I do not see an active maintenance into the project anymore.

Then I guess we could deprecate it and remove in the next major

eerison:
## Remove Mustache generator

Closes #1458 .

## Changelog

```markdown
### Deprecated
- Deprecated `Mustache:class` generator
```

### DIFF
--- a/src/Generator/Mustache.php
+++ b/src/Generator/Mustache.php
@@ -19,9 +19,25 @@ namespace Sonata\PageBundle\Generator;
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
  * @final since sonata-project/page-bundle 3.26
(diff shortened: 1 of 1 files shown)


VincentLanglet (approved):

jordisala1991:
Thanks @eerison

PR #1460, opened 2022-07-15
Remove old code related with Commands

eerison:
## Subject

I am targeting this branch, because {reason}.

Closes #1420 #1415

## Changelog

```markdown
### Removed
- Removed `--site=all` option for the commands
- Removed `DumpPageCommand::class`
- Removed `RenderBlockCommand::class`
- Removed `BaseCommand::class`

### Changed
- Changed `BaseCommand::class` to `Symfony\Command::class`
```

## To do

Create tickets:

- [x] #1465 Screenshot 2022-07-15 at 14 04 01 https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/CreateBlockContainerCommand.php#L62
- [x] https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/DumpPageCommand.php#L79 provide some solution to not get the service from container.
- [x] #1465 check if [MigrateBlockNameSettingCommand](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/MigrateBlockNameSettingCommand.php) and [MigrateToJsonTypeCommand](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/MigrateToJsonTypeCommand.php) are necessary 
- [x] Check if RenderBlockCommand is used (probably not):
    - it's trying to find a service that doesn't exist sonata.block.context_manager
    - and using two method from containerInterface that doesn't exist anymore : enterScope and leaveScope

PR #1461, opened 2022-07-15
Deprecate dump commands

eerison:
## Deprecate dump commands

I am targeting this branch, because those commands are not final in this release

## Changelog

```markdown
### Deprecated
- Deprecated `DumpPageCommand::class`
- Deprecated `RenderBlockCommand::class`
```

It's related with this PR: https://github.com/sonata-project/SonataPageBundle/pull/1460#issuecomment-1185559382

and it'll be necessary to remove #1420

### DIFF
--- a/src/Command/DumpPageCommand.php
+++ b/src/Command/DumpPageCommand.php
@@ -24,6 +24,10 @@ use Symfony\Component\Console\Output\OutputInterface;
  * Migrates the name setting of all blocks into a code setting.
  *
  * @final since sonata-project/page-bundle 3.26
+ *
+ * NEXT_MAJOR: Remove this class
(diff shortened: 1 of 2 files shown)


jordisala1991 (approved):
If we are sure those commands are not useful, lets burn them.

VincentLanglet (approved):
You the one with a project using this bundle @eerison ; so according to you this command are not useful ? (if we fix them)

eerison:
> You the one with a project using this bundle @eerison ; so according to you this command are not useful ? (if we fix them)

We never used those commands, and if someone use, they have opened a issue regarding the issues that are there.


And it'll be easier to remove container from commands

PR #1469, opened 2022-07-17
Increase PHPStan to level 4

jordisala1991:
## Subject

I am targeting this branch, because this is for next major version.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,6 @@
+parameters:
+    ignoreErrors:
+        - # NEXT_MAJOR: Fix this one in SonataBlockBundle 4.x
+            message: "#^PHPDoc type Doctrine\\\\Common\\\\Collections\\\\Collection\\<\\(int\\|string\\), Sonata\\\\PageBundle\\\\Model\\\\PageBlockInterface\\> of property Sonata\\\\PageBundle\\\\Entity\\\\BaseBlock\\:\\:\\$children is not covariant with PHPDoc type array\\<Sonata\\\\BlockBundle\\\\Model\\\\BlockInterface\\> of overridden property Sonata\\\\BlockBundle\\\\Model\\\\BaseBlock\\:\\:\\$children\\.$#"
+            count: 1
+            path: src/Entity/BaseBlock.php
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,4 +1,5 @@
 includes:
+    - phpstan-baseline.neon
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
 
 parameters:
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,7 +2,7 @@ includes:
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
 
 parameters:
-    level: 3
+    level: 4
     symfony:
         console_application_loader: ./phpstan-console-application.php
     paths:
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,7 +2,7 @@ includes:
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
 
 parameters:
-    level: 2
+    level: 3
     symfony:
         console_application_loader: ./phpstan-console-application.php
     paths:
--- a/src/Admin/BlockAdmin.php
+++ b/src/Admin/BlockAdmin.php
@@ -88,7 +88,7 @@ final class BlockAdmin extends BaseBlockAdmin
 
     protected function configureFormFields(FormMapper $form): void
     {
-        $block = $this->getSubject();
+        $block = $this->hasSubject() ? $this->getSubject() : null;
 
         if (!$block) { // require to have a proper running test suite at the sandbox level
             return;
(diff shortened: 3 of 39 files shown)


eerison:
you do this (increase phpstan's level) quite fast 😄

jordisala1991:
The hardest one is level 6 T_T

phpstan-baseline.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+    ignoreErrors:
+        - # NEXT_MAJOR: Fix this one in SonataBlockBundle 4.x
```
VincentLanglet, line 3: Can we agree about how we plan to solve it ?

Changing array to iterable in blockBundle ?
Changing to arraycollection in blockbundle ? 
Using array here ?

VincentLanglet (changes requested):

VincentLanglet:
Should we start the strict rule now or later ?

phpstan-baseline.neon
jordisala1991: it can be solved with iterable , I tried it already, but cant do on 3.x (I didnt tried on 4.x tho)

jordisala1991:
> Should we start the strict rule now or later ?

Probably asap but on another Pr , it gives around 200 errors.

phpstan-baseline.neon
VincentLanglet: Currently 
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Model/BlockInterface.php#L139-L144
is typehinted to `array`.

So it changing the typehint will be mandatory to use SonataBlockBundle 4 indeed.

IMHO you can already add getChildren() and setChildren in the PageBlockInterface with phpdoc about Collection.

jordisala1991: setChildren is not a method on BlockInterface (and it is not used on for PageBlockInterface

for getChildren, is it needed? for the moment it does not break anything.

eerison:
Hey @jordisala1991 should we add a badge to phpstan? like

```markdown
[![PhpStan][phpstan_badge]][phpstan_link]
[phpstan_badge]: https://img.shields.io/badge/PHPStan-level%204-brightgreen.svg?style=flat
[phpstan_link]: https://github.com/sonata-project/SonataPageBundle/blob/4.x/phpstan.neon.dist#L5
```

VincentLanglet:
> Hey @jordisala1991 should we add a badge to phpstan

The badge might be interesting, but keep in mind that the README is generated in dev-kit.
https://github.com/sonata-project/dev-kit/blob/master/templates/project/README.md.twig#L11-L14

And you cannot know the phpstan level of a project from dev-kit :/

phpstan-baseline.neon
VincentLanglet: It would have help the dev which wanted to bump to BlockBundle v4 to know what to do about this method.

eerison:
> > Hey @jordisala1991 should we add a badge to phpstan
> 
> The badge might be interesting, but keep in mind that the README is generated in dev-kit. https://github.com/sonata-project/dev-kit/blob/master/templates/project/README.md.twig#L11-L14
> 
> And you cannot know the phpstan level of a project from dev-kit :/

true :'(

phpstan-baseline.neon
VincentLanglet: I think the best would be to do the PR on BlockBundle now.

Does this https://github.com/sonata-project/SonataBlockBundle/pull/1087 seems ok to you ?

Hanmac:
> > Hey @jordisala1991 should we add a badge to phpstan
> 
> The badge might be interesting, but keep in mind that the README is generated in dev-kit. https://github.com/sonata-project/dev-kit/blob/master/templates/project/README.md.twig#L11-L14
> 
> And you cannot know the phpstan level of a project from dev-kit :/

shouldn't there be a way to make it configurable in the config somewhere?
like in https://github.com/sonata-project/dev-kit/blob/master/config/projects.yaml ?

i mean `{{ project.title }}` or `{{ project.package.name }}` need to come from somewhere?

another Question about the PHPStan updates, would it be easier to make sequence updates from 2-3, 3-4, 4-5 instead of doing a big jump to 2-5 ?

VincentLanglet:
> shouldn't there be a way to make it configurable in the config somewhere? like in https://github.com/sonata-project/dev-kit/blob/master/config/projects.yaml ?

All the config come from this file. Sometime ago, we had `phpstan: true` to say add the check to the ci. But now every project has phpstan checks.
Adding a config `phpstan: 8` seems to much to me, just to have a badge. Because bumping the level would require a PR to dev-kit. But if every project is level 8, we might consider something since it won't change much. Also, it could be interesting to have the same phpstan.neon files. Looking at https://github.com/sonata-project/SonataAdminBundle/blob/4.x/phpstan.neon.dist, https://github.com/sonata-project/SonataBlockBundle/blob/4.x/phpstan.neon.dist, https://github.com/sonata-project/SonataMediaBundle/blob/4.x/phpstan.neon.dist and https://github.com/sonata-project/SonataClassificationBundle/blob/4.x/phpstan.neon.dist.

I want to try if I have time.

> i mean `{{ project.title }}` or `{{ project.package.name }}` need to come from somewhere?

From the github API.

> another Question about the PHPStan updates, would it be easier to make sequence updates from 2-3, 3-4, 4-5 instead of doing a big jump to 2-5 ?

here it's just 2 to 4. The 5 is in another PR, because he's waiting this one to be merged first.

phpstan-baseline.neon
jordisala1991: It seems ok (i have left a comment) but can't be tested with this PR because here we are still using block bundle 3

VincentLanglet (approved):

PR #1448, opened 2022-07-13
Upgrade Sonatablock and SonataSeo

Daric971:
## Subject

I am targeting this branch, because I updated to allow version 3 of sonataSeo 

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed
- Config flies (block and admin)
- Admin class
- BlockService classes
- FormType classes
- PageService classes
- Listeners
- mapper
- Tests

### Deprecated

### Removed

### Fixed

### Security
```

Issue #1416, opened 2022-04-22
Support symfony 5 and Sonata 4


eerison:
## Support `Symfony 5` and `Sonata 4`

It's a sub task from #1410

**Note:** this issue is blocked by 
- [x] #1096 
- [x] #1415 
- [x] #1420 
- [x] #1433 
- [x] #1475 
- [x] #1417 
- [x] #1477
- [x] #1442 
- [x] #1491
- [x] ~~#1454~~

jordisala1991:
The Symfony Runtime is not mandatory to enable support for Symfony 5.4 and 6.0. IMO it should stay at its own issue (and will be solved too for 4.0)

eerison:
> The Symfony Runtime is not mandatory to enable support for Symfony 5.4 and 6.0. IMO it should stay at its own issue (and will be solved too for 4.0)

Hmmm I updated one project from 4.4 to 5.4 and it force me to use runtime, do you have any application running without symfony runtime?

eerison:
> The Symfony Runtime is not mandatory to enable support for Symfony 5.4 and 6.0. IMO it should stay at its own issue (and will be solved too for 4.0)

Yeah I got what do you mean!

VincentLanglet:
I'll close this one in favor of the only missing check which is https://github.com/sonata-project/SonataPageBundle/issues/1410

PR #1471, opened 2022-07-18
Increase PHPStan to level 5 and Psalm to level 7

jordisala1991:
## Subject

I am targeting this branch, because this is for the next major.

### DIFF
--- a/src/CmsManager/CmsManagerInterface.php
+++ b/src/CmsManager/CmsManagerInterface.php
@@ -13,11 +13,9 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\CmsManager;
 
-use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\PageBundle\Model\PageBlockInterface;
 use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Model\SiteInterface;
-use Symfony\Component\HttpFoundation\Request;
 
 /**
  * The CmsManagerInterface class is in charge of retrieving the correct page (cms page or action page).
@@ -31,9 +29,9 @@ interface CmsManagerInterface
     /**
      * @param string $name
      *
-     * @return BlockInterface|bool|null
+     * @return PageBlockInterface|null
      */
-    public function findContainer($name, PageInterface $page, ?BlockInterface $parentContainer = null);
+    public function findContainer($name, PageInterface $page, ?PageBlockInterface $parentContainer = null);
 
     /**
      * Returns a fully loaded page ( + blocks ) from a url.
@@ -106,12 +104,12 @@ interface CmsManagerInterface
     /**
      * Returns the list of loaded block from the current http request.
      *
-     * @return BlockInterface[]
+     * @return PageBlockInterface[]
      */
     public function getBlocks();
 
     /**
-     * @param string $page
+     * @param int|string|null $page
      *
      * @return PageInterface
      */
--- a/src/Command/CleanupSnapshotsCommand.php
+++ b/src/Command/CleanupSnapshotsCommand.php
@@ -42,7 +42,7 @@ final class CleanupSnapshotsCommand extends Command
         $this->setDescription('Cleanups the deprecated snapshots by a given site');
 
         $this->addOption('site', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Site id');
-        $this->addOption('keep-snapshots', null, InputOption::VALUE_OPTIONAL, 'Keep a given count of snapshots per page', 5);
+        $this->addOption('keep-snapshots', null, InputOption::VALUE_REQUIRED, 'Keep a given count of snapshots per page', 5);
     }
 
     protected function initialize(InputInterface $input, OutputInterface $output): void
@@ -71,7 +71,7 @@ final class CleanupSnapshotsCommand extends Command
     }
 
     /**
-     * @param array<int> $ids
+     * @param array<string> $ids
      *
      * @return array<SiteInterface>
      */
--- a/src/Command/CreateSnapshotsCommand.php
+++ b/src/Command/CreateSnapshotsCommand.php
@@ -64,7 +64,7 @@ final class CreateSnapshotsCommand extends Command
     }
 
     /**
-     * @param array<int> $ids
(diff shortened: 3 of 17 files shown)


src/Page/PageServiceManager.php
```diff
@@ -109,11 +110,16 @@ public function execute(PageInterface $page, Request $request, array $parameters
      */
     private function createResponse(PageInterface $page)
     {
-        if ($page->getTarget()) {
-            $page->addHeader('Location', $this->router->generate($page->getTarget()));
```
jordisala1991: @eerison can you try if this code is used on any of your projects? `$page->getTarget()` returns a PageInterface, but the router generate expects a string...

eerison: Hey @jordisala1991 I checked and this target is defined only in snapshot entity: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Resources/config/doctrine/BaseSnapshot.orm.xml#L13

But I'm my project all data have this field null, then it explain why this never throw an exception because it never going into the "if" condition, and this method `createResponse` is used in every page.

should I create an issue to check this more in deep and see if we can remove this field?

eerison: if it's not used then no make sense to have this `createResponse` method

src/Command/CleanupSnapshotsCommand.php
```diff
@@ -71,7 +71,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     }
 
     /**
-     * @param array<int> $ids
+     * @param array<int, string> $ids
```
VincentLanglet: The key of this array doesn't matter in this method.

Did you meant array<int|string> ?

src/Command/CreateSnapshotsCommand.php
```diff
@@ -64,7 +64,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     }
 
     /**
-     * @param array<int> $ids
+     * @param array<int, string> $ids
```
VincentLanglet: Same

src/Command/UpdateCoreRoutesCommand.php
```diff
@@ -64,7 +64,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     }
 
     /**
-     * @param array<int> $ids
+     * @param array<int, string> $ids
```
VincentLanglet: Same

src/Page/PageServiceManager.php
eerison: if you want I can care about this :)

jordisala1991:
Before this PR we need: https://github.com/sonata-project/SonataPageBundle/pull/1448

src/Page/PageServiceManager.php
eerison: I just asked for some other guys check in their projects if there is some id in `target_id`, But I checked in other project used for several years and there isn't any row with `target_id`

eerison: #1481 OK I created this issue to investigate a bit more about this `target_id`, when the 4.x branch is able to use sonata 4, I'll stop to work on that ticket and back for the tickets that are necessary for #1416

eerison: Hey @jordisala1991 this method can be remove on `4.x` ;) 

https://github.com/sonata-project/SonataPageBundle/pull/1482/files#diff-46f2a9ad831e7f18ef69582b9678dd4b7b7574820ef3fc78481de7ac7884ca8bR116

eerison: and just the 

```
new Response('', 200, $page->getHeaders() ?: []);
```

directly instead of `createResponse`

jordisala1991: Fixed.

src/Command/CleanupSnapshotsCommand.php
jordisala1991: I mean: `array<string>`. It is an argument from the console, so this array is made of strings.

src/CmsManager/CmsManagerInterface.php
```diff
@@ -106,12 +104,12 @@ public function setCurrentPage(PageInterface $page);
...
      * Returns the list of loaded block from the current http request.
      *
-     * @return BlockInterface[]
+     * @return PageBlockInterface[]
      */
     public function getBlocks();
 
     /**
-     * @param string $page
+     * @param int|string|null $page
```
VincentLanglet, line 112: What is the `page` ? Do we mean pageId ?

Cause elsewhere it's typehinted `PageInterface $page`, so the name seems to be wrong here.

VincentLanglet (changes requested):

src/Command/CleanupSnapshotsCommand.php
eerison: > I mean: `array<string>`. It is an argument from the console, so this array is made of strings.

should we convert string to int and require `array<int>`?

src/CmsManager/CmsManagerInterface.php
jordisala1991: it could be multiple things, an id, a route or null if you want the current page for the site.

eerison: But not a PageInterface, in this case it should be `pageId` instead of page, isn't it?

VincentLanglet: So currently it's `pageIdOrPageUrl` ? That's why it's called `page` but it's really misleading with `PageInterface $page`.

Dunno if we could find a better name or if we should support PageInterface.

But it's not fully related to the PR so I won't block it

VincentLanglet (approved):

src/CmsManager/CmsManagerInterface.php
jordisala1991: I think there will be more PRs related to this, so we can change it later.

src/Command/CleanupSnapshotsCommand.php
jordisala1991: I don't know if that is a good idea. Let's discuss that as a separate issue. (The user defines the primary key, and it can be defined as string if he wants, so doing what you say it could be a problem).

PR #1473, opened 2022-07-19
Remove migrations command

### LINKED ISSUE #1465: Check which command are necessary
When I was working in this PR: #1460 , I realized that some commands maybe are not used because there is possible problems, But should be good understand the reason to have those commands and improve the code and coverage with tests.

Commands that must be checked

- #1473 [MigrateBlockNameSettingCommand](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/MigrateBlockNameSettingCommand.php)
- #1473 [MigrateToJsonTypeCommand](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/MigrateToJsonTypeCommand.php)
- [x] [CreateBlockContainerCommand](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Command/CreateBlockContainerCommand.php#L62
) (I won't touch in this command for now, because I'm not sure if it's useful or not, we can handle this in other ticket if it's necessary)

eerison:
## Remove migrations command

I am targeting this branch, because those commands are final and not used anymore, check here https://github.com/sonata-project/SonataPageBundle/issues/1465#issuecomment-1188781828 for more details

Closes #1465.

## Changelog

```markdown
### Removed
- Removed the command `MigrateBlockNameSettingCommand::class`
- Removed the command `MigrateToJsonTypeCommand::class`

```


jordisala1991 (dismissed):

jordisala1991:
Do we need an upgrade note here @VincentLanglet ?

VincentLanglet:
> Do we need an upgrade note here @VincentLanglet ?

I would say no but we need to deprecate this command on 3.x first ; or am I missing something ?

eerison:
> > Do we need an upgrade note here @VincentLanglet ?
> 
> I would say no but we need to deprecate this command on 3.x first ; or am I missing something ?

Even the class as final do we need to deprecate before to remove?

VincentLanglet:
> > > Do we need an upgrade note here @VincentLanglet ?
> 
> > 
> 
> > I would say no but we need to deprecate this command on 3.x first ; or am I missing something ?
> 
> 
> 
> Even the class as final do we need to deprecate before to remove?

People can still inject this class somewhere else.
And since this is Commands, people can execute them. So we should trigger a deprecation when someone run them.

In a general way, every removal need to be deprecated before ^^

eerison:
> > > > Do we need an upgrade note here @VincentLanglet ?
> > 
> > 
> > > 
> > 
> > 
> > > I would say no but we need to deprecate this command on 3.x first ; or am I missing something ?
> > 
> > 
> > Even the class as final do we need to deprecate before to remove?
> 
> People can still inject this class somewhere else. And since this is Commands, people can execute them. So we should trigger a deprecation when someone run them.
> 
> In a general way, every removal need to be deprecated before ^^

True! I didn't thought that, I'll deprecate first ;)

eerison:
I opened the PR :)

#1476

VincentLanglet:
Let's wait for the merge of https://github.com/sonata-project/SonataPageBundle/pull/1476 and 3.x into 4.x then

VincentLanglet:
Done, you can update @eerison

eerison:
> Done, you can update @eerison

Done :)

VincentLanglet (approved):

Issue #1442, opened 2022-07-10
Solve the mess with array and ArrayCollection in Model/Entities.


VincentLanglet:
There is multiple method which are typehinted with `Foo[]` but returns a Collection.

Issue 1: BlockInterface::children
https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Entity/BaseBlock.php#L33
and we can find a method to disable the lazy loading here: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Model/Block.php#L55
which are called here: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Entity/BlockInteractor.php#L150-L155

But the `getChildren()` method is coming from BlockBundle, which is typehinted to array in 4.x 
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Model/BlockInterface.php#L144

Issue 2: PageInterface::children
There is the same issue
https://github.com/sonata-project/SonataPageBundle/search?q=ArrayCollection
And I found a call to `array_unshift` on `$page->getParents()` in PageExtension.

Issue 3: Transformer::children
There is also something wrong with the Transformer:
We set an arrayCollection here as a value
https://github.com/sonata-project/SonataPageBundle/blob/9343568991c81a787e41dd3a02079227b7d2d7d9/src/Entity/Transformer.php#L232
but the property is considered as an array of BlockInterface

Issue 4: SnapshotChildrenCollection implementation
This class is doing multiple calls like `$this->collection->offsetUnset` when `collection` is an array. Which doesn't work. It seem like the collection property was meant to be a Collection, but it does 
```
$this->collection = $this->transformer->getChildren($this->page);
```
when TransformerInterface::getChildren() returns an array according to the phpdoc

Issue 5: SnapshotChildrenCollection usage
We're passing this class to Page::setChildren
https://github.com/sonata-project/SonataPageBundle/blob/9343568991c81a787e41dd3a02079227b7d2d7d9/src/Model/SnapshotPageProxy.php#L109-L115
which is supposed to be an array (or Collection cf issue 2). But it's not.

Do you use those code @eerison ?

This issue will be a blocker for the 4.0 version.

eerison:
Hey @VincentLanglet 

> Issue 2: PageInterface::children
There is the same issue

No, what I use from  PageInterface is `getId()` only this

> Issue 3: Transformer::children
There is also something wrong with the Transformer:
We set an arrayCollection here as a value

[...]

PR #1452, opened 2022-07-14
Increase PHPStan to level 2

jordisala1991:
## Subject

I am targeting this branch, because this is for the next major.

## Changelog

```markdown
### Added
- Add `PageManager::createWithDefaults()` to create Pages with some default values.
```

PR #1479, opened 2022-07-20
Create test for BlockInteractor::loadPageBlocks

eerison:
## Improve BlockInteractor

I am targeting this branch, because it's missing test in this branch.

Issue #1442 

## To do

- [x] Coverage the whole `Sonata\PageBundle\Entity\BlockInteractor::loadPageBlocks` method

### DIFF
--- a/tests/Entity/BlockInteractorTest.php
+++ b/tests/Entity/BlockInteractorTest.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Tests\Entity;
+
+use Doctrine\Persistence\ManagerRegistry;
+use PHPUnit\Framework\TestCase;
+use Sonata\PageBundle\Entity\BlockInteractor;
+use Sonata\PageBundle\Model\Block;
+use Sonata\PageBundle\Model\BlockManagerInterface;
+use Sonata\PageBundle\Model\Page;
+use Sonata\PageBundle\Model\PageInterface;
+
+final class BlockInteractorTest extends TestCase
+{
+    /**
+     * @testdox It is returning a block list.
+     */
+    public function testLoadPageBlocks(): void
+    {
+        //Mock
+        $managerRegistryMock = $this->createMock(ManagerRegistry::class);
+        $blockManagerInterfaceMock = $this->createMock(BlockManagerInterface::class);
+        //NEXT_MAJOR: use PageInterface
+        $pageMock = $this->createMock(Page::class);
+        //NEXT_MAJOR: use BlockInterface
+        $blockMock = $this->createMock(Block::class);
+
+        $blockInteractorMock = $this
+            ->getMockBuilder(BlockInteractor::class)
+            ->setConstructorArgs([$managerRegistryMock, $blockManagerInterfaceMock])
+            ->onlyMethods(['getBlocksById'])
+            ->getMock();
+
+        $blockInteractorMock
+            ->expects(static::once())
+            ->method('getBlocksById')
+            ->willReturn([$blockMock]);
+
+        //Run
+        $blocks = $blockInteractorMock->loadPageBlocks($pageMock);
+
+        //Asserts
+        static::assertSame([$blockMock], $blocks);
+    }
+
+    /**
+     * @testdox it'll return an empty array for blocks that are already loaded.
+     */
+    public function testNotLoadBlocks(): void
+    {
+        //Mock
+        $managerRegistryMock = $this->createMock(ManagerRegistry::class);
+        $blockManagerInterfaceMock = $this->createMock(BlockManagerInterface::class);
+        $pageMock = $this->createMock(PageInterface::class);
+        $pageMock
+            ->expects(static::once())
+            ->method('getId')
+            ->willReturn(1);
+
+        $blockInteractor = new BlockInteractor($managerRegistryMock, $blockManagerInterfaceMock);
+
+        //Change property visibility
+        $reflection = new \ReflectionClass($blockInteractor);
+        $reflection_property = $reflection->getProperty('pageBlocksLoaded');
+        $reflection_property->setAccessible(true);
+        $reflection_property->setValue($blockInteractor, [1 => 'fake_value(block already loaded).']);
+
+        //Run
+        $result = $blockInteractor->loadPageBlocks($pageMock);
+
+        //Assert
+        static::assertSame([], $result);
+    }
+
+    /**
+     * @testdox It's adding a new block children and "disableChildrenLazyLoading"
+     *
+     * I'm using "containerBlock" and "emailButtonBlock", just for the test be more clear,
+     *  because usually the container block is the parent of others blocks in the page.
+     */
+    public function testDisableAndAddChildrenBlocks(): void
+    {
+        //Mock
+        $managerRegistryMock = $this->createMock(ManagerRegistry::class);
+        $blockManagerInterfaceMock = $this->createMock(BlockManagerInterface::class);
+
+        //NEXT_MAJOR: use PageInterface
+        $pageMock = $this->createMock(Page::class);
+        $containerBlockMock = $this->createMock(Block::class);
+        $containerBlockMock
+            ->expects(static::exactly(2))//NEXT_MAJOR: change this to static::once()
+            ->method('getId')
+            ->willReturn(22);
+
+        $emailButtonMock = $this->createMock(Block::class);
+        $emailButtonMock
+            ->expects(static::exactly(3))//NEXT_MAJOR: change this to static::once()
+            ->method('getParent')
+            ->willReturn($containerBlockMock);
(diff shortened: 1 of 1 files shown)


VincentLanglet:
> Coverage the whole Sonata\PageBundle\Model\Page::disableBlockLazyLoading method
> Coverage the whole Sonata\PageBundle\Model\Block::disableChildrenLazyLoading method

Be aware that the disable*LazyLoading methods are already removed in 4.x in https://github.com/sonata-project/SonataPageBundle/pull/1452/files#diff-7cacd8fe71f848bb7884d9c6ffe4692b99c9a5dcda497c1643cf52f7668e09d0

eerison:
> > Coverage the whole Sonata\PageBundle\Model\Page::disableBlockLazyLoading method
> > Coverage the whole Sonata\PageBundle\Model\Block::disableChildrenLazyLoading method
> 
> Be aware that the disable*LazyLoading methods are already removed in 4.x in https://github.com/sonata-project/SonataPageBundle/pull/1452/files#diff-7cacd8fe71f848bb7884d9c6ffe4692b99c9a5dcda497c1643cf52f7668e09d0

hmmm But isn't it reflect in some unexpected behaviour? 

My idea is, coverage those method with test before change, and after that try to refactor to return array instead of `ArrayCollection`.

I guess it's necessary to solve this issue: #1442, isn't it?

Edit 1: I guess we need to understand what those methods do before remove, don't you?

eerison:
Other thing, I saw that `setInitialized` doesn't exist anywhere, then in theory this `loadPageBlocks`  is not used, because if it's used, it'll lead in an internal server error, because it'll execute a method that doesn't exist

VincentLanglet:
> hmmm But isn't it reflect in some unexpected behaviour?
> 
> My idea is, coverage those method with test before change, and after that try to refactor to return array instead of `ArrayCollection`.
> 
> I guess it's necessary to solve this issue: #1442, isn't it?
> 
> Edit 1: I guess we need to understand what those methods do before remove, don't you?

The changes already started on 4.x. I think most of the time we'll try to use Collection instead of array.
The issue is kinda almost solved but your idea is good.

We made changes in 4.x which were not covered by tests. Seems better to add test on 3.x branch, then to see if those tests are still working in 4.x branch (which means that changes we made are ok).

eerison:
> > hmmm But isn't it reflect in some unexpected behaviour?
> > My idea is, coverage those method with test before change, and after that try to refactor to return array instead of `ArrayCollection`.
> > I guess it's necessary to solve this issue: #1442, isn't it?
> > Edit 1: I guess we need to understand what those methods do before remove, don't you?
> 
> The changes already started on 4.x. I think most of the time we'll try to use Collection instead of array. The issue is kinda almost solved but your idea is good.
> 
> We made changes in 4.x which were not covered by tests. Seems better to add test on 3.x branch, then to see if those tests are still working in 4.x branch (which means that changes we made are ok).

well any test is covering `loadPageBlocks`

VincentLanglet:
> setInitialized

Not sure but I would say it exists on PersistentCollection or something like this.

> well any test is covering loadPageBlocks

So, having green tests on 4.x means that the changes made by jordisala91 about this Collection/array mess are ok, isnt it ?
Or do you mean that "We have to add a test on loadPageBlocks and it will be ok" ?

eerison:
> Not sure but I would say it exists on PersistentCollection or something like this.

That's true, I found it into the `vendor/doctrine/orm/lib/Doctrine/ORM/PersistentCollection.php`

> So, having green tests on 4.x means that the changes made by jordisala91 about this Collection/array mess are ok, isnt it ?

No, because this code there isn't any test covering this code, then if you change, remove or add something the current tests won't get this.

> Or do you mean that "We have to add a test on loadPageBlocks and it will be ok" ?

I guess it is not just add the test, I guess we need to know why it's added, or wait if someone appear with some problem and explain why do we need this 😄

eerison:
in theory it was added for some cache reason: https://github.com/sonata-project/SonataPageBundle/commit/03341017fd55afc5f84388cda65b9626b4c67159#diff-d5aead13ce2c4f649eb8a18cfab321e378aec0ea64ce306d416057b108ab5609R182

I guess in the worst case it'll lead some cache issue, should we care about this or let this like it's, and check later?

VincentLanglet:
> in theory it was added for some cache reason: [0334101#diff-d5aead13ce2c4f649eb8a18cfab321e378aec0ea64ce306d416057b108ab5609R182](https://github.com/sonata-project/SonataPageBundle/commit/03341017fd55afc5f84388cda65b9626b4c67159#diff-d5aead13ce2c4f649eb8a18cfab321e378aec0ea64ce306d416057b108ab5609R182)
> 
> I guess in the worst case it'll lead some cache issue, should we care about this or let this like it's, and check later?

Since we're removing the cache from this bundle, I'm not sure it will be an issue then.

I would say that we'll still have alpha version to check if it works

eerison:
Ok, should we merge this test?

VincentLanglet:
> Ok, should we merge this test?

Sure, every new test is welcomed ^^

eerison:
> > Ok, should we merge this test?
> 
> Sure, every new test is welcomed ^^

Done, it's ready to review :)

tests/Entity/BlockInteractorTest.php
```diff
@@ -0,0 +1,126 @@
...
+
+use Doctrine\Persistence\ManagerRegistry;
+use PHPUnit\Framework\TestCase;
+use Sonata\PageBundle\Entity\BlockInteractor;
+use Sonata\PageBundle\Model\Block;
+use Sonata\PageBundle\Model\BlockManagerInterface;
+use Sonata\PageBundle\Model\Page;
+use Sonata\PageBundle\Model\PageInterface;
+
+class BlockInteractorTest extends TestCase
```
VincentLanglet:
```suggestion
final class BlockInteractorTest extends TestCase
```

tests/Entity/BlockInteractorTest.php
```diff
@@ -0,0 +1,126 @@
...
+{
+    /**
+     * @testdox It is returning a block list.
+     */
+    public function testLoadPageBlocks(): void
+    {
+        //Mock
+        $managerRegistryMock = $this->createMock(ManagerRegistry::class);
+        $blockManagerInterfaceMock = $this->createMock(BlockManagerInterface::class);
+        //TODO it's not possible to pass the "PageInterface", because this method "disableBlockLazyLoading" does not exist there.
```
VincentLanglet: Prefer a NEXT_MAJOR comment, change to PageInterface.

tests/Entity/BlockInteractorTest.php
```diff
@@ -0,0 +1,126 @@
...
+        $pageMock = $this->createMock(Page::class);
+        //TODO it's not possible to pass the "BlockInterface". because this method "disableChildrenLazyLoading" does not exist there.
```
VincentLanglet: same

tests/Entity/BlockInteractorTest.php
```diff
@@ -0,0 +1,126 @@
...
+     * I'm using "containerBlock" and "emailButtonBlock", just for the test be more clear,
+     *  because usually the container block is the parent of others blocks in the page.
+     */
+    public function testDisableAndAddChildrenBlocks(): void
+    {
+        //Mock
+        $managerRegistryMock = $this->createMock(ManagerRegistry::class);
+        $blockManagerInterfaceMock = $this->createMock(BlockManagerInterface::class);
+
+        //TODO after solve the problem with interface, change the class to interface.
```
VincentLanglet: same

tests/Entity/BlockInteractorTest.php
```diff
@@ -0,0 +1,126 @@
...
+        $pageMock = $this->createMock(Page::class);
+        $containerBlockMock = $this->createMock(Block::class);
+        $containerBlockMock
+            ->expects(static::exactly(2))//TODO It should be changed to once
```
VincentLanglet: same

tests/Entity/BlockInteractorTest.php
```diff
@@ -0,0 +1,126 @@
...
+            ->method('getId')
+            ->willReturn(22);
+
+        $emailButtonMock = $this->createMock(Block::class);
+        $emailButtonMock
+            ->expects(static::exactly(3))//TODO It should be changed to once
```
VincentLanglet: same

VincentLanglet (changes requested):

tests/Entity/BlockInteractorTest.php
eerison: Yeah I did it 😄

tests/Entity/BlockInteractorTest.php
eerison: Done!

tests/Entity/BlockInteractorTest.php
eerison: Done

tests/Entity/BlockInteractorTest.php
eerison: Done

tests/Entity/BlockInteractorTest.php
eerison: Done

tests/Entity/BlockInteractorTest.php
eerison: Done

VincentLanglet (approved):

PR #1480, opened 2022-07-21
Fix build

jordisala1991:
Had to rewrite a test to use KernelTestCase...

We have a lot of test that mock too much things, we will need to take a look at that at some point (other priorities first).

### DIFF
--- a/tests/Entity/BlockInteractorTest.php
+++ b/tests/Entity/BlockInteractorTest.php
@@ -13,33 +13,50 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Tests\Entity;
 
-use Doctrine\Persistence\ManagerRegistry;
-use PHPUnit\Framework\TestCase;
+use Doctrine\ORM\EntityManagerInterface;
 use Sonata\PageBundle\Entity\BlockInteractor;
-use Sonata\PageBundle\Model\BlockManagerInterface;
-use Sonata\PageBundle\Model\PageBlockInterface;
 use Sonata\PageBundle\Model\PageInterface;
+use Sonata\PageBundle\Tests\App\AppKernel;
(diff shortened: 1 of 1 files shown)


tests/Entity/BlockInteractorTest.php
```diff
@@ -13,114 +13,85 @@
...
+        $this->blockInteractor = new BlockInteractor($managerRegistry, $blockManager);
+    }
+
     /**
      * @testdox It is returning a block list.
      */
     public function testLoadPageBlocks(): void
     {
-        //Mock
-        $managerRegistryMock = $this->createMock(ManagerRegistry::class);
```
eerison, line 32: Why don't you pass just a mock instead of the service? It should be just a unit test 👀 .

jordisala1991: Because we should not try to mock the orm. There are plenty of test doing it, and it is not a good practice.

There are some reasons for it:
1. Mocking the orm just ensure you are calling a bunch of functions, and not if the queries / all the logic works and returns expected results. 
2. You are forced to mock some internals of the ORM due to final classes. I do not want to mock internal logic of classes because that is very fragile and it will break as soon as the internal class changes without BC break (and again it goes to the point 1 too).
3. You should avoid to mock what you don't own, there are a lot of articles about that.
4. The test is more complex to read with that mocking.

About doing a kernel test case on the Entity Folder, I do not see any problem about that, we usually don't split test by unit or integration folders. What we do is add a Functional folder for those test that do full application requests and check the generated html.

VincentLanglet (approved):

tests/Entity/BlockInteractorTest.php
eerison: Ok I got it! :)

Issue #1481, opened 2022-07-21
Remove target_id from page and snapshot in case it's not used.


eerison:
## Remove target_id from page and snapshot

in this PR: https://github.com/sonata-project/SonataPageBundle/pull/1471#discussion_r926304297 we realised that "maybe", target_id is not used, and to try to keep the code simple, we should see where it's used and remove in case it's not really necessary.

PR #1482, opened 2022-07-21
Remove target page code

eerison:
## Remove old target page code

I am targeting this branch, because the target page need to be deprecated in 3.x first.

Issue #1481 .

## Changelog

```markdown
### Deprecated
- Deprecated target page code from `src/Command/CloneSiteCommand.php`
- Deprecated target page code from `src/Entity/Transformer.php`
- Deprecated `src/Model/Page::target` property
- Deprecated `src/Model/Page::getTarget()` method
- Deprecated `src/Model/Page::setTarget()` method
- Deprecated `src/Model/PageInterface::getTarget()` method
- Deprecated `src/Model/PageInterface::setTarget()` method
- Deprecated `src/Model/Snapshot::target` property
- Deprecated `src/Model/Snapshot::targetId` property
- Deprecated `src/Model/Snapshot::getTarget()` method
- Deprecated `src/Model/Snapshot::setTarget()` method
- Deprecated `src/Model/Snapshot::getTargetId()` method
- Deprecated `src/Model/Snapshot::setTargetId()` method
- Deprecated `src/Model/SnapshotPageProxy::target` property
- Deprecated `src/Model/SnapshotPageProxy::getTargetId()` method
- Deprecated `src/Model/SnapshotPageProxy::setTargetId()` method
- Deprecated `src/Page/PageServiceManager::createResponse()` method
```

### DIFF
--- a/src/Command/CloneSiteCommand.php
+++ b/src/Command/CloneSiteCommand.php
@@ -136,7 +136,14 @@ final class CloneSiteCommand extends BaseCommand
                 }
             }
 
+            //NEXT_MAJOR: Remove this target condition block.
             if ($page->getTarget()) {
+                @trigger_error(
+                    'target page is deprecate since sonata-project/page-bundle 3.27.0'.
+                    ', and it will be removed in 4.0',
+                    \E_USER_DEPRECATED
+                );
+
                 if (\array_key_exists($page->getTarget()->getId(), $pageClones)) {
                     $output->writeln(
                         sprintf(
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -89,7 +89,13 @@ class Transformer implements TransformerInterface
             $snapshot->setParentId($page->getParent()->getId());
         }
 
+        //NEXT_MAJOR: Remove this "if" condition block.
         if ($page->getTarget()) {
+            @trigger_error(
+                'target page is deprecate since sonata-project/page-bundle 3.27.0'.
+                ', and it will be removed in 4.0',
+                \E_USER_DEPRECATED
+            );
             $snapshot->setTargetId($page->getTarget()->getId());
         }
 
@@ -108,7 +114,7 @@ class Transformer implements TransformerInterface
         $content['updated_at'] = $page->getUpdatedAt()->format('U');
         $content['slug'] = $page->getSlug();
         $content['parent_id'] = $page->getParent() ? $page->getParent()->getId() : null;
-        $content['target_id'] = $page->getTarget() ? $page->getTarget()->getId() : null;
+        $content['target_id'] = $page->getTarget() ? $page->getTarget()->getId() : null; //NEXT_MAJOR: Remove this line.
 
         $content['blocks'] = [];
         foreach ($page->getBlocks() as $block) {
--- a/src/Model/Page.php
+++ b/src/Model/Page.php
@@ -136,7 +136,11 @@ abstract class Page implements PageInterface
     protected $parents;
 
     /**
-     * @var PageInterface|null
+     * @var pageInterface|null
+     *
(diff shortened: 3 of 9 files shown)


eerison:
It was the places that I found where targetId is set, But it'll be executed only if exist `target` page

and I didn't find any where to set this `target_id`

even in the admin 
Screenshot 2022-07-21 at 13 29 16

then my thought  is, it's not used 👀

eerison:
@VincentLanglet and @jordisala1991 I know you didn't know the project, But in every place that uses this `target_id`, to be set, it depends that target already exist, But as never it will be set, then it'll never it'll be used!

the question is, should I deprecate in `3.x` and remove in `4.x` what do you think?

Note:

I checked in 2 projects that uses sonata page, and any of them uses this `target_id`

checked


```
bin/console doctrine:query:sql "select target_id, content from page__snapshot where target_id is not null or content not like '%target_id\":null%'"

bin/console doctrine:query:sql "select target_id, content from page__snapshot where target_id is not null"
```

both queries returned empty.

Edit 1: I want to confirm and know your opinion before invest time on that!

jordisala1991:
Some of that logic was added 10 years ago, here: https://github.com/sonata-project/SonataPageBundle/commit/fc9cd601227e0694e169bcfc365e1a57f0e1a3d2 but seems incomplete...

eerison:
> Some of that logic was added 10 years ago, here: [fc9cd60](https://github.com/sonata-project/SonataPageBundle/commit/fc9cd601227e0694e169bcfc365e1a57f0e1a3d2) but seems incomplete...

well if it's not used, i would remove it, it will help us to have less code to maintain

eerison:
Ok I got a new update, there is a really old project that uses `~3.7` version of sonata page, and in that project this target_id is used, I don't know if it was migrated from another old version like `2.x`, But there are some pages with target_id.

Edit 1: probably this comes from 2.x version look: https://github.com/sonata-project/SonataPageBundle/blob/2.x/Admin/PageAdmin.php#L208, in the 2.x version was possible to add `target` page, but it was removed on 3.x

Edit 2: I guess [here](https://github.com/sonata-project/SonataPageBundle/commit/0716a119d7545905bda39aeaf86514a19f5964eb#diff-e04bbf95fc49db300f66b3557d1ed1956594c892128b59492376c5287a18b6e5R414) was the mistake/feature 😄 , it was changed from `target` to `parent`, and there are 2 `parents` fields in the code.

Edit 3: It was introduced in `3.5.1` and in `3.6.0` branch the file was removed 👀

Edit 4: I colected some information and this feature was used for 

```
exactly that was the selling point to this field if you have the same content but on different urls just create it once then use this field for the second one.
```

eerison:
the feature was just removed, then I guess it's fine we remove the code related with this, don't you?
if someone need this feature again, he can provider a PR for this, what do you think?

jordisala1991:
So the feature was to kind of duplicate pages with same content? IMO if the feature was kind of removed and there are sone leftovers I would completely remove it, and if someone comes up with some need for it, we can always reimplement.
TBH this might be broken for a lot of years, we will not miss this code.

Lets deprecate all related to target id on 3.x and remove on 4.x. Can you do it?

eerison:
> So the feature was to kind of duplicate pages with same content? IMO if the feature was kind of removed and there are sone leftovers I would completely remove it, and if someone comes up with some need for it, we can always reimplement. TBH this might be broken for a lot of years, we will not miss this code.
> 
> Lets deprecate all related to target id on 3.x and remove on 4.x. Can you do it?

Yeah, let's do that way ;)

VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thanks @eerison

PR #1485, opened 2022-07-22
Upgrade to Sonata 4

jordisala1991:
## Subject

I am targeting this branch, because this includes a lot of BC breaks.

Part of #1416 #1417

## Changelog

```markdown
### Added
- Add support for Sonata ^4.
```


eerison:
I imagined that you could fix this faster then me 😄 

But for your pipeline pass you need to use `4.x as 4.14.0` or create a release in sonata admin

eerison:
because of this PR: https://github.com/sonata-project/SonataAdminBundle/pull/7867

jordisala1991:
I know, but I wanted to start the Pr to see how it looked.

src/Admin/PageAdmin.php
```diff
@@ -364,10 +366,6 @@ protected function configureFormFields(FormMapper $form): void
                 ->add('stylesheet', null, ['required' => false])
                 ->add('rawHeaders', null, ['required' => false])
             ->end();
-
-        $form->setHelps([
-            'name' => 'help_page_name',
```
jordisala1991, line 363: Maybe this needs to be placed elsewhere?

src/Resources/config/block.xml
```diff
@@ -46,5 +46,6 @@
             <argument type="service" id="twig"/>
             <argument type="service" id="sonata.page.manager.page"/>
         </service>
+        <service id="Sonata\BlockBundle\Block\BlockServiceManagerInterface" alias="sonata.block.manager"/>
```
jordisala1991, line 49: I think this alias should be something like "sonata.page.manager.block" to not collide with something declared on blockbundle.

src/Resources/config/orm.xml
```diff
@@ -34,10 +36,12 @@
             <argument>%sonata.page.site.class%</argument>
             <argument type="service" id="doctrine"/>
         </service>
+        <service id="Sonata\PageBundle\Model\SiteManagerInterface" alias="sonata.page.manager.site"/>
         <service id="sonata.page.block_interactor" class="%sonata.page.block_interactor.class%" public="true">
             <argument type="service" id="doctrine"/>
             <argument type="service" id="sonata.page.manager.block"/>
         </service>
+        <service id="Sonata\PageBundle\Model\BlockInteractorInterface" alias="sonata.page.block_interactor"/>
```
jordisala1991, line 43: What do you think of aliasing model interfaces with the orm implementation? we can do this way or bind directly the implementation.

src/Admin/SharedBlockAdmin.php
```diff
@@ -34,14 +34,14 @@ final class SharedBlockAdmin extends BaseBlockAdmin
      */
     protected $classnameLabel = 'shared_block';
 
-    public function getBaseRoutePattern()
+    protected function generateBaseRoutePattern(bool $isChildAdmin = false): string
     {
-        return sprintf('%s/%s', parent::getBaseRoutePattern(), 'shared');
+        return sprintf('%s/%s', parent::generateBaseRoutePattern(), 'shared');
```
VincentLanglet:
```suggestion
        return sprintf('%s/%s', parent::generateBaseRoutePattern($isChildAdmin), 'shared');
```

src/Admin/SharedBlockAdmin.php
```diff
@@ -34,14 +34,14 @@ final class SharedBlockAdmin extends BaseBlockAdmin
...
     }
 
-    public function getBaseRouteName()
+    protected function generateBaseRouteName(bool $isChildAdmin = false): string
     {
-        return sprintf('%s/%s', parent::getBaseRouteName(), 'shared');
+        return sprintf('%s/%s', parent::generateBaseRouteName(), 'shared');
```
VincentLanglet:
```suggestion
        return sprintf('%s/%s', parent::generateBaseRouteName($isChildAdmin), 'shared');
```

VincentLanglet (changes requested):

VincentLanglet:
We will need you @eerison to both try the 3.x branch (in order to release all the deprecation) and to try this branch (with Sonata 4) to confirm that everything work as expected.

For the 4.x branch, it can still be done later or when the alpha will be release. But the 3.x branch should be tested because it will be a pretty huge release.

eerison:
> We will need you @eerison to both try the 3.x branch (in order to release all the deprecation) and to try this branch (with Sonata 4) to confirm that everything work as expected.
> 
> For the 4.x branch, it can still be done later or when the alpha will be release. But the 3.x branch should be tested because it will be a pretty huge release.

Ok tomorrow on the morning I'll test 3.x release and I will confirm here :)

src/Admin/SharedBlockAdmin.php
jordisala1991: Fixed.

src/Admin/SharedBlockAdmin.php
jordisala1991: Fixed.

src/Admin/BaseBlockAdmin.php
```diff
@@ -50,102 +48,27 @@ abstract class BaseBlockAdmin extends AbstractAdmin
...
-            $block->postPersist($object);
-
-            @trigger_error(
-                'The '.__METHOD__.'() method is deprecated since sonata-project/block-bundle 3.12.0 and will be removed in version 4.0.',
-                \E_USER_DEPRECATED
-            );
-        }
-    }
-
     public function preRemove($object): void
```
jordisala1991: This should probably be typed

VincentLanglet (dismissed):

src/Controller/BlockAdminController.php
```diff
@@ -74,33 +82,30 @@ public function savePositionAction(?Request $request = null)
...
 
-    public function createAction(?Request $request = null)
+    public function createAction(Request $request): Response
     {
         $this->admin->checkAccess('create');
 
-        $sharedBlockAdminClass = $this->container->getParameter('sonata.page.admin.shared_block.class');
-        if (!$this->admin->getParent() && \get_class($this->admin) !== $sharedBlockAdminClass) {
+        $sharedBlockAdminClass = $this->getParameter('sonata.page.admin.shared_block.class');
+        if ($this->admin->isChild() && \get_class($this->admin) !== $sharedBlockAdminClass) {
```
jordisala1991: this is wrong

jordisala1991:
```suggestion
        if (!$this->admin->isChild() && \get_class($this->admin) !== $sharedBlockAdminClass) {
```

VincentLanglet (dismissed):

src/Admin/PageAdmin.php
jordisala1991: Fixed.

src/Resources/config/orm.xml
jordisala1991: @VincentLanglet . If we consider we won't implement mongodb integration anytime soon (I won't try it). This might be a good move, wdyt?

src/Resources/config/block.xml
jordisala1991: I did an alias to: "sonata.page.block.service_manager". Other option is to create the alias inside blockbundle, but that's not how we did it in other cases like mediabundle aliasing things of classification bundle, wdyt @VincentLanglet ?

(we can change it later anyway)

jordisala1991: I can't change this alias, should we define it inside blockbundle @VincentLanglet ?

The problem with that it is creating kind of circular dependency, if someone removes that alias there by mistake, this bundle will stop working.

IMHO is not a big deal and can stay here the alias.

src/Admin/BaseBlockAdmin.php
jordisala1991: Fixed.

VincentLanglet (approved):

src/Resources/config/orm.xml
VincentLanglet: How do we do in other bundle, like MediaBundle which supports both ORM and Mongo ?
IMHO, we should be consistent

PR #1486, opened 2022-07-23
Merge 3.x into 4.x

VincentLanglet:


VincentLanglet:
SnapshotManager config is
```
<service id="sonata.page.manager.snapshot" class="%sonata.page.manager.snapshot.class%" public="true">
            <argument>%sonata.page.snapshot.class%</argument>
            <argument type="service" id="doctrine"/>
            <argument type="service" id="sonata.page.proxy.snapshot.factory"/>
            <argument/>
        </service>
```
Seems like the we're injecting `''` instead of an array.

Any help is welcomed @jordisala1991 @eerison @Hanmac

eerison:
> %sonata.page.manager.snapshot.class%

I'm not sure, why do we need this `%sonata.page.manager.snapshot.class%`, I guess the idea is inject other class?

But it's the idea I guess we can do this using decorator?

VincentLanglet:
My main concern was about the fourth argument of the constructor.

But seems like the SnapshotManager templates are
- not used
- not injected in the config
- not in the interface

So I removed them

tests/Command/CloneSiteCommandTest.php
```diff
@@ -63,6 +60,9 @@ protected function setUp(): void
         $this->application->add($command);
     }
 
+    /**
+     * @group legacy
```
jordisala1991, line 64: why group legacy?

jordisala1991 (approved):
Only a pedantic question

tests/Command/CloneSiteCommandTest.php
jordisala1991: Answer myself to the question, because this is needed: https://github.com/sonata-project/SonataPageBundle/pull/1488
We should remember to remove this on that PR.

PR #1493, opened 2022-07-26
Increase PHPStan to level 6

jordisala1991:
## Subject

I am targeting this branch, because this is for next major.

## Changelog

```markdown
### Changed
- Change `SnapshotChildrenCollection` to be a lazy `Collection` from Doctrine
```


src/Model/Template.php
```diff
@@ -55,18 +76,20 @@ public function getContainers()
...
     }
 
     /**
-     * @param $code
+     * @param string $code
+     *
+     * @return array<string, mixed>
      *
-     * @return array
+     * @phpstan-return Container|array{}
```
VincentLanglet: Seems weird to return `array{}` if no container was found.

It should be a LogicException (or at least null).

jordisala1991:
```suggestion
     * @phpstan-return Container|null
```

src/Model/Template.php
```diff
@@ -55,18 +76,20 @@ public function getContainers()
...
     public function addContainer($code, $meta): void
     {
         $this->containers[$code] = $this->normalize($meta);
     }
 
     /**
-     * @param $code
+     * @param string $code
+     *
+     * @return array<string, mixed>
```
jordisala1991:
```suggestion
     * @return array<string, mixed>|null
```

src/Model/Template.php
jordisala1991: I went for null instead of exceptions

VincentLanglet (approved):

src/Model/Template.php
VincentLanglet: It makes sens since there is no has method

PR #1506, opened 2022-07-30
Remove Slugify Method

jordisala1991:
## Subject

## Changelog

```markdown
### Removed
- Remove `setSlugifyMethod` and `getSlugifyMethod` static methods from Page.
```


jordisala1991:
TBH I don't even know why is this needed. Since there is only that call to the setter, and the getter is never used (AFAIK). @eerison do you have another opinion?

VincentLanglet (commented):
Phpstan is failing.
You need to update `public static function slugify($text)`

jordisala1991:
Yep, now I saw the usage... I will refactor code to not need any of this black magic

eerison:
Hey @jordisala1991 I didn't check how it's done

But I know that we use the slugfy for pages, but if I remember correct we can't set, it's based on the page name 🤔

src/DependencyInjection/Configuration.php
```diff
@@ -112,9 +112,6 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ])
                 ->prototype('scalar')->end()
             ->end()
-            ->scalarNode('slugify_service')
```
jordisala1991, line 115: To me, there is no point on doing this kind of thing, wdyt @VincentLanglet ? Symfony already provides decoration to change services. And since the slugifier is only used on the page manager, you can also override that definition.

eerison: Yeah i think the same, and I guess it's the same reason that we have service class as parameter

Like

`%snapshot.manager.class%`

VincentLanglet: I agree ; but if we want to do this the good way, we have to deprecate it on 3.x

jordisala1991: https://github.com/sonata-project/SonataPageBundle/pull/1511

jordisala1991: So, this one is done.

VincentLanglet (approved):
We need a PR on 3.x with a BC way to use the new constructor signature of PageManager

jordisala1991:
I need some help for all those 3.x BC things... otherwise it will take ages to get there. In this case, the manager will be final on 4.x. Do we need to provide such a BC path?

VincentLanglet:
> I need some help for all those 3.x BC things... otherwise it will take ages to get there. In this case, the manager will be final on 4.x. Do we need to provide such a BC path?

If someone call
```
new PageManager($class, $registry, $defaults, $pageDefaults);
```
on 4.x it will crash

So we need to do
```
public function __construct($class, ManagerRegistry $registry, Slugify|array $slugify = [], array $default = [], array $pageDefaults = []) 
```
on 3.x.
is_array($slugify) => deprecation message

jordisala1991:
> > I need some help for all those 3.x BC things... otherwise it will take ages to get there. In this case, the manager will be final on 4.x. Do we need to provide such a BC path?
> 
> If someone call
> 
> ```
> new PageManager($class, $registry, $defaults, $pageDefaults);
> ```
> 
> on 4.x it will crash
> 
> So we need to do
> 
> ```
> public function __construct($class, ManagerRegistry $registry, Slugify|array $slugify = [], array $default = [], array $pageDefaults = []) 
> ```
> 
> on 3.x. is_array($slugify) => deprecation message

Is not that easy, I will need to inject the variable, and try to use it. But we don't know how the user is calling the setSlugifyMethod on the page model. So at the end, all of it could be a bc break.

VincentLanglet:
> Is not that easy, I will need to inject the variable, and try to use it. But we don't know how the user is calling the setSlugifyMethod on the page model. So at the end, all of it could be a bc break.

If it's too hard, we can accept the non-solvabel BC break to release 4.x faster then

PR #1515, opened 2022-08-03
Create test for templateManager

### LINKED ISSUE #1497: Breadcrumbs are not appearing into the pages
In 4.x branch the homepage looks strange.

it's the page for 3.x branch
Screenshot 2022-07-26 at 16 46 10

it's the same page for 4.x branch

Screenshot 2022-07-26 at 16 58 28

eerison:
## Cover templateManager

I am targeting this branch, because in 4.x the page content is not rendered, and I need to understand first how it was working, something was changed or removed in 4.x that I can't identify.

Issue #1497

### DIFF
--- a/src/Resources/views/layout.html.twig
+++ b/src/Resources/views/layout.html.twig
@@ -13,12 +13,12 @@ file that was distributed with this source code.
 {% block sonata_page_container %}
     <div class="container">
         <div class="content">
-            <div class="row page-header" id="header">
+            <div class="row page-header">
                 {{ sonata_page_render_container('header', 'global') }}
             </div>
 
             {% block sonata_page_breadcrumb %}
-                <div class="row" id="breadcrumb">
+                <div class="row page-breadcrumb">
                     {{ sonata_block_render_event('breadcrumb', { 'context': 'page', 'current_uri': app.request.requestUri }) }}
                 </div>
             {% endblock %}
@@ -32,7 +32,7 @@ file that was distributed with this source code.
                 </div>
             {% endif %}
 
-            <div class="row" id="content">
+            <div class="row">
                 {% block page_content %}
                     {% if content is defined %}
                         {{ content|raw }}
@@ -58,7 +58,7 @@ file that was distributed with this source code.
             {% endif %}
         </div>
 
-        <footer class="row" id="footer">
+        <footer class="row">
             {{ sonata_page_render_container('footer', 'global') }}
         </footer>
     </div>
--- a/src/Resources/views/layout.html.twig
+++ b/src/Resources/views/layout.html.twig
@@ -13,12 +13,12 @@ file that was distributed with this source code.
 {% block sonata_page_container %}
     <div class="container">
         <div class="content">
-            <div class="row page-header">
+            <div class="row page-header" id="header">
                 {{ sonata_page_render_container('header', 'global') }}
             </div>
 
             {% block sonata_page_breadcrumb %}
-                <div class="row">
+                <div class="row" id="breadcrumb">
                     {{ sonata_block_render_event('breadcrumb', { 'context': 'page', 'current_uri': app.request.requestUri }) }}
                 </div>
             {% endblock %}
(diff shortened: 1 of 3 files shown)


eerison:
I'm trying to mock the `sonata.templating` service, but I looks like removed

```
root@4d1918ef933d:/app# bin/console debug:container sonata.templating

Information for Service "sonata.templating"
===========================================

 ---------------- ------------------------------------------ 
  Option           Value                                     
 ---------------- ------------------------------------------ 
  Service ID       sonata.templating                         
  Class            Sonata\BlockBundle\Templating\TwigEngine  
  Tags             -                                         
  Public           no                                        
  Synthetic        no                                        
  Lazy             no                                        
  Shared           yes                                       
  Abstract         no                                        
  Autowired        no                                        
  Autoconfigured   no                                        
 ---------------- ------------------------------------------ 


 ! [NOTE] The "sonata.templating" service or alias has been removed or inlined when the container was compiled. 
```

do you have any idea why @VincentLanglet or @Daric971

Edit 1: and It leads to other question, is this `make lint-symfony-container` working? because I tried to run this locally passing a service that doesn't exist and it doesn't broken

VincentLanglet:
I know nothing about this service :/

eerison:
Finally I could cover `breadcrumb` behaviour in test, I hope it helps me to fix in 4.x.

eerison:
Could you check this PR @VincentLanglet @jordisala1991 , please :)

src/Resources/views/layout.html.twig
```diff
@@ -13,12 +13,12 @@ file that was distributed with this source code.
 {% block sonata_page_container %}
     <div class="container">
         <div class="content">
-            <div class="row page-header">
```
VincentLanglet: I don't think we put id often ; so I dont like adding some for test, and would prefer something else than
```
static::assertCount(1, $crawler->filter('#header'));
```

Not sure what's sonata_page_render_container is doing but you might find somehting to search

VincentLanglet (changes requested):

src/Resources/views/layout.html.twig
eerison: But in this case I'm testing the page, and I'm checking if there is an element into the page!
I guess it is fine to test elements into the page using "id", don't you?

VincentLanglet: breadcrumb, header, content, ... are way too generic words. What if someone use js on it ? What if he's already is using those ?

Adding id, it just a risk of BC-break with no benefit for the code. It's not like we're using js on something.
You're adding them because you're writting tests. And you should never have your test pushing you to modify your src code.

You can work on the already existing classes. Adding `page-breadcrumb` class and so on might be a feature to allow overriding the design for user ; it seems better than ids.

But I don't understand the purpose of your test, checking for div in the page ; Checking for url, 200, etc might be enough.

@jordisala1991 wrote a lot of tests on 4.x branch and never needed to add id, or to find some div.

eerison: What I want to make sure is if the `breadcrumbs` are rendered into the page!

eerison: > page-breadcrumb

Ok I'll try to use `page-breadcrumb`

eerison: Hey @VincentLanglet could you check again please!

I just let the testing checking `breadcrumb` only, and I'm following the same way from `page-header`

VincentLanglet (approved):

eerison:
Hey @jordisala1991 could you review this PR please. and if it's possible merge into 4.x, tomorrow I would try to investigate the problem related with #1497

jordisala1991 (approved):
IMO this test would be much more easy on 4.x with the tests i have already added for render pages.

VincentLanglet:
Can you open a PR of the merge of 3.x into 4.x @eerison ? There is a lot of issue with this test when I try to merge

eerison:
Yeah, tomorrow I can do this :)

PR #1501, opened 2022-07-26
Fix transformer for blocks without name or position

jordisala1991:
## Subject

I am targeting this branch, because this only happens on 4.x branch.

Closes #1496

PR #1516, opened 2022-08-03
Transformer test

Hanmac:
## Subject

Some Tests for Transformer class using mocks.
Currently tested: `create`, `load` and `loadBlock`

Should i do a Mock case for the `getChildren` method too trying to Mock the DB Query, or should i later do a Functional Test against a Test Database?

Also the question, what exactly is the SnapshotPageProxy class for?

I am targeting this branch, because I don't know if that counts as BC or not.

### DIFF
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -150,9 +150,7 @@ class Transformer implements TransformerInterface
         $page->setJavascript($content['javascript']);
         $page->setStylesheet($content['stylesheet']);
         $page->setRawHeaders($content['raw_headers']);
-        if (isset($content['title'])) {
-            $page->setTitle($content['title']);
-        }
+        $page->setTitle($content['title'] ?? null);
         $page->setMetaDescription($content['meta_description']);
         $page->setMetaKeyword($content['meta_keyword']);
         $page->setName($content['name']);
@@ -179,17 +177,13 @@ class Transformer implements TransformerInterface
 
         $block->setPage($page);
         $block->setId($content['id']);
-        if (isset($content['name'])) {
-            $block->setName($content['name']);
-        }
+        $block->setName($content['name'] ?? null);
         $block->setEnabled($content['enabled']);
         if (isset($content['position'])) {
             $block->setPosition($content['position']);
         }
         $block->setSettings($content['settings']);
-        if (isset($content['type'])) {
-            $block->setType($content['type']);
-        }
+        $block->setType($content['type'] ?? null);
 
         $createdAt = new \DateTime();
         $createdAt->setTimestamp((int) $content['created_at']);
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -144,13 +144,15 @@ class Transformer implements TransformerInterface
         $page->setSite($snapshot->getSite());
         $page->setEnabled($snapshot->getEnabled());
 
-        $content = $this->fixPageContent($snapshot->getContent());
+        $content = $snapshot->getContent();
 
         $page->setId($content['id']);
         $page->setJavascript($content['javascript']);
         $page->setStylesheet($content['stylesheet']);
         $page->setRawHeaders($content['raw_headers']);
-        $page->setTitle($content['title']);
+        if (isset($content['title'])) {
+            $page->setTitle($content['title']);
+        }
         $page->setMetaDescription($content['meta_description']);
         $page->setMetaKeyword($content['meta_keyword']);
         $page->setName($content['name']);
@@ -173,15 +175,19 @@ class Transformer implements TransformerInterface
     {
         $block = $this->blockManager->create();
 
-        $content = $this->fixBlockContent($content);
-
         $block->setPage($page);
         $block->setId($content['id']);
-        $block->setName($content['name']);
+        if (isset($content['name'])) {
+            $block->setName($content['name']);
+        }
         $block->setEnabled($content['enabled']);
-        $block->setPosition($content['position']);
+        if (isset($content['position'])) {
+            $block->setPosition($content['position']);
+        }
         $block->setSettings($content['settings']);
-        $block->setType($content['type']);
+        if (isset($content['type'])) {
+            $block->setType($content['type']);
+        }
 
         $createdAt = new \DateTime();
         $createdAt->setTimestamp((int) $content['created_at']);
@@ -242,31 +248,9 @@ class Transformer implements TransformerInterface
     }
 
     /**
-     * @return array
-     */
-    protected function fixPageContent(array $content)
-    {
-        if (!\array_key_exists('title', $content)) {
-            $content['title'] = null;
-        }
-
-        return $content;
-    }
-
-    /**
-     * @return array
-     */
-    protected function fixBlockContent(array $content)
-    {
-        if (!\array_key_exists('name', $content)) {
-            $content['name'] = null;
-        }
-
-        return $content;
-    }
-
-    /**
-     * @return array
+     * @return array<string, mixed>
+     *
+     * @phpstan-return BlockContent
      */
     protected function createBlocks(BlockInterface $block)
     {
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -147,7 +147,7 @@ class Transformer implements TransformerInterface
         $page->setSite($snapshot->getSite());
         $page->setEnabled($snapshot->getEnabled());
 
-        $content = $snapshot->getContent();
+        $content = $this->fixPageContent($snapshot->getContent());
 
         $page->setId($content['id']);
         $page->setJavascript($content['javascript']);
@@ -177,6 +177,8 @@ class Transformer implements TransformerInterface
     public function loadBlock(array $content, PageInterface $page)
     {
         $block = $this->blockManager->create();
+
+        $content = $this->fixBlockContent($content);
 
         $block->setPage($page);
         $block->setId($content['id']);
@@ -248,6 +250,34 @@ class Transformer implements TransformerInterface
         }
 
         return $this->children[$page->getId()];
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this.
+     *
+     * @phpstan-return PageContent
+     */
+    protected function fixPageContent(array $content): array
+    {
+        if (!\array_key_exists('title', $content)) {
+            $content['title'] = null;
+        }
+
+        return $content;
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this.
+     *
+     * @phpstan-return BlockContent
+     */
+    protected function fixBlockContent(array $content): array
+    {
+        if (!\array_key_exists('name', $content)) {
+            $content['name'] = null;
+        }
+
+        return $content;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Is it ready to be reviewed ?

Hanmac:
> Is it ready to be reviewed ?

From my point of view yes

or should i try to rebase that for 3.x ?

VincentLanglet:
> or should i try to rebase that for 3.x ?

Seems BC so yes, if it's not too hard would be great to have them on 3.x

Hanmac:
it seems the tests are waiting for something ;p

Hanmac:
@VincentLanglet maybe the problem is the web interface but I can't see why the tests are failing :(

VincentLanglet:
Test are considered as failing when you have self-deprecation:
```
Remaining self deprecation notices (2)

  2x: target page is deprecate since sonata-project/page-bundle 3.[27](https://github.com/sonata-project/SonataPageBundle/runs/7667108866?check_suite_focus=true#step:8:28).0, and it will be removed in 4.0
    2x in TransformerTest::testTransformerPageToSnapshot from Sonata\PageBundle\Tests\Entity
```

Either you should solve the self-deprecation or if you can't, add `@group legacy` above the tests

VincentLanglet (approved):

src/Entity/Transformer.php
```diff
@@ -242,31 +251,9 @@ public function getChildren(PageInterface $page)
     }
 
     /**
-     * @return array
-     */
-    protected function fixPageContent(array $content)
```
eerison, line 247: 1 - why do you need to remove this, just to cover with tests?
2 - as this class is not final, and if someone is using this method, it will break!

tests/Entity/TransformerTest.php
```diff
@@ -0,0 +1,249 @@
...
+                [
+                    'id' => 'block234',
+                    'name' => 'block2',
+                    'enabled' => false,
+                    'position' => 0,
+                    'settings' => [],
+                    'type' => 'type',
+                    'created_at' => $datetime->format('U'),
+                    'updated_at' => $datetime->format('U'),
+                    'blocks' => [],
```
eerison, line 234: are you covering this in other test, because I can't see!

and i't important, because it's the blocks inside the containers block for exmaple!

- header block container
   - menu block 1
   - menu block 2
- content block container
    - text block 1
    - text block 2

eerison: you can see the json example here: https://github.com/sonata-project/SonataPageBundle/pull/1501#issuecomment-1196325158

src/Entity/Transformer.php
VincentLanglet: Indeed, missed that ; I thought this was private/final.

We have to keep this on 3.x and add a NEXT_MAJOR to remove it in 4.x

VincentLanglet (dismissed):

src/Entity/Transformer.php
```diff
@@ -29,6 +29,9 @@
  * This class transform a SnapshotInterface into PageInterface.
  *
  * @final since sonata-project/page-bundle 3.26
+ *
+ * @phpstan-import-type PageContent from TransformerInterface
+ * @phpstan-import-type BlockContent from TransformerInterface
  */
 class Transformer implements TransformerInterface
```
eerison: if we going to remove this class on 4.x we should deprecate this class to remove on the next major!

Hanmac: i don't plan on removing this class in 4.x, just to use Serializer

eerison: But here leads to a problem if we deprecate this class to remove on 4.x we are not providing any other alternative for whom is extending this class!

should we implement the serializer solution in 3.x? if yes then we should deprecate this class together with serializer solution!

what do you think @VincentLanglet and @Hanmac ?

tests/Entity/TransformerTest.php
Hanmac: the current test data is Parent Page -> Child Page
The Child Page has 2 Blocks, one of them if the child of the other

but i can try your test data too

src/Entity/Transformer.php
VincentLanglet: So far the class is not removed and the implemenation of the Serializer is not provided. So that's another topic.

src/Entity/Transformer.php
Hanmac: curious, i thought they where only private in 4.x

might be a change in between, i can just add them back, and to be removed later in 4.x

tests/Entity/TransformerTest.php
eerison: > The Child Page has 2 Blocks, one of them if the child of the other

where are you checking this?

src/Entity/Transformer.php
eerison: Ok!

eerison: > i don't plan on removing this class in 4.x, just to use Serializer

hmm I thought that, it could be removed! because it is doing the serializer functionality! in theory you just need to pass the object into the normalizer like `$serializer->normalizer($block)`

tests/Entity/TransformerTest.php
Hanmac: block2 is the child of block1

src/Entity/Transformer.php
Hanmac: yeah, the Serializer will be called inside the Transformer when doing:
`$transformer->create($page) => $snapshot`

i didn't wanted to do it in this MR yet, because it most likely will be a BC break

eerison: ok :)

src/Model/TransformerInterface.php
```diff
@@ -17,6 +17,48 @@
...
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @phpstan-type BlockContent array{
+ *   id: int|string|null,
+ *   name?: string|null,
+ *   enabled: boolean,
+ *   position: int|null,
+ *   settings: array<string, mixed>,
+ *   type: string,
+ *   created_at: int|string|null,
```
jordisala1991: This was copied from a previous version of the 4.x branch, can you update it?

Do we need to backport this? PHPStan on this branch is so low, that will not catch problems here (opposite from what it happens on 4.x)

Hanmac: what exactly need to be updated?
or should i just leave it out for now? (it does exist in 4.x anyway)

jordisala1991: Check for differences on 4.x, multiple things changed.

tests/Entity/TransformerTest.php
eerison: Yeah I saw this, but you are not checking the output (array), you need to check if the block2 is converted as expected! because for now you are expecting that the children are empty `[]`

eerison: it should be like

```
'blocks' => [
                [
                    'id' => 'block234',
                    'name' => 'block1',
                    'enabled' => false,
                    'position' => 0,
                    'settings' => [],
                    'type' => 'type',
                    'created_at' => $datetime->format('U'),
                    'updated_at' => $datetime->format('U'),
                    'blocks' => [
                         [
                               'id' => 'block456',
                               'name' => 'block2',
                               'enabled' => false,
                         ],
                    ],
```

src/Model/TransformerInterface.php
Hanmac: i compared, besides the order i don't see any difference

again, should i just remove this and let it only be for 4.x?

VincentLanglet: Different order mean conflicts during the merge :/

Might indeed be better to not use type in 3.x if it's not useful to have less conflict during the merge. 🐙

tests/Entity/TransformerTest.php
Hanmac: the problem is you read this wrong,
we are already in the page blocks, from Line214-216

https://github.com/sonata-project/SonataPageBundle/pull/1516/files#diff-a14815f04837102213a846d666a6c4397b67e75309b476ebdd646dfc298bfd73R214-R216

so Line226+ is the first Block,
and Line236 is its child

src/Model/TransformerInterface.php
jordisala1991: > i compared, besides the order i don't see any difference
> 
> again, should i just remove this and let it only be for 4.x?

There is at least one difference. On 4.x created_type is numeric-string|null instead of int|string|null. And there are probably more, like type is string|null...

VincentLanglet (dismissed):

tests/Entity/TransformerTest.php
eerison: hmmm I got it! but just remember that we can have many levels of blocks, I don't know if it's usefull atm, but for now it's possible!

something like

- container 1
   - container 1.2
       .....
       - some other block 1

src/Entity/Transformer.php
```diff
@@ -150,7 +150,9 @@ public function load(SnapshotInterface $snapshot)
         $page->setJavascript($content['javascript']);
         $page->setStylesheet($content['stylesheet']);
         $page->setRawHeaders($content['raw_headers']);
-        $page->setTitle($content['title']);
+        if (isset($content['title'])) {
+            $page->setTitle($content['title']);
```
jordisala1991: On 4.x we are doing different: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/Transformer.php#L142

src/Entity/Transformer.php
```diff
@@ -177,11 +179,17 @@ public function loadBlock(array $content, PageInterface $page)
 
         $block->setPage($page);
         $block->setId($content['id']);
-        $block->setName($content['name']);
+        if (isset($content['name'])) {
+            $block->setName($content['name']);
```
jordisala1991: On 4.x we are doing different: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/Transformer.php#L167

VincentLanglet (approved):

src/Entity/Transformer.php
```diff
@@ -177,11 +177,13 @@ public function loadBlock(array $content, PageInterface $page)
 
         $block->setPage($page);
         $block->setId($content['id']);
-        $block->setName($content['name']);
+        $block->setName($content['name'] ?? null);
         $block->setEnabled($content['enabled']);
-        $block->setPosition($content['position']);
+        if (isset($content['position'])) {
+            $block->setPosition($content['position']);
```
jordisala1991, line 183: I would do the same here. Use the ?? operator with null.

Hanmac: i can do this for this branch because it is built against block bundle 3.x

but PageBundle 4.x will have problems against block bundle 4.x because:
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Model/BlockInterface.php#L46

see your MR for the problem: https://github.com/sonata-project/SonataPageBundle/pull/1501


so for this MR, i can fix it, and then let PageBundle 4.x be fixed later?

jordisala1991 (approved):

jordisala1991:
Thanks @Hanmac

PR #1518, opened 2022-08-04
Increase Psalm to level 4

jordisala1991:
## Subject

I am targeting this branch, because this is only possible on 4.x.

### DIFF
--- a/src/CmsManager/BaseCmsPageManager.php
+++ b/src/CmsManager/BaseCmsPageManager.php
@@ -28,7 +28,7 @@ abstract class BaseCmsPageManager implements CmsManagerInterface
     protected $currentPage;
 
     /**
-     * @var PageBlockInterface[]
+     * @var array<PageBlockInterface|null>
      */
     protected $blocks = [];
 
--- a/src/CmsManager/CmsManagerInterface.php
+++ b/src/CmsManager/CmsManagerInterface.php
@@ -92,9 +92,7 @@ interface CmsManagerInterface
     public function setCurrentPage(PageInterface $page);
 
     /**
-     * Returns the list of loaded block from the current http request.
-     *
-     * @return PageBlockInterface[]
+     * @return array<PageBlockInterface|null>
      */
     public function getBlocks();
 
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -100,8 +100,8 @@ final class Transformer implements TransformerInterface
         $content['meta_keyword'] = $page->getMetaKeyword();
         $content['template_code'] = $page->getTemplateCode();
         $content['request_method'] = $page->getRequestMethod();
-        $content['created_at'] = $page->getCreatedAt()->format('U');
-        $content['updated_at'] = $page->getUpdatedAt()->format('U');
+        $content['created_at'] = (int) $page->getCreatedAt()->format('U');
+        $content['updated_at'] = (int) $page->getUpdatedAt()->format('U');
         $content['slug'] = $page->getSlug();
         $content['parent_id'] = $page->getParent() ? $page->getParent()->getId() : null;
 
@@ -213,6 +213,9 @@ final class Transformer implements TransformerInterface
                 ->getQuery()
                 ->execute();
 
+            /**
+             * @var Collection<array-key, PageInterface>
+             */
             $collection = new ArrayCollection();
 
             foreach ($snapshots as $snapshot) {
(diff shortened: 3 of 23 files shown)


jordisala1991:
I need this PR released to fix last errors: https://github.com/sonata-project/SonataBlockBundle/pull/1093

src/Entity/Transformer.php
```diff
@@ -100,8 +100,8 @@ public function create(PageInterface $page)
         $content['meta_keyword'] = $page->getMetaKeyword();
         $content['template_code'] = $page->getTemplateCode();
         $content['request_method'] = $page->getRequestMethod();
-        $content['created_at'] = $page->getCreatedAt()->format('U');
-        $content['updated_at'] = $page->getUpdatedAt()->format('U');
+        $content['created_at'] = (int) $page->getCreatedAt()->format('U');
+        $content['updated_at'] = (int) $page->getUpdatedAt()->format('U');
```
jordisala1991, line 104: psalm does not report as numeric-string, and even if it does, it makes sense to me to typehint as int|null, since it is better to store it correctly instead of having to convert to int on the decode.

jordisala1991:
The missing errors are related to: https://github.com/sonata-project/SonataBlockBundle/pull/1093

VincentLanglet:
Can you try 5.x ; cf https://github.com/sonata-project/SonataBlockBundle/pull/1093#issuecomment-1206149553

jordisala1991:
Now it pass builds @VincentLanglet

src/Model/Site.php
```diff
@@ -231,7 +231,11 @@ public function setMetaKeywords($metaKeywords): void
     public function getUrl()
     {
         if ($this->isLocalhost()) {
-            return $this->getRelativePath();
+            $relativePath = $this->getRelativePath();
```
VincentLanglet: When relativePath is null, it will be used in the sprintf and it wont make sens.

Also, the sprintf shouldnt be used if isLocalhost.

I think we should throw a logicException if the relativePath is null (or used an assert if we're not sure).

Or allowing to return null (or returning '').

src/Request/SiteRequestContext.php
```diff
@@ -46,7 +46,11 @@ public function getHost(): string
         $site = $this->getSite();
 
         if ($site && !$site->isLocalhost()) {
-            return $site->getHost();
+            $host = $site->getHost();
+
+            if ($host) {
```
VincentLanglet: I wonder if we should throw an exception when it's null.

jordisala1991: I went for the exception here.

src/Model/Site.php
jordisala1991: I went for allowing null here.

src/CmsManager/BaseCmsPageManager.php
```diff
@@ -28,7 +28,7 @@ abstract class BaseCmsPageManager implements CmsManagerInterface
     protected $currentPage;
 
     /**
-     * @var PageBlockInterface[]
+     * @var array<PageBlockInterface|null>
```
VincentLanglet, line 31: Isn't weird to allow null ? Shouldn't the value be filtered if needed when set.

src/CmsManager/CmsManagerInterface.php
```diff
@@ -92,9 +92,7 @@ public function getCurrentPage();
     public function setCurrentPage(PageInterface $page);
 
     /**
-     * Returns the list of loaded block from the current http request.
-     *
-     * @return PageBlockInterface[]
+     * @return array<PageBlockInterface|null>
```
VincentLanglet, line 95: Isn't weird to allow null ? Shouldn't the value be filtered if needed when set.

VincentLanglet (changes requested):

src/CmsManager/BaseCmsPageManager.php
jordisala1991: I was not sure about it, this property is used as a "kind of" cache:

https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/CmsManager/CmsPageManager.php#L137-L144

With that code null values get always fetched, even though we know those are null, but I changed that code to check for key exist instead (in this same PR):

https://github.com/sonata-project/SonataPageBundle/blob/14b56cf73b716587f55730d88d7130455c67be0d/src/CmsManager/CmsPageManager.php#L137-L144

So the logic now will see if the key is defined and return the null. Previously it would check on database for those nulls.

WDYT? I can change to not allow nulls too.

VincentLanglet: Oh, I see. It can make sens then.

VincentLanglet (approved):

PR #1519, opened 2022-08-05
Increase PHPStan to level 8

jordisala1991:
## Subject

I am targeting this branch, because this can only be applied to 4.x.

## Changelog

```markdown
### Changed
- `PageAdmin` now injects its required dependencies through `__construct` instead of set methods.
```

### DIFF
--- a/src/Admin/Extension/CreateSnapshotAdminExtension.php
+++ b/src/Admin/Extension/CreateSnapshotAdminExtension.php
@@ -49,13 +49,13 @@ final class CreateSnapshotAdminExtension extends AbstractAdminExtension
     private function createByPage(object $object): void
     {
         if ($object instanceof PageBlockInterface) {
-            $page = $object->getPage();
-        } elseif ($object instanceof PageInterface) {
-            $page = $object;
-        } else {
+            $object = $object->getPage();
+        }
+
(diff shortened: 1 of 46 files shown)


src/Admin/Extension/CreateSnapshotAdminExtension.php
```diff
@@ -50,6 +50,10 @@ private function createByPage(object $object): void
     {
         if ($object instanceof PageBlockInterface) {
```
VincentLanglet, line 51:
```
if ($object instanceof PageBlockInterface) {
    $object = $object->getPage();
}

if (!$object instanceof PageInterface) {
    return;
}

jordisala1991: I don't like a lot to replace variables that comes from parameters, but I will try to find a cleaner option.

jordisala1991: I didn't found a cleaner solution, I applied your patch.

jordisala1991:
This one is probably a bit harder to review, had to change some logics around to avoid null variables @VincentLanglet .

Not sure if I followed the best option in each specific case.

VincentLanglet (approved):

PR #1528, opened 2022-08-08
Increase Psalm to level 2

jordisala1991:
## Subject

I am targeting this branch, because this can only be applied to 4.x.

## Changelog

```markdown
### Changed
- `BlockAdmin` and its childs now inject their required dependencies through `__construct` instead of set methods.
```

### DIFF
--- a/src/Admin/PageAdmin.php
+++ b/src/Admin/PageAdmin.php
@@ -22,6 +22,7 @@ use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Route\RouteCollectionInterface;
 use Sonata\AdminBundle\Security\Acl\Permission\AdminPermissionMap;
 use Sonata\AdminBundle\Show\ShowMapper;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\DoctrineORMAdminBundle\Filter\CallbackFilter;
 use Sonata\PageBundle\Exception\InternalErrorException;
 use Sonata\PageBundle\Exception\PageNotFoundException;
@@ -196,10 +197,12 @@ final class PageAdmin extends AbstractAdmin
             ->add('parent')
             ->add('edited')
             ->add('hybrid', CallbackFilter::class, [
-                'callback' => static function ($queryBuilder, $alias, $field, $data): void {
+                'callback' => static function (ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data): void {
+                    $builder = $queryBuilder->getQueryBuilder();
+
                     if (\in_array($data['value'], ['hybrid', 'cms'], true)) {
-                        $queryBuilder->andWhere(sprintf('%s.routeName %s :routeName', $alias, 'cms' === $data['value'] ? '=' : '!='));
-                        $queryBuilder->setParameter('routeName', PageInterface::PAGE_ROUTE_CMS_NAME);
+                        $builder->andWhere(sprintf('%s.routeName %s :routeName', $alias, 'cms' === $data['value'] ? '=' : '!='));
+                        $builder->setParameter('routeName', PageInterface::PAGE_ROUTE_CMS_NAME);
                     }
                 },
                 'field_options' => [
--- a/src/Model/SnapshotChildrenCollection.php
+++ b/src/Model/SnapshotChildrenCollection.php
@@ -19,6 +19,10 @@ use Doctrine\Common\Collections\AbstractLazyCollection;
  * @extends AbstractLazyCollection<array-key, PageInterface>
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
+ *
+ * @see $collection property is initialized in the doInitialize method
  */
 final class SnapshotChildrenCollection extends AbstractLazyCollection
 {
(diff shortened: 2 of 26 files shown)


src/Admin/PageAdmin.php
```diff
@@ -196,7 +198,9 @@ protected function configureDatagridFilters(DatagridMapper $filter): void
             ->add('parent')
             ->add('edited')
             ->add('hybrid', CallbackFilter::class, [
-                'callback' => static function ($queryBuilder, $alias, $field, $data): void {
+                'callback' => static function (ProxyQueryInterface $queryBuilder, string $alias, string $field, array $data): void {
+                    \assert($queryBuilder instanceof ProxyQuery);
```
VincentLanglet: Can't we use ProxyQuery in the typehint of the callback then ?

src/Admin/PageAdmin.php
```diff
@@ -196,7 +198,9 @@ protected function configureDatagridFilters(DatagridMapper $filter): void
...
+
                     if (\in_array($data['value'], ['hybrid', 'cms'], true)) {
                         $queryBuilder->andWhere(sprintf('%s.routeName %s :routeName', $alias, 'cms' === $data['value'] ? '=' : '!='));
                         $queryBuilder->setParameter('routeName', PageInterface::PAGE_ROUTE_CMS_NAME);
```
VincentLanglet: Might be better to rely on the getQueryBuilder() method which exists on the two interface if we want to later add mongo support https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/blob/4.x/src/Datagrid/ProxyQueryInterface.php

src/Model/SnapshotChildrenCollection.php
```diff
@@ -19,6 +19,8 @@
  * @extends AbstractLazyCollection<array-key, PageInterface>
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
```
VincentLanglet, line 23: Can you add an explanation about the why ?

And we should do a PR for fixing the phpdoc of https://github.com/doctrine/collections/blob/1.6.x/lib/Doctrine/Common/Collections/AbstractLazyCollection.php

Do you want to do it or should I ?

tests/App/AppKernel.php
```diff
@@ -91,6 +96,9 @@ protected function configureRoutes($routes): void
         $routes->import($this->getProjectDir().'/config/routes.yaml');
     }
 
+    /**
+     * @psalm-suppress DeprecatedClass
```
VincentLanglet, line 100: A comment about it ?

VincentLanglet (changes requested):

src/Admin/PageAdmin.php
jordisala1991: I followed docs of DoctrineORMAdminBundle: https://docs.sonata-project.org/projects/SonataDoctrineORMAdminBundle/en/4.x/reference/filter_field_definition/#callback.

Maybe those docs need an update?

What I don't like about that ProxyQueryInterface is that is specific for the DoctrineORMAdminBundle but does not represent all the method you can call for the real object... IMO is something that could be fixed for 5.0.

For this PR, do you prefer to change typehint?

src/Admin/PageAdmin.php
jordisala1991: Oh, ok I will try.

src/Model/SnapshotChildrenCollection.php
jordisala1991: If you have time to do it, it would be nice. IMHO in this case I see nothing wrong with that class. Psalm complains about a property that is not set on the construct by design, because it is a lazy collection.

tests/App/AppKernel.php
jordisala1991: We didn't add it on other bundles. This method is rather simple, it will be removed once SonataBlock bundle is version 5 only.

src/Admin/PageAdmin.php
VincentLanglet: > What I don't like about that ProxyQueryInterface is that is specific for the DoctrineORMAdminBundle but does not represent all the method you can call for the real object... IMO is something that could be fixed for 5.0.

I personally dont like the `__call`, all the method you can call on the real object are just the same than `->getQueryBuilder->yourMethod()`. But I didn't removed it because I assume a lot of people are using it.

I think you wont need an assert if you call getQueryBuilder first

src/Model/SnapshotChildrenCollection.php
VincentLanglet: It should be
```
/**
     * The backed collection to use
     *
     * @psalm-var Collection<TKey,T>|null
     * @var Collection<mixed>|null
     */
    protected $collection;
```
then

jordisala1991: Yep, it is more safe that way. But it is not wrong as of now, because all the methods that use collection will initialize it first.

src/Admin/PageAdmin.php
jordisala1991: Fixed, probably we could deprecate the __call in favour of getQueryBuilder.

src/Admin/PageAdmin.php
jordisala1991: Fixed.

src/Model/SnapshotChildrenCollection.php
jordisala1991: Added a comment.

VincentLanglet (approved):

PR #1531, opened 2022-08-09
Move to PHP configuration

jordisala1991:
## Subject

I am targeting this branch, because this is easier to apply to 4.x.

## Changelog

```markdown
### Changed
- Change from xml configuration format to PHP, for the internal bundle configuration.
```


eerison:
What is the reason for it? Because symfony doesn't need to convert xml file to php? or is there other reason? I'm just curios about this 😄 , I guess I heard something that symfony was doing the same thing for internal files!

VincentLanglet:
> What is the reason for it? Because symfony doesn't need to convert xml file to php? or is there other reason? I'm just curios about this 😄 , I guess I heard something that symfony was doing the same thing for internal files!

Php is better than xml/yaml. You have validation, autocompletion, etc

jordisala1991:
And you don't need php-xml extension to parse the configuration files (we need for other parts of the app, but we will someday get to a point where we only need php files to run the app).

VincentLanglet (approved):

PR #1534, opened 2022-08-10
Add tests for Site and Snapshot Admins

jordisala1991:
## Subject

I am targeting this branch, because this increases coverage for next release.

* First commit, add new functional tests. Fails with: 

`The "symfony/intl" component is required to use "Symfony\Component\Form\Extension\Core\Type\LocaleType". Try running "composer require symfony/intl`

Fails at: https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Admin/SiteAdmin.php#L105

IMHO if the Site concept is something mandatory, we should add this dep to require section.

* Second commit, add dependency with symfony intl

## Changelog

```markdown
### Added
- Add missing dependency to `symfony/intl`.
```

### DIFF
--- a/src/Admin/PageAdmin.php
+++ b/src/Admin/PageAdmin.php
@@ -62,11 +62,11 @@ final class PageAdmin extends AbstractAdmin
 
     protected function configureRoutes(RouteCollectionInterface $collection): void
     {
-        $collection->add('compose', '{id}/compose', [
-            'id' => null,
+        $collection->add('compose', $this->getRouterIdParameter().'/compose', [
+            $this->getIdParameter() => null,
         ]);
-        $collection->add('compose_container_show', 'compose/container/{id}', [
-            'id' => null,
+        $collection->add('compose_container_show', 'compose/container/'.$this->getRouterIdParameter(), [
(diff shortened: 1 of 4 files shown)


src/Admin/PageAdmin.php
```diff
@@ -62,11 +62,11 @@ public function __construct(
...
     {
-        $collection->add('compose', '{id}/compose', [
-            'id' => null,
+        $collection->add('compose', $this->getRouterIdParameter().'/compose', [
+            $this->getIdParameter() => null,
         ]);
-        $collection->add('compose_container_show', 'compose/container/{id}', [
-            'id' => null,
+        $collection->add('compose_container_show', 'compose/container/'.$this->getRouterIdParameter(), [
+            $this->getIdParameter() => null,
```
jordisala1991, line 69: pedantic change, I saw it being used on SiteAdmin, I like it more that using id directly  wdyt @VincentLanglet ?

VincentLanglet: Indeed seems nice

VincentLanglet (approved):

PR #1538, opened 2022-08-10
Add more tests

jordisala1991:
## Subject

More fixes to existing tests and add new tests. With this one we cover all the urls (GET), we miss the POST actions. like actually creating pages, and other things (next PRs).

### DIFF
--- a/src/Admin/BaseBlockAdmin.php
+++ b/src/Admin/BaseBlockAdmin.php
@@ -17,7 +17,6 @@ use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
-use Sonata\AdminBundle\Route\RouteCollectionInterface;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Sonata\PageBundle\Model\PageBlockInterface;
 use Sonata\PageBundle\Model\PageInterface;
@@ -117,11 +116,6 @@ abstract class BaseBlockAdmin extends AbstractAdmin
         return [
             'type' => $this->getRequest()->get('type'),
         ];
-    }
-
-    protected function configureRoutes(RouteCollectionInterface $collection): void
-    {
-        $collection->add('view', $this->getRouterIdParameter().'/view');
     }
 
     protected function configureListFields(ListMapper $list): void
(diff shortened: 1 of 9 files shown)


tests/Functional/Admin/BlockAdminTest.php
```diff
@@ -48,7 +48,7 @@ public static function provideCrudUrlsCases(): iterable
         yield 'List Block' => ['/admin/tests/app/sonatapageblock/list'];
         yield 'Edit Block' => ['/admin/tests/app/sonatapageblock/1/edit'];
         yield 'Remove Block' => ['/admin/tests/app/sonatapageblock/1/delete'];
-        yield 'Compose preview Block' => ['/admin/tests/app/sonatapageblock/2/compose_preview'];
+        yield 'Compose preview Block' => ['/admin/tests/app/sonatapageblock/2/compose-preview'];
```
jordisala1991, line 51: IMO underscores on urls do not make sense, better use -

src/Admin/BaseBlockAdmin.php
```diff
@@ -119,11 +118,6 @@ protected function configurePersistentParameters(): array
         ];
     }
 
-    protected function configureRoutes(RouteCollectionInterface $collection): void
-    {
-        $collection->add('view', $this->getRouterIdParameter().'/view');
```
jordisala1991, line 124: This route was added 8 years ago, without any comment, but it was never implemented.

jordisala1991: actually 12 years ago, probably on one of the first commits, probably never worked.

src/Controller/BlockAdminController.php
```diff
@@ -86,10 +86,6 @@ public function createAction(Request $request): Response
     {
         $this->admin->checkAccess('create');
 
-        if (!$this->admin->isChild()) {
```
jordisala1991, line 89: This check was modified by me on a previous PR. IMO we can remove it. Create action is executable on by child admin OR by shared block admin.

jordisala1991: But the remaining one (BlockAdmin) does not have the create route, I have remove it.

VincentLanglet (changes requested):
If I understand correctly you're changing the route name/url.

This is a Major BC break which should be in the Upgrade note then.

jordisala1991:
> If I understand correctly you're changing the route name/url.
> 
> This is a Major BC break which should be in the Upgrade note then.

Added UPGRADE note.

VincentLanglet (approved):

PR #1540, opened 2022-08-10
Add more tests for exception controller

jordisala1991:
## Subject

Again more tests in 2 commits, the first one with tests , the second one with fixes needed to pass tests.

### DIFF
--- a/src/Controller/PageController.php
+++ b/src/Controller/PageController.php
@@ -37,7 +37,7 @@ final class PageController extends AbstractController
     public function __construct(
         ExceptionListener $exceptionListener,
         PageServiceManagerInterface $pageServiceManager,
-        CmsManagerSelectorInterface $cmsSelector,
+        CmsManagerSelectorInterface $cmsSelector
     ) {
         $this->exceptionListener = $exceptionListener;
         $this->pageServiceManager = $pageServiceManager;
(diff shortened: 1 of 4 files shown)


src/Controller/PageController.php
```diff
@@ -61,16 +61,16 @@ public function exceptionsList(): Response
     /**
      * @throws InternalErrorException|AccessDeniedException
      */
-    public function exceptionEdit(Request $request, string $code): Response
+    public function exceptionEdit(Request $request, int $code): Response
```
jordisala1991, line 64: I added this typehint in a previous PR, so no change should be needed on the UPGRADE note.

src/Controller/PageController.php
```diff
@@ -37,7 +37,7 @@ final class PageController extends AbstractController
     public function __construct(
         ExceptionListener $exceptionListener,
         PageServiceManagerInterface $pageServiceManager,
-        CmsManagerSelectorInterface $cmsSelector,
+        CmsManagerSelectorInterface $cmsSelector
```
jordisala1991, line 40: We did not have this loaded on the container, so we didn't catch this error (only for php <= 7.4)

VincentLanglet (approved):

PR #1536, opened 2022-08-10
fix Cleanup Query if no snapshot to keep

Hanmac:
## Subject

The Query is faulty if there is no Snapshot to keep.
I should probably extend the Testcases for this too.

I am targeting this branch, because it is a Bug Fix.

Closes #1535.

## Changelog

```markdown
### Fixed
- SnapshotManager Cleanup Query if there are no Snapshots to keep
```

PR #1542, opened 2022-08-10
Add test for Cleanup Snapshots Command

jordisala1991:
## Subject

More test, this one does not fix anything because it was previously fixed here: https://github.com/sonata-project/SonataPageBundle/pull/1536

But it catched the error, it got fixed after I merged the PR and merge to 4.x the fix.


tests/Command/CleanupSnapshotsCommandTest.php
```diff
@@ -1,92 +0,0 @@
-<?php
```
jordisala1991, line 1: IMO it makes more sense to test this as a completely functional (e2e) test, instead of mocking things.

VincentLanglet (approved):

PR #1550, opened 2022-08-11
Added TransactionalManagerInterface

eerison:
## Subject

## Changelog

```markdown
### Added

### Changed

### Deprecated

### Removed

### Fixed

### Security
```


src/Model/SnapshotManagerInterface.php
```diff
@@ -20,7 +20,7 @@
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-interface SnapshotManagerInterface extends ManagerInterface
+interface SnapshotManagerInterface extends ManagerInterface, TransactionalManagerInterface
```
eerison, line 24: Ok here is the point that I want to have an agreement with you

is it fine I extends this in `SnapshotManagerInterface` or do you see any problem for this?

@VincentLanglet  @jordisala1991

src/Model/TransactionalManagerInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Sonata\PageBundle\Model;
+
+interface TransactionalManagerInterface
```
VincentLanglet: This interface need to be introduce in sonata-doctrine-extensions instead.

This is not a PR for this bundle.

src/Entity/SnapshotManager.php
```diff
@@ -168,4 +168,14 @@ public function createSnapshotPageProxy(TransformerInterface $transformer, Snaps
         return $this->snapshotPageProxyFactory
             ->create($this, $transformer, $snapshot);
     }
+
+    function beginTransaction(): void
+    {
+        $this->getEntityManager()->beginTransaction();
+    }
+
+    function commit(): void
```
VincentLanglet: We should add `rollBack` to the interface too.

(But first the pr should be done on sonata-doctrine-extensions instead)

VincentLanglet (changes requested):

src/Model/TransactionalManagerInterface.php
eerison: Then I'll introduce in `sonata-doctrine-extensions` and here, correct?

VincentLanglet: Just in sonata-doctrine-extension ; there is no need to duplicate this here.

VincentLanglet: The code will be
```
if ($this->snapshotManager instanceof TransactionalManagerInterface) {
            $this->snapshotManager->beginTransaction();
        }
```

eerison: Ok, But I can't depends the last version of `sonata-doctrine-extension`, because it forces to use php 8

Edit 1: Or can I add `^7.4 | ^8.0` in composer.json?

eerison: > The code will be
> 
> ```
> if ($this->snapshotManager instanceof TransactionalManagerInterface) {
>             $this->snapshotManager->beginTransaction();
>         }
> ```

and how do I do in the interface, I mean `SnapshotManagerInterface`?

VincentLanglet: > Ok, But I can't depends the last version of `sonata-doctrine-extension`, because it forces to use php 8
> 
> Edit 1: Or can I add `^7.4 | ^8.0` in composer.json?

I see two options:
- Introducing the Interface in 1.x and 2.x of sonata-doctrine-extensions.
- Just introducing the interface in 2.x and in SonataPageBundle using
```
// TODO: Use TransactionalManagerInterface check instead when drop doctrine-extension 2
if ($this->snapshotManager instanceof BaseEntityManager) {
    $this->snapshotManager->getEntityManager()->beginTransaction();
}
```

This might be ok to introduce the interface in 1.x ; WDYT @jordisala1991 ?

eerison: Ok question: Can we introduce this in 1.x and depends the 1.x version 👀 

and remove the support for 2.x: [`https://github.com/sonata-project/SonataPageBundle/blob/4.x/composer.json#L30`](https://github.com/sonata-project/SonataPageBundle/blob/4.x/composer.json#L30)

when we drop the 7.4 support we can support 2.x again, what do you think 👀

VincentLanglet: > Ok question: Can we introduce this in 1.x and depends the 1.x version 👀

I edited ; I meant 1 and 2, rather than 2 and 3

VincentLanglet: > Ok question: Can we introduce this in 1.x and depends the 1.x version 👀
> 
> and remove the support for 2.x: [`https://github.com/sonata-project/SonataPageBundle/blob/4.x/composer.json#L30`](https://github.com/sonata-project/SonataPageBundle/blob/4.x/composer.json#L30)
> 
> when we drop the 7.4 support we can support 2.x again, what do you think 👀

No need to remove support for 2.x if you introduce the interface in 1.x.

You'll support `^1.18 || ^2.1` because of both of these version will have the interface

eerison: 🤯 you are right!

Then I'll open the PR to 1.x ok?

eerison: Done: https://github.com/sonata-project/sonata-doctrine-extensions/pull/443

composer.json
```diff
@@ -27,7 +27,7 @@
         "doctrine/persistence": "^2.1 || ^3.0",
         "sonata-project/admin-bundle": "^4.15",
         "sonata-project/block-bundle": "^4.16",
-        "sonata-project/doctrine-extensions": "^1.8 || ^2.0",
+        "sonata-project/doctrine-extensions": "1.x-dev as 1.18",
```
eerison, line 30:
```suggestion
        "sonata-project/doctrine-extensions": "^1.18 || ^2.1",
```

It can be applied after generate new releases in that bundle!

src/Model/SnapshotManagerInterface.php
```diff
@@ -14,13 +14,14 @@
...
 use Sonata\Doctrine\Model\ManagerInterface;
+use Sonata\Doctrine\Model\TransactionalManagerInterface;
 
 /**
  * @extends ManagerInterface<SnapshotInterface>
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-interface SnapshotManagerInterface extends ManagerInterface
+interface SnapshotManagerInterface extends ManagerInterface, TransactionalManagerInterface
```
VincentLanglet, line 24: The fact we force the SnapshotManagerInterface to implements TransactionalManagerInterface is not a good idea IMHO, it will block Mongo support.

src/Service/CleanupSnapshotService.php
```diff
@@ -35,16 +34,12 @@ public function cleanupBySite(SiteInterface $site, int $keepSnapshots): void
     {
         $pages = $this->pageManager->findBy(['site' => $site->getId()]);
 
-        if ($this->snapshotManager instanceof BaseEntityManager) {
-            $this->snapshotManager->getEntityManager()->beginTransaction();
-        }
+        $this->snapshotManager->beginTransaction();
```
VincentLanglet, line 37: if ($this->snapshotManager instanceof TransactionalManagerInterface) {
    $this->snapshotManager->beginTransaction();
}

VincentLanglet (changes requested):

src/Service/CleanupSnapshotService.php
eerison: if I do this, it won't work, this method will not do anything! IMO it doesn't make sense to add this "if"

What do you think to create `SnapshotEntityManagerInterface` and extends it from `SnapshotManagerInterface` and `TransactionalManagerInterface`?

jordisala1991: I agree with @VincentLanglet , this interface was introduced precisely for that reason, we will only do transactions if the manager supports them. You do not want the model folder to force the use of the interface, because someone could implement MongoDB (without transactions) and it will be forced to add those methods, even if he knows they make no sense in its implementation.

eerison: Ok I got your point, But this "concrete" class was build to work only with ORM! if the user wants to make this work for ODM, then he should create a new class implementing `CleanupSnapshotBySiteInterface`.

if you look into the implementation we gonna have more problems if the user try to use ODM here like

```php
$pages = $this->pageManager->findBy(['site' => $site->getId()]);
```

it will broken too, this class definitely doesn't expect to receive an `ODM`


if `SnapshotManagerInterface` is used for `ORM` and `ODM`, it's fine, let's create an abstraction of this class like `SnapshotEntityManagerInterface` and require this into the constructor!

eerison: > Ok I got your point, But this "concrete" class was build to work only with ORM! if the user wants to make this work for ODM, then he should create a new class implementing `CleanupSnapshotBySiteInterface`.
> 
> if you look into the implementation we gonna have more problems if the user try to use ODM here like
> 
> ```
> $pages = $this->pageManager->findBy(['site' => $site->getId()]);
> ```
> 
> it will broken too, this class definitely doesn't expect to receive an `ODM`
> 
> if `SnapshotManagerInterface` is used for `ORM` and `ODM`, it's fine, let's create an abstraction of this class like `SnapshotEntityManagerInterface` and require this into the constructor!

maybe we could add such examples into the doc

VincentLanglet: > Ok I got your point, But this "concrete" class was build to work only with ORM! if the user wants to make this work for ODM, then he should create a new class implementing `CleanupSnapshotBySiteInterface`.

The class/interface was not build to only work with ORM. That's all the purpose of using abstraction from sonata-doctrine-extensions.

> if you look into the implementation we gonna have more problems if the user try to use ODM here like
> 
> ```
> $pages = $this->pageManager->findBy(['site' => $site->getId()]);
> ```
> 
> it will broken too, this class definitely doesn't expect to receive an `ODM`

https://github.com/doctrine/mongodb-odm/blob/2.4.x/lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php#L172

I don't know MongoDB a lot but findBy exists in ODM

> if `SnapshotManagerInterface` is used for `ORM` and `ODM`, it's fine, let's create an abstraction of this class like `SnapshotEntityManagerInterface` and require this into the constructor!

I prefer to let possible the fact that one day ODM will be supported, rather by restricting to ORM in the constructor.

eerison: > I prefer to let possible the fact that one day ODM will be supported, rather by restricting to ORM in the constructor.

But the `CleanupSnapshotBySiteInterface` can support ODM, he just need to decorate this class [src/Service/CleanupSnapshotService.php](https://github.com/sonata-project/SonataPageBundle/pull/1550/files/69e68b0627e541e78ae32779b3edb2ad04e0516a#diff-112198c8b3816a6af7235c0c526ba39544d633700e4538d9ab1fc7e7c770d8f6)!

But let's imagine that someone going to implement something to support `clean snapshot` for ODM! then he needs to make conditions to check if it's ORM or ODM, don't you think that should be better he Create a new class and just implement the `CleanupSnapshotBySiteInterface`?

eerison: and what's happen if I want to use other thing different to save my snapshots? like redis, then should I check if it's `ORM`, `ODM` or `Redis`?

VincentLanglet: > But the `CleanupSnapshotBySiteInterface` can support ODM, he just need to decorate this class [src/Service/CleanupSnapshotService.php](https://github.com/sonata-project/SonataPageBundle/pull/1550/files/69e68b0627e541e78ae32779b3edb2ad04e0516a#diff-112198c8b3816a6af7235c0c526ba39544d633700e4538d9ab1fc7e7c770d8f6)!

With the check
```
if ($this->snapshotManager instanceof TransactionalManagerInterface)
```
he won't have to decorate at all.

> But let's imagine that someone going to implement something to support `clean snapshot` for ODM! then he needs to make conditions to check if it's ORM or ODM, don't you think that should be better he Create a new class and just implement the `CleanupSnapshotBySiteInterface`?

I don't see why he need to check ORM or ODM. The goal is that this class support both.

jordisala1991: In theory, to implement another persistency strategy, you just need to replicate the "Entity" folder (Which is specific for doctrine ORM) and name it whatever you want (Redis, Document for ODM), take a look at other bundles like SonataMediaBundle, SonataClassificationBundle and SonataUserBundle.

VincentLanglet: See https://github.com/sonata-project/SonataMediaBundle/tree/4.x/src/Document as an exemple.

There is a GalleryManager which implements the GalleryManagerInterface (defined in the Model folder).
If the GalleryManagerINterface would have extends the TransactionalManagerInterface it would have block the ODM support.

eerison: > In theory, to implement another persistency strategy, you just need to replicate the "Entity" folder (Which is specific for doctrine ORM) and name it whatever you want (Redis, Document for ODM), take a look at other bundles like SonataMediaBundle, SonataClassificationBundle and SonataUserBundle.

- Could you provide one example?

- But look I do not want to change all my persist locale, just the snapshot stuffs because I could use "Doctrine ORM" to save my data for others entities, But the snapshot json, I want to save in redis!

is it possible to do this with doctrine?

eerison: > There is a GalleryManager which implements the GalleryManagerInterface (defined in the Model folder).
> If the GalleryManagerINterface would have extends the TransactionalManagerInterface it would have block the ODM support.

I checked, and I got the idea, I know that `GalleryManagerINterface` should works for both, like `SnapshotManagerInterface`, But I still doesn't agree with this 😄 for me no make sense depends of an Interface that need to add "ifs" to check which code should be executed! IMO we should depends the interface that offers that methods that are necessary for my code works! IMO it's a good start to let the class do more that would suppose to do!

eerison: But well here we have 2 votes x 1 vote 😄 

I guess it won't implement more then `ORD` and `ODM` and in case I want to do something customised like check in different place I can still make my own implementation!

eerison: Ok But as ATM it just work for `ORM` can I check throw an exeption in case he is not using `ORM`? when someone implements the ODM, then we can remove the `Exception`!

if yes, which exception should I return?

src/Model/SnapshotManagerInterface.php
jordisala1991, line 24:
```suggestion
interface SnapshotManagerInterface extends ManagerInterface
```

This should be removed.

src/Service/CleanupSnapshotService.php
jordisala1991: No, we need to maintain the if with the check for instanceof, that the most compatible solution.

VincentLanglet: I don't see why an exception should be thrown 

>  But I still doesn't agree with this 😄 for me no make sense depends of an Interface that need to add "ifs" to check which code should be executed!  IMO we should depends the interface that offers that methods that are necessary for my code works!

If it implements Transactional, you run the code inside a transaction
If it don't, you still run the code, but without beginTransaction and commit. The code will still work and save the things.
```
$this->createByPage($page);
```
works for ORM and ODM

So the Transactional is not mandatory.

eerison:
Ok I'm struggling a bit to mock the final class like `SnapshotManager`

What do you think to add it? https://tomasvotruba.com/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/

VincentLanglet:
> Ok I'm struggling a bit to mock the final class like `SnapshotManager`
> 
> What do you think to add it? https://tomasvotruba.com/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/

You can use the real class `new SnapshotManaer()` (and mock the dependencies if needed) ; that's generally how we do it.

eerison:
> > Ok I'm struggling a bit to mock the final class like `SnapshotManager`
> > What do you think to add it? https://tomasvotruba.com/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/
> 
> You can use the real class `new SnapshotManaer()` (and mock the dependencies if needed) ; that's generally how we do it.

My test is: check if the `commit` and `beginTransaction` are called! 

I saw that I need to check this by `registry`, I can do this, But maybe this dependency could let's ours tests simplest.

eerison:
> > > Ok I'm struggling a bit to mock the final class like `SnapshotManager`
> > > What do you think to add it? https://tomasvotruba.com/blog/2019/03/28/how-to-mock-final-classes-in-phpunit/
> > 
> > 
> > You can use the real class `new SnapshotManaer()` (and mock the dependencies if needed) ; that's generally how we do it.
> 
> My test is: check if the `commit` and `beginTransaction` are called!
> 
> I saw that I need to check this by `registry`, I can do this, But maybe this dependency could let's ours tests simplest.

But we have a problem here, to mock it, I will need to mock methods created in others dependencies, and it's a problem, because if the dependency change  their implementation, my test gonna broken!

and I remember that @jordisala1991 mention that it's a bad practice mock things that are not in our code/context, and it makes sense!

Note: Well I wouldn't not to mock the container to make this work :'(, IMO it should be just a simple unit test.

eerison:
```php
// Mocks
        $objectManagerMock = $this->createMock(EntityManagerInterface::class);
        $objectManagerMock
            ->expects(static::once())
            ->method('beginTransaction');

        $objectManagerMock
            ->expects(static::once())
            ->method('commit');

        $registryMock = $this->createMock(ManagerRegistry::class);
        $registryMock->method('getManagerForClass')->willReturn($objectManagerMock);
        $snapshotPageProxyFactoryMock = $this->createMock(SnapshotPageProxyFactoryInterface::class);

        $snapshotManagerMock = new SnapshotManager(Snapshot::class, $registryMock, $snapshotPageProxyFactoryMock);
```

error: `TypeError : Return value of Sonata\Doctrine\Entity\BaseEntityManager::getRepository() must be an instance of Doctrine\ORM\EntityRepository, null returned`

I'm getting errors related with implementation of others stuffs 😞

eerison:
I'm closing this PR in favor of #1603

Issue #1453, opened 2022-07-14
after create a new snapshot, it goes for page edit, and should back to the list


eerison:
after create a new snapshot, it goes for page edit, and should back to the list

mesiarm:
Maybe it could be simplified - add another button in edit page to update and create snapshot with one click.

eerison:
> Maybe it could be simplified - add another button in edit page to update and create snapshot with one click.

In dev environment it already happen, but if we do this in prod, no make sense maintain the snapshot

eerison:
And IMO we should keep the snapshot feature as it's!

haivala:
I would say that snapshots should be able to be timed in prod so that when time comes another content would be shown to users that are not logged in.

I have used them like this: edit content -> make publication -> edit content again and make new snapshot -> then edit times of the snapshots in question.

eerison:
> I would say that snapshots should be able to be timed in prod so that when time comes another content would be shown to users that are not logged in.
> 
> I have used them like this: edit content -> make publication -> edit content again and make new snapshot -> then edit times of the snapshots in question.

Hey @haivala 

Maybe the issue's description is not really clear, but the snapshot behavior shouldn't be changed here.


It's just for change the redirect after create the snapshot on admin, because for now it's redirecting to page update, but it should return to snapshot list.

eerison:
> I would say that snapshots should be able to be timed in prod so that when time comes another content would be shown to users that are not logged in.
> 
> I have used them like this: edit content -> make publication -> edit content again and make new snapshot -> then edit times of the snapshots in question.

But if you want to elaborate an issue explaining better your proposal and open a PR, it should be good 😊

mesiarm:
I agree, timing should be kept, but add another way to do same with fewer clicks. Button in edit page could have split dropdown, which opens small dialog with calendar and submit button. Or default action to publish now.

haivala:
I really haven't thought how the timed snapshots could or should be achieved better (Like in the admin UI) but at the current state it is possible but needs little bit of hassle. I just wanted to open a discussion about the snapshots and let you know how timed snapshots can be used now. This is just a part of the UI where I imagined this functionality should be added.

I'm on a vacation now but after it l probably can open issue about this properly.

jordisala1991:
On 4.x this is fixed.

PR #1551, opened 2022-08-11
Add form tests for SnapshotAdmin

jordisala1991:
## Subject

Closes https://github.com/sonata-project/SonataPageBundle/issues/1453.


Hanmac:
Question: why do you make the transformer create a new Snapshot object, when you just grab the data and put it into the existing one?

wouldn't it be better to have a new Transformer method instead?

jordisala1991:
> Question: why do you make the transformer create a new Snapshot object, when you just grab the data and put it into the existing one?
> 
> wouldn't it be better to have a new Transformer method instead?

I didn't tried that path, but at some point I will have to redo all this setters. If I do it on transformer, I will need a need method just for this case (missing feature from AdminBundle to completely replace the object on prePersist(). And that method probably will replicate some logic already present on Transformer, at the end I am not sure if that's better idea TBH.

The real problem is that for this case it would be convenient to not rely on the object provided by parameter and be able to return the new object I want to use, but that's not implemented on SonataAdminBundle (and is a BC break, so we can't do it)

src/Admin/SnapshotAdmin.php
```diff
@@ -30,13 +33,67 @@ final class SnapshotAdmin extends AbstractAdmin
...
+        $object->setEnabled($snapshot->getEnabled());
+        $object->setRouteName($snapshot->getRouteName());
+        $object->setPageAlias($snapshot->getPageAlias());
+        $object->setType($snapshot->getType());
+        $object->setName($snapshot->getName());
+        $object->setPosition($snapshot->getPosition());
+        $object->setDecorate($snapshot->getDecorate());
+        $object->setSite($snapshot->getSite());
+        $object->setParentId($snapshot->getParentId());
+        $object->setContent($snapshot->getContent());
```
jordisala1991: I do not like this a lot, but not sure is a better way to do this... Have talked to @VincentLanglet already, but that's a missing feature from the AdminBundle. I can just do set on the already given object.

src/Admin/SnapshotAdmin.php
```diff
@@ -55,6 +112,7 @@ protected function configureDatagridFilters(DatagridMapper $filter): void
     protected function configureFormFields(FormMapper $form): void
     {
         $form
+            ->add('page')
```
jordisala1991: This is not ideal, I would like to hide this field... need to think more about this.

src/Controller/SnapshotAdminController.php
```diff
@@ -35,51 +30,10 @@ final class SnapshotAdminController extends CRUDController
     public static function getSubscribedServices(): array
     {
         return [
-            'sonata.page.manager.page' => PageManagerInterface::class,
             'sonata.page.manager.snapshot' => SnapshotManagerInterface::class,
-            'sonata.page.service.create_snapshot' => CreateSnapshotService::class,
         ] + parent::getSubscribedServices();
     }
 
-    public function createAction(Request $request): Response
```
jordisala1991, line 44: this create was trying to replace the original, which causes issues like this one: https://github.com/sonata-project/SonataPageBundle/issues/1453

cc/ @eerison 

Also this create for some reason use a custom form type, instead of relying on the admin. IMHO this is a bad idea, and that's why I tried to remove it. So we can fix both issues, because this PR also solves the other issue: https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1210757295

src/Admin/SnapshotAdmin.php
Hanmac: i still think we might add a new function to Transformer to handle this

src/Admin/SnapshotAdmin.php
```diff
@@ -30,13 +33,67 @@ final class SnapshotAdmin extends AbstractAdmin
...
+
+        $object->setPage($this->pageManager->find($pageId));
+    }
+
+    protected function prePersist(object $object): void
+    {
+        $page = $this->isChild() ? $this->getParent()->getSubject() : $object->getPage();
+        \assert(null === $page || $page instanceof PageInterface);
+
+        if (null === $page) {
```
VincentLanglet: You could change the check by if !instanceof PageInterface

src/Admin/SnapshotAdmin.php
jordisala1991: Already answered about that, it will not solve the issue, just hide it somewhere else.

VincentLanglet: The method create does
```
public function create(PageInterface $page): SnapshotInterface
    {
        $snapshot = $this->snapshotManager->create();
```

If we do 
```
public function create(PageInterface $page, ?SnapshotInterface snapshot): SnapshotInterface
    {
        $snapshot ??= $this->snapshotManager->create();
```
it avoid you to set everything.

(And then we should rename the method) ?

@jordisala1991

jordisala1991: I will give a try to modify the create method.

jordisala1991: It should be better now.

src/Admin/SnapshotAdmin.php
jordisala1991: I have changed code.

VincentLanglet (approved):

PR #1552, opened 2022-08-11
Add more CRUD form tests

jordisala1991:
## Subject

There are a ton of errors to fix :)


src/Block/SharedBlockBlockService.php
```diff
@@ -154,8 +154,12 @@ private function getBlockBuilder(): FormBuilderInterface
             'translation_domain' => 'SonataPageBundle',
             'edit' => 'list',
         ]);
+        $fieldDescription->setAssociationAdmin($this->sharedBlockAdmin);
 
-        return $this->sharedBlockAdmin->getFormBuilder()->create('blockId', ModelListType::class, [
+        return $this->sharedBlockAdmin->getFormContractor()->getFormBuilder(
+            $this->sharedBlockAdmin->getUniqId(),
+            ['data_class' => $this->sharedBlockAdmin->getClass()]
+        )->create('blockId', ModelListType::class, [
```
jordisala1991: This kind of fix should be applied to other bundles. There is still a lot of code to write to do such fields.

There were 2 problems:
1. Field description didn't had association admin, and it break on a template from sonataAdmin, fixed line 157.
2. `getFormBuilder` does a lot of things, it instantiate al the form for that admin, and it needs a subject, so it breaks. I picked the first lines where it creates de form builder. There are other options to fix this too. One of them is use the original $form from above, but we are wrapping it on a PageFormMapper (For no reason). 

IMO we might refactor this further, and remove PageFormMapper too, it is not needed.

jordisala1991:
It should work with second commit, but the solution I don't like it a lot, AND I need to add more test because other kind of blocks are broken too.

Edit: nice it works on my computer, but not on github actions, for some reason, will need to take a deeper look.

VincentLanglet:
> Edit: nice it works on my computer, but not on github actions, for some reason, will need to take a deeper look.

You have a great computer :p

tests/Functional/Admin/SharedBlockAdminTest.php
```diff
@@ -45,11 +45,151 @@ public function testCrudUrls(string $url, array $parameters = []): void
...
+
+    /**
+     * @return iterable<array<string|array<string, mixed>>>
+     *
+     * @phpstan-return iterable<array{0: string, 1: array<string, mixed>, 2: string, 3?: array<string, mixed>}>
+     */
+    public static function provideFormUrlsCases(): iterable
+    {
+        // Blocks From SonataBlockBundle
+        // yield 'Create Shared Block - Container' => ['/admin/tests/app/sonatapageblock/shared/create', [
```
jordisala1991, line 85: This one is broken on SonataBlockBundle.

tests/Functional/Admin/SharedBlockAdminTest.php
```diff
@@ -45,11 +45,151 @@ public function testCrudUrls(string $url, array $parameters = []): void
...
+            'shared_block[enabled]' => 1,
+            'shared_block[settings][title]' => 'Title',
+            'shared_block[settings][translation_domain]' => 'SonataPageBundle',
+            'shared_block[settings][icon]' => 'fa fa-home',
+            'shared_block[settings][current]' => 1,
+            // 'shared_block[settings][pageId]' => 1,
+            'shared_block[settings][class]' => 'custom_class',
+        ]];
+
+        // yield 'Create Shared Block - Breadcrumb' => ['/admin/tests/app/sonatapageblock/shared/create', [
```
jordisala1991, line 161: This one is broken on SonataSeoBundle.

tests/Functional/Admin/SharedBlockAdminTest.php
```diff
@@ -45,11 +45,151 @@ public function testCrudUrls(string $url, array $parameters = []): void
...
+            'uniqid' => 'shared_block',
+            'type' => 'sonata.page.block.children_pages',
+        ], 'btn_create_and_list', [
+            'shared_block[name]' => 'Name',
+            'shared_block[enabled]' => 1,
+            'shared_block[settings][title]' => 'Title',
+            'shared_block[settings][translation_domain]' => 'SonataPageBundle',
+            'shared_block[settings][icon]' => 'fa fa-home',
+            'shared_block[settings][current]' => 1,
+            // 'shared_block[settings][pageId]' => 1,
```
VincentLanglet, line 157: Should we remove this line or uncomment it ?

tests/Functional/Admin/SharedBlockAdminTest.php
VincentLanglet: Can it be fix ?

tests/Functional/Admin/SharedBlockAdminTest.php
VincentLanglet: Can it be fix ?

src/Block/SharedBlockBlockService.php
VincentLanglet: What's the issue ? And in which bundle should we apply the same fix ?

It was never reported before

I would say that PageFormMapper was maybe needed prior to https://github.com/sonata-project/SonataBlockBundle/commit/cd2f8f3ad22f377a6ee542c9c040a30d6795c261. But now the AdminFormMapper implements the BlockBundle FormMapper it's not needed.

Does removing this class (in this PR) would simplify your bugfix ?

jordisala1991: The problem is:

1. Some of our bundles define blocks (and they do require block bundle) 👍 
2. Some of that blocks are EditableBlocks (an interface defined on block bundle) 👍 
3. That interface is only testable HERE, because those editableBlocks are for SonataPageBundle 👎 
4. When we move for example SonataMediaBundle from 3.x to 4.x we migrated those blocks to newer SonataBlock version, but we didn't test those because of point 3. (and we didn't try to unit test AFAIK, will need to check) 👎 
5. Now that we are migrating this bundle, some errors appear.

Those are other examples of the same problem:
https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Block/MediaBlockService.php#L210-L226

I don't know if there are a lot of them (I think not), but this kind of dependencies are hard to test.

tests/Functional/Admin/SharedBlockAdminTest.php
jordisala1991: Yes, we need to remove this field: https://github.com/sonata-project/SonataBlockBundle/blob/97cc8c77d4c87203e2cfd052c0853ef932239955/src/Block/Service/ContainerBlockService.php#L37

tests/Functional/Admin/SharedBlockAdminTest.php
jordisala1991: This one is a bit problematic... Take a look at the class:

2.x: https://github.com/sonata-project/SonataSeoBundle/blob/2.x/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L29

Here the class extends MenuBlockService which implements EditableBlockService 👍 

3.x: https://github.com/sonata-project/SonataSeoBundle/blob/3.x/src/Block/Breadcrumb/BaseBreadcrumbMenuBlockService.php#L29

Here the class does not extend MenuBlockService, and does not implement EditableBlockService 👎 

To fix this we will need to BC break, but it should be accepted IMO, because that block is completely unusable.

jordisala1991: Maybe @eerison or other users of this bundle could help us with it. IMHO we should not block this PR for it, because it will need some thinking, we can create an issue or something.

tests/Functional/Admin/SharedBlockAdminTest.php
jordisala1991: There is a pageId on the settings, but not sure how to make it populate with pages, because it needs a page on the block. I leave it commented to not forget to think about it...

tests/Functional/Admin/SharedBlockAdminTest.php
VincentLanglet: > To fix this we will need to BC break, but it should be accepted IMO, because that block is completely unusable.

Yeah, I agree

> we can create an issue or something.

Can you create an issue with the list of things to fix then ? Then we'll merge this one.

src/Block/SharedBlockBlockService.php
jordisala1991: Take a look at my last commit, not using PageFormMapper, allows to \assert for AdminFormMapper and we have getFormBuilder there.

tests/Functional/Admin/SharedBlockAdminTest.php
eerison: > Maybe @eerison or other users of this bundle could help us with it. IMHO we should not block this PR for it, because it will need some thinking, we can create an issue or something.

Create an issue to check it's fine!

But this `breadcrumb menu` is a bit trick! I'm still struggling with this in #1497  😢

jordisala1991: One for this: https://github.com/sonata-project/SonataSeoBundle/issues/685

src/Block/SharedBlockBlockService.php
```diff
@@ -134,18 +135,17 @@ public function load(BlockInterface $block): void
...
             'translation_domain' => 'SonataPageBundle',
             'edit' => 'list',
         ]);
         $fieldDescription->setAssociationAdmin($this->sharedBlockAdmin);
 
-        return $this->sharedBlockAdmin->getFormContractor()->getFormBuilder(
-            $this->sharedBlockAdmin->getUniqId(),
-            ['data_class' => $this->sharedBlockAdmin->getClass()]
-        )->create('blockId', ModelListType::class, [
+        \assert($form instanceof AdminFormMapper);
```
VincentLanglet: I don't like the assert here. 

Nothing is preventing
```
configureEditForm
```
to be called with another FormMapper.

I would change this too
```
private function getBlockBuilder(AdminFormMapper $form): FormBuilderInterface
```
and
```
public function configureEditForm(FormMapper $form, BlockInterface $block): void
    {
        if (!$form instanceof AdminFormMapper) {
             throw new \InvalidArgumentException();
        }
```
(And we can consider later to make the EditableBlockService generics)

VincentLanglet (changes requested):

tests/Functional/Admin/SharedBlockAdminTest.php
jordisala1991: And this PR: https://github.com/sonata-project/SonataBlockBundle/pull/1098

src/Block/SharedBlockBlockService.php
jordisala1991: What do you think now?

VincentLanglet (approved):

PR #1556, opened 2022-08-12
Add test for update disposition

jordisala1991:
## Subject

I think this one is broken too...

### DIFF
--- a/src/Controller/BlockAdminController.php
+++ b/src/Controller/BlockAdminController.php
@@ -16,7 +16,6 @@ namespace Sonata\PageBundle\Controller;
 use Sonata\AdminBundle\Controller\CRUDController;
 use Sonata\AdminBundle\Exception\BadRequestParamHttpException;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
-use Sonata\PageBundle\Admin\PageAdmin;
 use Sonata\PageBundle\Model\BlockInteractorInterface;
 use Sonata\PageBundle\Model\PageBlockInterface;
 use Symfony\Component\HttpFoundation\Request;
@@ -36,7 +35,6 @@ final class BlockAdminController extends CRUDController
     {
         return [
             'sonata.page.block_interactor' => BlockInteractorInterface::class,
-            'sonata.page.admin.page' => PageAdmin::class,
             'sonata.block.manager' => BlockServiceManagerInterface::class,
         ] + parent::getSubscribedServices();
     }
@@ -56,12 +54,8 @@ final class BlockAdminController extends CRUDController
                 throw new HttpException(400, 'wrong parameters');
             }
 
-            $result = $this->container->get('sonata.page.block_interactor')->saveBlocksPosition($params, false);
+            $result = $this->container->get('sonata.page.block_interactor')->saveBlocksPosition($params);
             $status = 200;
-
-            $pageAdmin = $this->container->get('sonata.page.admin.page');
-            $pageAdmin->setRequest($request);
-            $pageAdmin->update($pageAdmin->getSubject());
         } catch (HttpException $e) {
             $status = $e->getStatusCode();
             $result = [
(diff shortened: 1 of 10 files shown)


jordisala1991:
Even if test pass still broken, positions are not updated. Will fix later.

jordisala1991:
Wait, do not fully review yet.

src/Entity/BlockManager.php
```diff
@@ -25,42 +24,17 @@
  */
 final class BlockManager extends BaseEntityManager implements BlockManagerInterface
 {
-    public function updatePosition($id, int $position, $parentId = null, $pageId = null, bool $partial = true): PageBlockInterface
+    public function updatePosition($id, int $position): PageBlockInterface
     {
-        $entityManager = $this->getEntityManager();
+        $block = $this->find($id);
 
-        if ($partial) {
```
jordisala1991, line 32: This partial does not make sense to me, there is only one place this method is used, and it always gets false....

@VincentLanglet IMO we should remove it, it even mentions performance problems when is true 😓

src/Controller/BlockAdminController.php
```diff
@@ -49,19 +47,15 @@ public function savePositionAction(Request $request): Response
...
             }
 
-            $result = $this->container->get('sonata.page.block_interactor')->saveBlocksPosition($params, false);
-
+            $result = $this->container->get('sonata.page.block_interactor')->saveBlocksPosition($params);
             $status = 200;
-
-            $pageAdmin = $this->container->get('sonata.page.admin.page');
-            $pageAdmin->setRequest($request);
-            $pageAdmin->update($pageAdmin->getSubject());
```
jordisala1991, line 64: If I don't remove this, this method only make sense in the context of child, but if I remove it, the page itself won't get the edited set to true.

I guess I will have to recover those lines, but that means I have to limit the usage of this route to child of blocks, do you agree @VincentLanglet ?

Otherwise we could move the setEdited(true) to the prePersist of block admin (it is now on the blockAdmin, but that does not get triggered when saving through modelManager).

What to do, @VincentLanglet . I like to remove those 3 lines here, and add on prePersist and preUpdate, will try that way, and wait for your feedback.

src/Controller/BlockAdminController.php
```diff
@@ -49,19 +47,15 @@ public function savePositionAction(Request $request): Response
         $this->admin->checkAccess('savePosition');
 
         try {
-            $params = $request->get('disposition');
+            // TODO: Change to $request->query->all('filter') when support for Symfony < 5.1 is dropped.
+            $params = $request->request->all()['disposition'] ?? [];
 
-            if (!\is_array($params)) {
+            if ([] === $params) {
```
jordisala1991, line 53: This was completely broken before.

src/Entity/BlockInteractor.php
```diff
@@ -67,18 +67,21 @@ public function getBlocksById(PageInterface $page): array
...
-    public function saveBlocksPosition(array $data = [], bool $partial = true): bool
+    public function saveBlocksPosition(array $data = []): bool
     {
         $em = $this->getEntityManager();
         $em->getConnection()->beginTransaction();
 
         try {
             foreach ($data as $block) {
-                if (!isset($block['id'], $block['position'], $block['parent_id'], $block['page_id'])) {
+                if (!isset($block['id'], $block['position']) || !is_numeric($block['position'])) {
```
jordisala1991, line 77: This was broken too, because always wanted 4 parameters. It was due an update of phpstan.

src/Entity/BaseBlock.php
```diff
@@ -23,11 +23,18 @@ abstract class BaseBlock extends Block
     public function prePersist(): void
     {
         $this->createdAt = new \DateTime();
-        $this->updatedAt = new \DateTime();
+
+        $this->preUpdate();
```
VincentLanglet: I dunno if it's not better to duplicate the logic.

If someone override the preUpdate method
```
public function preUpdate(): void
{
    parent::preUpdate();

    // do things
}
```
he may not want to execute them in the prePersist too.

VincentLanglet (changes requested):

src/Entity/BaseBlock.php
jordisala1991: Changed.

VincentLanglet (approved):

PR #1558, opened 2022-08-14
Add test for Page render being an editor

jordisala1991:
## Subject

Adding more tests, those ones are the first that do not modify coverage, but still are useful tests, and will improve in next PRs, trying to render all kind of block types.

### DIFF
--- a/tests/Functional/Admin/PageAdminTest.php
+++ b/tests/Functional/Admin/PageAdminTest.php
@@ -171,9 +171,19 @@ final class PageAdminTest extends WebTestCase
         yield 'Remove Snapshot Page' => ['/admin/tests/app/sonatapagesnapshot/1/delete', [], 'btn_delete'];
 
         // Block child pages
-        yield 'Create Block Page' => ['/admin/tests/app/sonatapagepage/1/sonatapageblock/create', [
-            'type' => 'sonata.block.service.text',
-        ], 'btn_create_and_list', []];
+        yield 'Create Block Page - Children Pages' => ['/admin/tests/app/sonatapagepage/1/sonatapageblock/create', [
+            'uniqid' => 'block',
+            'type' => 'sonata.page.block.children_pages',
+        ], 'btn_create_and_list', [
+            'block[name]' => 'Name',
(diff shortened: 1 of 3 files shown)


tests/Functional/Admin/SharedBlockAdminTest.php
```diff
@@ -144,20 +144,6 @@ public static function provideFormUrlsCases(): iterable
...
-            'uniqid' => 'shared_block',
-            'type' => 'sonata.page.block.children_pages',
-        ], 'btn_create_and_list', [
-            'shared_block[name]' => 'Name',
-            'shared_block[enabled]' => 1,
-            'shared_block[settings][title]' => 'Title',
-            'shared_block[settings][translation_domain]' => 'SonataPageBundle',
-            'shared_block[settings][icon]' => 'fa fa-home',
-            'shared_block[settings][current]' => 1,
-            // 'shared_block[settings][pageId]' => 1,
```
jordisala1991, line 157: Fixed this one, This block currently cannot be used as a shared block, in order to be able to select a pageId you need to be in the context, of the blockAdmin as a child of the pageAdmin

tests/Functional/Admin/PageAdminTest.php
```diff
@@ -171,9 +171,19 @@ public static function provideFormUrlsCases(): iterable
         yield 'Remove Snapshot Page' => ['/admin/tests/app/sonatapagesnapshot/1/delete', [], 'btn_delete'];
 
         // Block child pages
-        yield 'Create Block Page' => ['/admin/tests/app/sonatapagepage/1/sonatapageblock/create', [
-            'type' => 'sonata.block.service.text',
-        ], 'btn_create_and_list', []];
+        yield 'Create Block Page - Children Pages' => ['/admin/tests/app/sonatapagepage/1/sonatapageblock/create', [
```
jordisala1991, line 174: Here we will only add block creation that can be only tested this way, so we don't repeat too many test for the same thing.

VincentLanglet (approved):

PR #1559, opened 2022-08-15
Modernise frontend code

jordisala1991:
## Subject

## Changelog

```markdown
### Changed
- Change frontend compilation system from Gulp to Webpack
- Change final minified file to `app.js` and `app.css`
```


jordisala1991:
Your latest PR on admin bundle might have break some builds @VincentLanglet

VincentLanglet:
> Your latest PR on admin bundle might have break some builds @VincentLanglet

I'm using the modelManager in some methods now, so if it's not set in unit test it will fail indeed.

VincentLanglet:
https://github.com/sonata-project/SonataPageBundle/pull/1560

jordisala1991:
Two things fixed here (besides the full migrations and upgrade of the code to be a little better):

* Composer page now kinda works (not it does not fail on the js part, but rather when saving blocks on the PHP part, to see on another PR)
* Layout is a bit better on Sonata, before it was a bit broken on some parts.

cc @eerison

VincentLanglet (approved):

VincentLanglet:
Nice job

PR #1563, opened 2022-08-16
Fix block container children

jordisala1991:
## Subject

This together with: https://github.com/sonata-project/SonataBlockBundle/pull/1100 makes the children blocks work on the edit block (the composer page still needs some work).

### DIFF
--- a/src/Admin/BaseBlockAdmin.php
+++ b/src/Admin/BaseBlockAdmin.php
@@ -84,7 +84,11 @@ abstract class BaseBlockAdmin extends AbstractAdmin
 
     protected function alterNewInstance(object $object): void
     {
-        $object->setType($this->getPersistentParameter('type'));
+        $type = $this->getPersistentParameter('type');
+
+        if (null !== $type) {
+            $object->setType($type);
+        }
 
         $this->loadBlockDefaults($object);
     }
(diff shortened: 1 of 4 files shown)


src/Admin/BaseBlockAdmin.php
```diff
@@ -84,7 +84,11 @@ protected function alterObject(object $object): void
 
     protected function alterNewInstance(object $object): void
     {
-        $object->setType($this->getPersistentParameter('type'));
+        $type = $this->getPersistentParameter('type');
+
+        if (null !== $type) {
+            $object->setType($type);
```
jordisala1991, line 90: We need to only do the setType if we are sure is a string, because of the signature:

```
    public function setType(string $type): void
    {
        $this->type = $type;
    }
```

TBH this is happening a lot, considering that model is meant to be an entity, IMHO it would make much more sense to treat it like that and make all fields that starts with a null value accept a null one (Like we are doing on each bundle that has an entity).

But I guess we can deal with that in another PR, wdyt  @VincentLanglet ?

src/Admin/BlockAdmin.php
```diff
@@ -95,10 +95,6 @@ protected function configureRoutes(RouteCollectionInterface $collection): void
         $collection->add('save_position', 'save-position');
         $collection->add('switch_parent', 'switch-parent');
         $collection->add('compose_preview', $this->getRouterIdParameter().'/compose-preview');
-
-        if (!$this->isChild()) {
-            $collection->remove('create');
```
jordisala1991, line 100: If we remove the create route, no new rows can be added on the collection. (the add button does not show).

jordisala1991:
@VincentLanglet This one can be reviewed, it fixes the add button on the container page children.

VincentLanglet (approved):

PR #1566, opened 2022-08-20
Make composer page usable again

jordisala1991:
## Subject

Fixing a few bugs that are only present on 4.x since we changed a lot of code.


src/Admin/BlockAdmin.php
```diff
@@ -137,10 +137,10 @@ protected function configureFormFields(FormMapper $form): void
 
         $form->with('general', $generalGroupOptions);
 
-        if (false === $isComposer) {
-            $form->add('name');
-        } else {
+        if (false !== $isComposer) {
```
jordisala1991, line 140: We might want to change this again in the future, the composer query variable arrives with "1" when is set, and it gets "false" default value when its not.

I think it is safer this way, for now atleast.

src/Admin/BlockAdmin.php
```diff
@@ -154,7 +154,7 @@ protected function configureFormFields(FormMapper $form): void
             $containerBlockTypes = $this->containerBlockTypes;
 
             // need to investigate on this case where $page == null ... this should not be possible
-            if ($isStandardBlock && null !== $page && [] === $containerBlockTypes) {
+            if ($isStandardBlock && null !== $page && [] !== $containerBlockTypes) {
```
jordisala1991, line 157: I made a mistake when applying phpstan changes, this should have been an !== since the begining. This fixes the relation between parents and childs on composer page.

src/Controller/PageAdminController.php
```diff
@@ -253,16 +253,19 @@ public function composeContainerShowAction(Request $request): Response
         // filter service using the template configuration
         if (null !== $page) {
             $template = $this->container->get('sonata.page.template_manager')->get($page->getTemplateCode());
+            $code = $block->getSetting('code');
 
-            $container = $template->getContainer($block->getSetting('code'));
+            if (null !== $code) {
```
jordisala1991: `getContainer` function does not acept null, we need to protect with an if, because getSetting can return null.

src/Controller/PageAdminController.php
```diff
@@ -253,16 +253,19 @@ public function composeContainerShowAction(Request $request): Response
...
+                $container = $template->getContainer($code);
 
-            if (isset($container['blocks']) && \count($container['blocks']) > 0) {
-                foreach ($blockServices as $code => $service) {
-                    if (\in_array($code, $container['blocks'], true)) {
-                        continue;
-                    }
+                if (null !== $container) {
```
jordisala1991, line 261: better to chekc with null, $container is an array (with 'blocks' set) or null

src/Resources/views/BlockAdmin/compose_preview.html.twig
```diff
@@ -33,10 +33,26 @@
...
+            data-label-enable="{{ ('composer.enable'|trans({}, 'SonataPageBundle') ~ ' <i class="fa fa-toggle-on"></i>')|e }}"
+            data-label-disable="{{ ('composer.disable'|trans({}, 'SonataPageBundle') ~ ' <i class="fa fa-toggle-off"></i>')|e }}"
+        >
+            <a
+                class="badge bg-{{ child.enabled ? 'yellow' : 'green' }}"
+                href="{{ path('sonata_admin_set_object_field_value', {
+                    'objectId': child.id,
+                    'context': 'list',
+                    'field': 'enabled',
+                    '_sonata_admin': 'sonata.page.admin.block'
```
jordisala1991, line 47: I fixed indentation to be more readable, but also there is a bugfix here, before we used "code" query parameter, on Sonata 4 it should be "_sonata_admin". Otherwise you can't toggle publish of blocks.

assets/js/composer.js
```diff
@@ -507,7 +507,7 @@ PageComposer.prototype = {
         data: $form.serialize(),
         type: method,
         headers: {
-          Accept: 'text/html, application/xhtml+xml;',
+          Accept: 'application/json',
```
jordisala1991, line 510: This fixes an error on edit block and clicking to save. (I fixed this on another place in this js, and forgot to fix this one aswell)

VincentLanglet:
There is no test related to all these changes ?

jordisala1991:
All this changes would require to setup panther tests to see something fixed I guess.

src/Admin/BlockAdmin.php
jordisala1991: This one could be tested with our current suite

src/Admin/BlockAdmin.php
jordisala1991: This one can also be covered with test with our current suite (but a panther test would do it bettrr

src/Controller/PageAdminController.php
jordisala1991: We could have more test for the compose page (but panther test would do it better since we could test js behavior)

src/Resources/views/BlockAdmin/compose_preview.html.twig
jordisala1991: Cant be really tested without panther test

assets/js/composer.js
jordisala1991: Cant be really tested without pnther tests

jordisala1991:
I left some comment on where we could add test and where it would be better to setup some panther testing.

src/Admin/BlockAdmin.php
jordisala1991: This is now covered.

src/Admin/BlockAdmin.php
jordisala1991: This is now covered.

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #1567, opened 2022-08-20
Increase coverage

jordisala1991:
Some more coverage for previous changes.

### DIFF
--- a/tests/Functional/Admin/BlockAdminTest.php
+++ b/tests/Functional/Admin/BlockAdminTest.php
@@ -249,15 +249,16 @@ final class BlockAdminTest extends WebTestCase
 
         $parentBlock = new SonataPageBlock();
         $parentBlock->setType('sonata.page.block.container');
-        $parentBlock->setPage($page);
 
         $parentBlock2 = new SonataPageBlock();
         $parentBlock2->setType('sonata.page.block.container');
-        $parentBlock2->setPage($page);
 
         $block = new SonataPageBlock();
         $block->setType('sonata.block.service.text');
         $block->setParent($parentBlock);
+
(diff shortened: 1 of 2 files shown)


tests/Functional/Admin/BlockAdminTest.php
```diff
@@ -249,15 +249,16 @@ private function prepareData(): void
...
         $parentBlock2->setType('sonata.page.block.container');
-        $parentBlock2->setPage($page);
 
         $block = new SonataPageBlock();
         $block->setType('sonata.block.service.text');
-        $block->setParent($parentBlock);
+
+        $page->addBlock($parentBlock);
+        $page->addBlock($parentBlock2);
+        $page->addBlock($block);
```
jordisala1991: Is better to add blocks like this, otherwise they don't get related on the $page side, I think it is because how doctrine works, maybe the object is not refreshed, anyway, better do it this way.

tests/Functional/Admin/PageAdminTest.php
```diff
@@ -262,12 +262,14 @@ private function prepareData(): void
 
         $parentBlock = new SonataPageBlock();
         $parentBlock->setType('sonata.page.block.container');
-        $parentBlock->setPage($page);
+        $parentBlock->setSetting('code', 'content');
```
jordisala1991, line 265: with this we can see what happens when a block is not orphan and is related to a section of the page.

tests/Functional/Admin/BlockAdminTest.php
VincentLanglet: When you add, it also call the other setter. But when you set, it doesn't.

VincentLanglet (approved):

PR #1569, opened 2022-08-21
Clear cache

jordisala1991:
## Subject

## Changelog

```markdown
### Added

### Changed

### Deprecated

### Removed

### Fixed

### Security
```


phpunit.xml.dist
```diff
@@ -37,5 +37,8 @@ It's auto-generated by sonata-project/dev-kit package.
     <php>
         <ini name="precision" value="8" />
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+        <env name="KERNEL_CLASS" value="\Sonata\PageBundle\Tests\App\AppKernel" />
+        <server name="APP_ENV" value="test" force="true" />
+        <server name="APP_DEBUG" value="false" />
```
VincentLanglet, line 42: This file is auto generated ; do you have something in mind ?

jordisala1991:
I still need to understand why coverage does not increase, it does not make any sense tbh. there is some files on dependency injection (the extension class) that somehow is not covered.

jordisala1991:
Still, this would improve by a LOT the DX on Sonata. And it would make all test run independent from changes on config and all (previously I had to manually edit the cache folder to get a clean build)

phpunit.xml.dist
jordisala1991: Yep, the same we generate the console file, we can generate the namespace for the AppKernel file.

jordisala1991:
There are 3 things here:
1. use variables to define kernel class like in normal sf projects (and avoid extra code to define on each test case which kernel to use)
2. clear cache to avoid broken test when doing changes to code.
3. if possible, see why the coverage does not increase.

phpunit.xml.dist
```diff
@@ -37,5 +37,8 @@ It's auto-generated by sonata-project/dev-kit package.
     <php>
         <ini name="precision" value="8" />
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+        <env name="KERNEL_CLASS" value="\Sonata\PageBundle\Tests\App\AppKernel" />
```
eerison, line 40: hey @jordisala1991  I tried to do this before But @VincentLanglet mentioned  that this file is generated automatically

VincentLanglet: He plans to update the dev-kit file generator

jordisala1991:
Now it reflects the real coverage, and we dont have issues about regenerating cache. wdyt @VincentLanglet ?

VincentLanglet (commented):
Seems nice. We will have to update all the projects ?

jordisala1991:
I think most of them yes, dev kit will handle part of the update and we will have to do the rest.

jordisala1991:
This will be dev kit PR: https://github.com/sonata-project/dev-kit/pull/2154

Keep in mind that other projects might not benefit from increase coverage, but all of them should get the other benefits (remove the getkernelclass funct and avoid problems with cache)

eerison:
why was it closed 👀

jordisala1991:
Because it is already done (not merged on all branches, but on the stable ones. I will merge this one manually since it is probably the one which will cause more conflicts.

PR #1570, opened 2022-08-23
Minor improvements

VincentLanglet:


eerison:
@VincentLanglet I have created this PR: https://github.com/sonata-project/SonataPageBundle/pull/1571

just to help you with the page change for layout!

after you merge this, you can close this #1571

eerison:
Please apply those 2 changes

https://github.com/sonata-project/SonataPageBundle/pull/1571/files#diff-f0461df8f4a53195f2b565ad37746df1e8076ff748edb18c3e2689581f2db4caR86

https://github.com/sonata-project/SonataPageBundle/pull/1571/files#diff-99821531fe8569e4745708ce6f87e337c82557391aaa9927f3c737260fae004bR22

src/Controller/PageAdminController.php
```diff
@@ -237,37 +257,50 @@ public function composeAction(Request $request): Response
...
 
-                if (null !== $container) {
-                    foreach ($blockServices as $code => $service) {
-                        if (\in_array($code, $container['blocks'], true)) {
-                            continue;
-                        }
+            $template = $templateManager->get($page->getTemplateCode());
+            if (null === $template) {
+                throw new NotFoundHttpException(
+                    sprintf('unable to find the template with code : %s', $templateCode)
```
jordisala1991:
```suggestion
                    sprintf('Unable to find the template with code : %s', $templateCode)
```

VincentLanglet: Done

src/Block/BreadcrumbBlockService.php
```diff
@@ -83,7 +83,7 @@ protected function getMenu(BlockContextInterface $blockContext): ItemInterface
 
     private function getName(): string
     {
-        return 'sonata.page.block.breadcrumb';
+        return 'page';
```
jordisala1991: IMO this is not the correct change. The metadata can return the previous value and the handleContext can check against page, and we can remove this method.

VincentLanglet: I reverted.

I didn't follow this dicussion so prefer to let @eerison doing another PR about it.

jordisala1991 (approved):

src/Block/BreadcrumbBlockService.php
eerison: well but this values need to be equals passed by template!

if we use something different of `page`, then we still have the same problem to use `sonata.page.block.breadcrumb`

what is your suggestion for this @jordisala1991 ?

PR #1572, opened 2022-08-23
Fix breadcrumb context

jordisala1991:
## Subject

I am targeting this branch, because this fixes unwanted BC break on breadcrumb context.

### DIFF
--- a/src/Block/BreadcrumbBlockService.php
+++ b/src/Block/BreadcrumbBlockService.php
@@ -42,14 +42,14 @@ final class BreadcrumbBlockService extends BaseBreadcrumbMenuBlockService
 
     public function getMetadata(): MetadataInterface
     {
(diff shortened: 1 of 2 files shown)


jordisala1991:
This is what I had in mind @eerison @VincentLanglet .

src/Block/BreadcrumbBlockService.php
```diff
@@ -42,14 +42,14 @@ public function __construct(
 
     public function getMetadata(): MetadataInterface
     {
-        return new Metadata($this->getName(), null, null, 'SonataPageBundle', [
+        return new Metadata('sonata.page.block.breadcrumb', null, null, 'SonataPageBundle', [
```
eerison, line 45: Hmmm well it works too 😅😅😅

I don't know if it's the best approach or not.

I don't see any problem to use like you did

VincentLanglet (approved):

PR #1574, opened 2022-08-26
Remove show route for blocks

jordisala1991:
## Subject

I am targeting this branch, because this is in preparation for 4.0

## Changelog

```markdown
### Removed
- Remove show route for block since blocks are meant to be edited through compose page.
```

### DIFF
--- a/src/Admin/SharedBlockAdmin.php
+++ b/src/Admin/SharedBlockAdmin.php
@@ -59,7 +59,7 @@ final class SharedBlockAdmin extends BaseBlockAdmin
     protected function configureListFields(ListMapper $list): void
     {
         $list
-            ->addIdentifier('name')
+            ->addIdentifier('name', null, ['route' => ['name' => 'edit']])
             ->add('type')
             ->add('enabled', null, ['editable' => true])
             ->add('updatedAt');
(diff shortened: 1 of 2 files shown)


jordisala1991:
@eerison does it work for you? now you should get directly to edit page

VincentLanglet:
> @eerison does it work for you? now you should get directly to edit page

I'm not sure about this. I can't remember if it will redirect to the edit page or just doesn't show any links.

jordisala1991:
I think it will now show any links if you configure the show route as default. I guess we need to change both things

jordisala1991:
IMHO both changes are correct, it does not make a lot of sense to expose the show route if we didnt configure any fields to show.

eerison:
Hey @jordisala1991 It's going to edit page :)

src/Admin/SharedBlockAdmin.php
```diff
@@ -59,7 +59,7 @@ protected function configureQuery(ProxyQueryInterface $query): ProxyQueryInterfa
     protected function configureListFields(ListMapper $list): void
     {
         $list
-            ->addIdentifier('name')
+            ->addIdentifier('name', null, ['route' => ['name' => 'edit']])
```
VincentLanglet, line 62: I don't think it's a bad thing to declare the route manually because if not, it rely on the default config of sonata admin. So it could be weird if the user use another default route for all his admins.

VincentLanglet (approved):

PR #1577, opened 2022-08-29
Bug/Composer page is not working

eerison:
## Subject

## Changelog

```markdown
### Added

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

### DIFF
--- a/src/Controller/PageAdminController.php
+++ b/src/Controller/PageAdminController.php
@@ -294,6 +294,7 @@ final class PageAdminController extends CRUDController
             }
             $container = $template->getContainer($blockCode);
 
+            //TODO here is the problem, for some reason blockServices are empty, but in 3.x it isn't!
             if (null !== $container) {
                 foreach ($blockServices as $code => $service) {
                     if (\in_array($code, $container['blocks'], true)) {
--- a/tests/Functional/Admin/PageAdminTest.php
+++ b/tests/Functional/Admin/PageAdminTest.php
@@ -59,7 +59,7 @@ final class PageAdminTest extends WebTestCase
         yield 'Show Page' => ['/admin/tests/app/sonatapagepage/1/show'];
         yield 'Remove Page' => ['/admin/tests/app/sonatapagepage/1/delete'];
         yield 'Compose Page' => ['/admin/tests/app/sonatapagepage/1/compose'];
-        yield 'Compose Show Page' => ['/admin/tests/app/sonatapagepage/compose/container/1'];
+        yield 'Compose Container Show' => ['/admin/tests/app/sonatapagepage/compose/container/1'];
 
         // Snapshot child pages
         yield 'List Snapshot Page' => ['/admin/tests/app/sonatapagepage/1/sonatapagesnapshot/list'];
--- a/tests/Functional/Controller/PageAdminControllerTest.php
+++ b/tests/Functional/Controller/PageAdminControllerTest.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Sonata\PageBundle\Tests\Functional\Controller;
+
+use Doctrine\Persistence\ManagerRegistry;
+use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
+use Sonata\PageBundle\Admin\BlockAdmin;
+use Sonata\PageBundle\Controller\PageAdminController;
+use Sonata\PageBundle\Entity\BlockManager;
+use Sonata\PageBundle\Page\TemplateManagerInterface;
+use Sonata\PageBundle\Tests\App\AppKernel;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\HttpFoundation\Request;
+
+final class PageAdminControllerTest extends WebTestCase
+{
+    public function testComposerContainerShow(): void
+    {
+        $kernel = self::bootKernel();
+        $container = $kernel->getContainer();
+
+        $blockServiceManager = $this->createMock(BlockServiceManagerInterface::class);
+        $managerRegistry = $this->createMock(ManagerRegistry::class);
+
+        $container->set('sonata.page.admin.block', new BlockAdmin($blockServiceManager));
+        $container->set('sonata.block.manager', new BlockManager('foo', $managerRegistry));
+        $container->set('sonata.page.template_manager', $container->get('twig'));
+        $request = $this->createMock(Request::class);
+
+        $pageAdminController = new PageAdminController();
+        $pageAdminController->composeContainerShowAction($request);
+    }
+
+    protected static function getKernelClass(): string
+    {
+        return AppKernel::class;
+    }
+}


src/Controller/PageAdminController.php
```diff
@@ -294,6 +294,7 @@ public function composeContainerShowAction(Request $request): Response
             }
             $container = $template->getContainer($blockCode);
 
+            //TODO is the problem, in 3.x block services have items, but in 4.x it's empty.
```
eerison: Hey @VincentLanglet in this action I basically just need the `getObject`, from [AdminInterface](https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Admin/AdminInterface.php#L221), could we create an interface with this? I guess inject the whole AdminInterface to use only one method is a lot, WDYT?

src/Admin/PageAdmin.php
```diff
@@ -62,6 +62,7 @@ public function __construct(
     protected function configureRoutes(RouteCollectionInterface $collection): void
     {
         $collection->add('compose', $this->getRouterIdParameter().'/compose');
+        //TODO how is it linked with src/Controller/PageAdminController::composeContainerShowAction, where the route is defined?
```
eerison: do you have any idea where this router is defined? `compose_container_show`

Other question: WDYT to  extract this [composeContainerShowAction](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Controller/PageAdminController.php#L257) for other controller like `PageAdminComposeContainerShowController` or just `PageAdminComposeController` with `composeContainerShowAction` and `composerAction`?

jordisala1991:
What is not working? maybe if you add a failing test first it would be easier to see. (I tried the compose page quite a lot and fixed all the errors there)

eerison:
> What is not working? maybe if you add a failing test first it would be easier to see. (I tried the compose page quite a lot and fixed all the errors there)

the problem posted here: https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1223876198

and yes, I'm trying to add a failing test first, to find out what exactly is broken.

src/Admin/PageAdmin.php
VincentLanglet: I would say It's just linked with the url you gave `compose/container/{id}`

src/Controller/PageAdminController.php
VincentLanglet: I don't see a reason to create another interface. There is no issue using AdminInterface

src/Admin/PageAdmin.php
eerison: hmmm I see, but do you have any idea where the route is defined for https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Controller/PageAdminController.php#L257 👀

src/Controller/PageAdminController.php
eerison: Ok it's not a problem to use the AdminInterface, it gonna work, but don't you think the AdminInterface is quite huge 👀

src/Admin/PageAdmin.php
jordisala1991: Here: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Admin/PageAdmin.php#L65

jordisala1991: The first parameter gets converted camelCase and added "Action" at the end, and that's how it knows the name.

src/Controller/PageAdminController.php
```diff
@@ -38,6 +38,18 @@
...
+
+    public function __construct(
+        BlockServiceManagerInterface $blockServiceManager,
+        TemplateManagerInterface $templateManager
+    ) {
+        $this->blockServiceManager = $blockServiceManager;
+        $this->templateManager = $templateManager;
+
+    }
+
```
jordisala1991: IMO, we should not change how we load things inside CRUDController, we use the container, with the function below `getSubscribedServices`

src/Controller/PageAdminController.php
```diff
@@ -256,22 +268,18 @@ public function composeAction(Request $request): Response
      */
     public function composeContainerShowAction(Request $request): Response
     {
-        $blockAdmin = $this->container->get('sonata.page.admin.block');
-        \assert($blockAdmin instanceof BlockAdmin);
-
-        if (false === $blockAdmin->isGranted('LIST')) {
+        //TODO Can I use this directly from symfony controller?
+        if (false === $this->isGranted('LIST')) {
```
jordisala1991: nope, this should remain as using blockAdmin, it generates the permission different. And IMO we should change for $blockAdmin->checkAccess('list') (but I don't see any relation with your bug report tbh, so it should be done in a separate PR).

src/Controller/PageAdminController.php
```diff
@@ -256,22 +268,18 @@ public function composeAction(Request $request): Response
...
             throw new AccessDeniedException();
         }
 
         $id = $request->get($this->admin->getIdParameter());
-        $block = $blockAdmin->getObject($id);
+        $block = $this->admin->getObject($id);
```
jordisala1991: This admin is a pageAdmin, not a block, it will return a page, this change is not correct.

src/Controller/PageAdminController.php
jordisala1991: Yes it is huge, but splitting it is a gigantic work, let's keep things simple for now 🙏

jordisala1991:
> > What is not working? maybe if you add a failing test first it would be easier to see. (I tried the compose page quite a lot and fixed all the errors there)
> 
> the problem posted here: [#1495 (comment)](https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1223876198)
> 
> and yes, I'm trying to add a failing test first, to find out what exactly is broken.

I would say you should revert all the `src` changes, and focus only on `tests` folder, otherwise it would be quite hard to identify the bug.

src/Admin/PageAdmin.php
VincentLanglet: I think it's done with the name instead.

when you call `add->('compose_container_show')` the RouteCollection/RouteGenerator are automatically looking for a `composeContainerShow`
https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/cookbook/recipe_custom_action/

src/Controller/PageAdminController.php
eerison: yeah it's not related with my issue, I'm just trying to make my test simple!

src/Admin/PageAdmin.php
eerison: > I think it's done with the name instead.
> 
> when you call `add->('compose_container_show')` the RouteCollection/RouteGenerator are automatically looking for a `composeContainerShow` https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/cookbook/recipe_custom_action/

hmmmm I guess I got it :D, Thanks :)

src/Controller/PageAdminController.php
eerison: hmmm then don't we inject anything into the controllers' construct 👀

VincentLanglet: We don't inject on the CRUDController because people are extending them, and we let them the constructor free.

Here the class is final so we could inject, but I agree that since it's already has a SubscribedServices we can keep using it for consistency.

src/Controller/PageAdminController.php
VincentLanglet: If you provide failing test when you changed src, we can't know if the test if failing because of the bug or because of your changes. So src code should not be changed.

eerison:
it was solved in other PR :)

PR #1580, opened 2022-08-29
Simplify AppKernel configuration

jordisala1991:
This is a try to see if we can simplify a bit our AppKernel file. It is strange to store caches on a temporal folder outisde the project structure, and makes things a bit obscure.

With this change, the cache could always be on the same place for all projects, and the user can easily see it or manipulate it.

### DIFF
--- a/tests/App/AppKernel.php
+++ b/tests/App/AppKernel.php
@@ -63,16 +63,6 @@ final class AppKernel extends Kernel
         return $bundles;
     }
 
-    public function getCacheDir(): string
-    {
-        return $this->getBaseDir().'cache';
-    }
-
-    public function getLogDir(): string
-    {
-        return $this->getBaseDir().'log';
-    }
-
     public function getProjectDir(): string
     {
         return __DIR__;
@@ -93,8 +83,6 @@ final class AppKernel extends Kernel
      */
     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
     {
-        $container->setParameter('app.base_dir', $this->getBaseDir());
-
         $loader->load($this->getProjectDir().'/config/config.yaml');
 
         if (class_exists(AuthenticatorManager::class)) {
@@ -106,10 +94,5 @@ final class AppKernel extends Kernel
         if (class_exists(HttpCacheHandler::class)) {
             $loader->load($this->getProjectDir().'/config/config_sonata_block_v4.yaml');
         }
-    }
-
-    private function getBaseDir(): string
-    {
-        return sys_get_temp_dir().'/sonata-page-bundle/var/';
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ phpstan.neon
 /.phpunit.result.cache
 /docs/_build/
 npm-debug.log
+tests/App/var
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -9,6 +9,8 @@ parameters:
     paths:
         - src
         - tests
+    excludePaths:
+        - tests/App/var
     treatPhpDocTypesAsCertain: false
     checkGenericClassInNonGenericObjectType: true
     checkInternalClassCaseSensitivity: true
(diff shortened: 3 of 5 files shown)


tests/App/AppKernel.php
```diff
@@ -107,9 +95,4 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
             $loader->load($this->getProjectDir().'/config/config_sonata_block_v4.yaml');
         }
     }
-
-    private function getBaseDir(): string
-    {
-        return sys_get_temp_dir().'/sonata-page-bundle/var/';
```
jordisala1991, line 113: Never liked to use this `sys_get_temp_dir`

jordisala1991:
```

  189x: The "sonata.twig.deprecated_template_extension" service is deprecated since sonata-project/twig-extensions 1.4.
    30x in SharedBlockAdminTest::testFormsUrls from Sonata\PageBundle\Tests\Functional\Admin
    30x in PageAdminTest::testFormsUrls from Sonata\PageBundle\Tests\Functional\Admin
    19x in PageAdminTest::testCrudUrls from Sonata\PageBundle\Tests\Functional\Admin
    12x in SiteAdminTest::testFormsUrls from Sonata\PageBundle\Tests\Functional\Admin
    12x in SnapshotAdminTest::testFormsUrls from Sonata\PageBundle\Tests\Functional\Admin
    ...
```

For some reason this deprecation starts triggering, do you know how to fix it @VincentLanglet ?

Also, do you like this idea?

jordisala1991:
Also, this change can allow one thing:

I normally don't create projects to tests features inside bundles, I use the bundle itself with the functional test and all, it works, but it is not perfect since sometime you want to test the features yourself on the browser. I have added here a small index.php (not commited to the repository) that just loads the test kernel and starts the app. Without this change, for some reason the cache is not shared between tests or between cli commands, so I can't work properly with it (I can't for example, run commands to create database since the temporal file for some reason is different between cli and the php server).

In the future, every Sonata bundle could have a `make try` or whatever name we want to use, that allows to start a local server with that bundle. Maybe in the admin-bundle it does not make sense, but for example here, it could let you see the admin with site/page/snapshot and all, and test the features to see if they work or not.

I use a simply `php -S`, dockerization is not required since we use sqlite. BUT it could be added later too.

wdyt ?

tests/App/AppKernel.php
VincentLanglet: Why ? It's done on other projects https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/tests/App/AppKernel.php#L107

VincentLanglet:
> For some reason this deprecation starts triggering, do you know how to fix it @VincentLanglet ?

No idea... we need to look at the twig-extensions project

tests/App/AppKernel.php
jordisala1991: It's done on all sonata projects, I know.

jordisala1991: I probably have all /tmp folder full of sonata cacheses of my previous work on sonata

VincentLanglet: tmp content is automatically delete (not sure when thought).

I would say it's better to avoid creating a directory in the project which will never be empty.

jordisala1991: O my system this method is returning different when calling it from cli and when it is through a webserver, do you know a workaround for that? I dont really mind to have a directory created on the project, it kinda belongs to the project, having it outside makes things like: lets do a rm on the dir somehow impossible unless you start dumping around

VincentLanglet: With an example file: test.php
```
<?php
    echo sys_get_temp_dir() . PHP_EOL;
?>
```

You can run TMPDIR=/custom/location php test.php

I dont know much more about this function.

jordisala1991:
For now I won't invest more time on this thing. But it was interesting to investigate about it.

PR #1582, opened 2022-08-30
Add panther test case

jordisala1991:
## Subject

This adds some coverage for the javascript needed to run the compose page, here we are testing:

1. The compose page loads (already tested with normal tests)
2. That you can click on a container block (Top content)
3. That it should display the block list and be empty
4. That you can open the select and select the first block type (it should be text type block)
5. that you can click on the add new button
6. That you can click on save
7. And it should be saved and now you have one element on the list.


jordisala1991:
Just trying to make this setup work. ATM is not that easy since DAMA doctrine test bundle is not compatible with panther. We can't benefit from it, but it should (or at least I hope it) be possible to still benefit from it from normal functional test, and kind of disable it for Panther tests.

We have some similar setup on doctrine orm and doctrine mognodb, but it uses fixtures and not data loaded for each test (I don't like it a lot). Also it does not reset database between each test.

Maybe I need to create another DB just for this tests, but I am not sure then how to load data to that DB.

Any ideas on how to do this @VincentLanglet ? Maybe there is another way of doing this that we might want to explore.

Maybe @dmaicher can give it opinion too.

What should we do?

1. Create another database for e2e tests that gets loaded with fixtures on custom_bootstrap, then disable dama for that database
2. Use same database and load data on each test ? (how to do it?), somehow disabling dama on specific test files
3. Is there another option?

VincentLanglet:
> We have some similar setup on doctrine orm and doctrine mognodb, but it uses fixtures and not data loaded for each test (I don't like it a lot). Also it does not reset database between each test.

Not sure why you dont like it, but fixtures seemed fine to me for other projects

jordisala1991:
depends on the project, here those fixtures would mean that your database is not stable. Thats why its recommended to have your test do the whole thing and reset database between each test. For example:

If I have one test that inserts a site and generates core routes and pages and then I assert on how many pages or snapshots are generated, there are 3 or more options:

1. I know my tests are isolated so my database only contains what I want, and I dont depend on the execution order. I can test how many pges are created.
2. I use fixtures on the boostrap file and do not reset database between each test. Then I can only compare how many was before and after, because I would probably have more stuff in the database from
another tests, my tests are not isolated.
3. I use fixtures but I reset to initial state on each test. This might be a good compromise between 1 and 2, BUT the problem is, each time your initial data changes because you add more cases to your tests, a lot of things could break, and you end up with a lot of initial data, that you dont know for which test are needed.


Personally I like the 1st because I know for each test what I am adding to the database, and I know exactly what I can expect after each action. Also there are good bundles like DAMA and Foundry that helps with that (it is not something that I made, it is how normally you test with database data).

The problem with DAMA in this case is that when you start running your phpunit tests it alters how data is cached and stored (not sure how it does tbh) and it is not compatible with panther, so we need to do a solution that works with both kind of tests: functionals without real browser, and functionals with real browser.

tests/Functional/Admin/PantherTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+        self::prepareData();
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        self::rollbackDatabase();
+
+        StaticDriver::setKeepStaticConnections(true);
+    }
```
jordisala1991: This allows us to bypass DAMA Doctrine test bundle and load data specifically for this test class. WDYT @VincentLanglet ? It looks good enough for me.

VincentLanglet: Yeah seems nice

tests/Functional/BasePantherTestCase.php
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+namespace Sonata\PageBundle\Tests\Functional;
+
+use DAMA\DoctrineTestBundle\Doctrine\DBAL\StaticDriver;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\SchemaTool;
+use Symfony\Component\Panther\PantherTestCase;
+
+abstract class BasePantherTestCase extends PantherTestCase
```
jordisala1991, line 21: We can reuse this logic for all panther tests (even if I recommend to add as less as possible tests that run with real browser given how much it affects to performance)

Only do panther tests if you need to interact with the page and it includes javascript, otherwise webtestcase are good enough

jordisala1991: For me there are a few cases for this bundle, maybe only the compose page and maybe the collection type of the container block

tests/Functional/Admin/PantherTest.php
```diff
@@ -0,0 +1,68 @@
...
+        $crawler->filter('.select2-results__options li')->first()->click();
+        $crawler->filter('.page-composer__block-type-selector__confirm')->first()->click();
+
+        $client->waitFor('.page-composer__container__child');
+
+        $crawler->filter('.page-composer__container__child button')->first()->click();
+
+        $client->waitFor('.page-composer__container__child__enabled');
+
+        static::assertCount(1, $crawler->filter('.page-composer__container__children li'));
```
jordisala1991: Not super expert on the API of panther, for sure there could be better ways to looks for elements and click them, this is the one I found.

jordisala1991:
wdyt @VincentLanglet ? It could be a nice addition (the idea is not to replace other tests but complement them). Also is not a replacement of javascript unit test (unable to do them now because of how the javascript part is coded, it will be possible when we switch code to stimulus as seen on the date picker PR).

also notice that we should not abuse this kind of tests.

.gitignore
```diff
@@ -12,3 +12,4 @@ phpstan.neon
 /.phpunit.result.cache
 /docs/_build/
 npm-debug.log
+/tests/App/public/bundles
```
VincentLanglet, line 15: This require a dev-kit pr

phpunit.xml.dist
```diff
@@ -31,12 +31,14 @@ It's auto-generated by sonata-project/dev-kit package.
...
 
     <extensions>
+        <extension class="Symfony\Component\Panther\ServerExtension" />
         <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension" />
     </extensions>
 
     <php>
         <ini name="precision" value="8" />
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+        <server name="PANTHER_WEB_SERVER_DIR" value="./tests/App/public/" />
```
VincentLanglet, line 41: need a dev kit pr

jordisala1991: This is already on dev-kit

.gitignore
jordisala1991: Same, already on dev kit https://github.com/sonata-project/dev-kit/blob/788cc95853a037eb1c4bcf51e66439b0b73b4f5d/templates/project/.gitignore.twig#L18

tests/Functional/Admin/PantherTest.php
```diff
@@ -0,0 +1,68 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Tests\Functional\Admin;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Sonata\PageBundle\Tests\App\Entity\SonataPagePage;
+use Sonata\PageBundle\Tests\Functional\BasePantherTestCase;
+
+final class PantherTest extends BasePantherTestCase
```
jordisala1991: Class name should be changed to something more meaningful.

jordisala1991: Done.

.gitignore
VincentLanglet: I thought we had some `panther: true` to add. Forgot I made change about this.

tests/Functional/Admin/BrowserTest.php
```diff
@@ -0,0 +1,68 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Tests\Functional\Admin;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Sonata\PageBundle\Tests\App\Entity\SonataPagePage;
+use Sonata\PageBundle\Tests\Functional\BasePantherTestCase;
+
+final class Browser extends BasePantherTestCase
```
VincentLanglet:
```suggestion
final class BrowserTest extends BasePantherTestCase
```

VincentLanglet (changes requested):
Linter fail

VincentLanglet (approved):

PR #1583, opened 2022-08-31
Add frontend css file for pages

### LINKED ISSUE #1573: Make bootstrap works in 4.x branch
As this bug need to make a bit more work, I would prefer to create a issue for this and handle it here!

those are the steps mentioned by @jordisala1991 

https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1225269266
```
These are the steps I would follow:

Install with npm the same version of bootstrap we have on adminBundle
Create a new frontend.css and frontend.js that only expose bootstrap
Add it on webpack config as new entry points
Add that to the page to see if it everything renders correctly
If not, change the html to fit how it looks on 3.x
Update docs , upgrade note or whatever to reflect changes
Once that gets merged we might want to also take a look at the editor links we have on frontend, but I would treat that as separate topic.
```

others comments related with this issue: 
- https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1210303865

eerison:
## Fixing the site's page to work as 3.x

I am targeting this branch, because the bootstrap wasn't loading in site's page.

Closes #1573.

## Changelog

```markdown
### Added
- Added `assets/css/frontend.scss` to use in site's page
```

## To do

- [x] Fix pipelines
- [x] Update the documentation;


eerison:
Hey @jordisala1991 to make this accessible from my project, I guess I should run some build here, don't you?

eerison:
I remember some similar issue related with admin bundle

eerison:
Screenshot 2022-09-02 at 09 13 18

any idea why it's not loading ?

eerison:
Ok I found the issue

it need to be added here

```
sonata_page:
    assets:
        stylesheets:
            - bundles/sonatapage/frontend.css
```

eerison:
Screenshot 2022-09-02 at 09 29 36
Now it's look nice again :)

eerison:
Screenshot 2022-09-02 at 10 01 16

eerison:
Can you review @jordisala1991 and @VincentLanglet

docs/reference/installation.rst
```diff
@@ -121,6 +121,12 @@ SonataPageBundle Configuration
             not_found: [404] # render 404 page with "not_found" key (name generated: _page_internal_error_{key})
             fatal: [500] # so you can use the same page for different http errors or specify specific page for each error
 
+        # This configuration is optional,
+        # If you don't want to load the sonata page's layout for your website just remove it, and load your css.
+        assets:
+            stylesheets:
+                - bundles/sonatapage/frontend.css
```
jordisala1991: IMO this does not look correct. https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/DependencyInjection/Configuration.php#L140

This should be changed.

eerison: ohhhh ok then I should create a app.css and ~let this as~ use the default value

jordisala1991: No, frontend.css is fine, (there is already an app.css), but you need to change the configuration.

eerison: Hey @jordisala1991 But we have a problem here, this app.js is used for admin not for FE (site)

eerison: or should add this as default value? like

```php
                         ->defaultValue([
                                'bundles/sonatapage/app.css',
                                'bundles/sonatapage/frontend.css',
                            ])
```

eerison: @jordisala1991  I have a question here, the page loads app.css and frontend.css, I guess we should load only frontend.css, don't you?

eerison: the question is how!

eerison: Ok we can handle this in other topic, let's make this working as 3.x

jordisala1991: you should only let the frontend.css line, the other remove it.

eerison: Here is the problem

https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Resources/views/base_layout.html.twig#L25

it's loading all css, maybe a other solution could be add a new option like `frontend_ stylesheet`

what do you think?

eerison:
it worked with default configuration

docs/reference/installation.rst
eerison: I would say it can be handled in other PR, because the current approach for this is loading app.css and it was loading bootstrap and admin css/js files in all pages, then we could provide some improvement later

jordisala1991: You should have to change configuration and only leave one file.

eerison:
Screenshot 2022-09-02 at 15 53 37

Ok now it's loading only frontend.css :)

docs/reference/installation.rst
eerison: Could you check again @jordisala1991

src/DependencyInjection/Configuration.php
```diff
@@ -145,6 +145,16 @@ public function getConfigTreeBuilder(): TreeBuilder
                             ->defaultValue([])
                             ->prototype('scalar')->end()
                         ->end()
+                        ->arrayNode('frontend_stylesheets')
```
jordisala1991: There is no need to add a new option, please use the above. It is only used for the frontend.

jordisala1991 (changes requested):

src/DependencyInjection/Configuration.php
jordisala1991: Or even better, IMO this template will be normally replaced for the user because he will be using its own design without bootstrap. 

I think we should deprecate / remove this config and place the css without being a config option.

eerison: Yeah i guess it'll be replaced too, but if we don't provide an Frontend css, when the user I install the page bundle the standard layout will be ugly

jordisala1991: No, it wont, by default we will provide frontend.css. but it doesn't need to be configurable, it can be just placed on the template.

eerison: Hmmm ok, I'll try that tomorrow :)

src/Resources/views/base_layout.html.twig
```diff
@@ -21,24 +21,13 @@ file that was distributed with this source code.
...
                 {% block page_stylesheets %} {# Deprecated block #}
-                    {% for stylesheet in sonata_page.assets.stylesheets %}
-                        <link rel="stylesheet" href="{{ asset(stylesheet) }}" media="all">
-                    {% endfor %}
+                        <link rel="stylesheet" href="{{ asset('bundles/sonatapage/frontend.css') }}" media="all">
                 {% endblock %}
             {% endblock %}
 
             {% block sonata_page_javascripts %}
                 {% block page_javascripts %} {# Deprecated block #}
```
eerison, line 29: Hey @jordisala1991 I will create a new issue (or just add a comment in 4.x issue) to remove those deprecating blocks, ok?

eerison:
could you check again @jordisala1991

jordisala1991 (approved):
We should also deprecate the css and js config on 3.x and remove it on 4.x after (other Prs)

VincentLanglet (approved):

VincentLanglet:
> We should also deprecate the css and js config on 3.x and remove it on 4.x after (other Prs)

Can you add it to the todo @eerison ?

PR #1585, opened 2022-09-02
Add tests for page types

jordisala1991:
## Subject

Added more test to ensure hybrid pages work. (This can be later used to document how hybrid pages work)

Still missing (next PRs): 
* Disabled pages (For editors they appear to be published)
* Dynamic pages
* Snapshot types

### DIFF
--- a/src/Model/TransformerInterface.php
+++ b/src/Model/TransformerInterface.php
@@ -41,7 +41,6 @@ use Doctrine\Common\Collections\Collection;
  *     blocks: array<mixed>,
  *   }>,
  * }
- *
  * @phpstan-type PageContent array{
  *   id: int|string|null,
  *   parent_id?: int|string|null,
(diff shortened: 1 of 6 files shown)


src/Model/TransformerInterface.php
```diff
@@ -41,7 +41,6 @@
  *     blocks: array<mixed>,
  *   }>,
  * }
- *
```
eerison, line 44: I'm fixing this too, in my PR 😅

VincentLanglet: With the new php-cs-fixer release I was wondering about updating our cs-standard @jordisala1991

jordisala1991: What do you want to change? It's fine by me, but until it gets updated, we should apply the fixes right?

VincentLanglet: Please take a look at a proposal https://github.com/sonata-project/SonataAdminBundle/pull/7920

we can order the param/return/throw/phpstan-tags
and we can configure the separation to be similar to our current code style

VincentLanglet (approved):

PR #1588, opened 2022-09-03
DevKit updates for 3.x branch

SonataCI:


VincentLanglet:
We have a probleme here @jordisala1991 ^^'

jordisala1991:
A big, ugly and problematic merge with tons of conflicts? Just break it and I will do the merge tomorrow.

VincentLanglet:
> A big, ugly and problematic merge with tons of conflicts? Just break it and I will do the merge tomorrow.

Indeed the merge will be ugly, but first, we cannot use the config from dev-kit since it cannot install the latest version of php-cs-fixer.

We might remove the update of the php-cs-fixer config for SonataPage 3.x branch.

jordisala1991:
I see 2 options:

1. Ignore php cs fixer failing on this branch.
2. Add this new config conditionally if not on page bundle 3.x otherwise revert to older config.

VincentLanglet:
> I see 2 options:
> 
> 1. Ignore php cs fixer failing on this branch.
> 2. Add this new config conditionally if not on page bundle 3.x otherwise revert to older config.

I'm ok ignoring this, this avoid extra work

PR #1591, opened 2022-09-04
Add tests for snapshots types

jordisala1991:
## Subject

Missing tests added.

### DIFF
--- a/tests/App/config/services.php
+++ b/tests/App/config/services.php
@@ -12,6 +12,7 @@ declare(strict_types=1);
  */
 
 use Psr\Container\ContainerInterface;
+use Psr\Log\NullLogger;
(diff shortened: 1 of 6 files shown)


tests/App/config/services.php
```diff
@@ -23,5 +24,7 @@
         ->set(DynamicController::class)
             ->public()
             ->tag('container.service_subscriber')
-            ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)]);
+            ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)])
+
+        ->set('logger', NullLogger::class);
```
jordisala1991, line 29: with this we reduce the output between tests. I guess it could be applied to all repositories that use kernelTestCase.

VincentLanglet (approved):

PR #1597, opened 2022-09-07
Accept enabled as string '1' and '0'

jordisala1991:
## Subject

I am targeting this branch, because this is where the bug happens.

### DIFF
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -179,7 +179,7 @@ final class Transformer implements TransformerInterface
             $block->setName($content['name']);
         }
 
-        $block->setEnabled($content['enabled']);
+        $block->setEnabled(\in_array($content['enabled'], ['1', true], true));
 
         if (isset($content['position']) && is_numeric($content['position'])) {
             $block->setPosition((int) $content['position']);
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -179,6 +179,7 @@ final class Transformer implements TransformerInterface
             $block->setName($content['name']);
         }
 
+        // NEXT_MAJOR: Simplify this code by removing the in_array function and assign directly.
         $block->setEnabled(\in_array($content['enabled'], ['1', true], true));
 
         if (isset($content['position']) && is_numeric($content['position'])) {
--- a/src/Model/TransformerInterface.php
+++ b/src/Model/TransformerInterface.php
@@ -18,6 +18,10 @@ use Doctrine\Common\Collections\Collection;
 /**
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  *
+ * NEXT_MAJOR: Restrict and simplify BlockContent type:
+ *   - position: int|null
+ *   - enabled: boolean
+ *
  * @phpstan-type BlockContent array{
  *   id: int|string|null,
  *   name?: string|null,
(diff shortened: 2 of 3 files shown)


src/Entity/Transformer.php
```diff
@@ -179,7 +179,7 @@ public function loadBlock(array $content, PageInterface $page): PageBlockInterfa
             $block->setName($content['name']);
         }
 
-        $block->setEnabled($content['enabled']);
+        $block->setEnabled(\in_array($content['enabled'], ['1', true], true));
```
eerison, line 183: and whats happen when it's '0'?

eerison: ahh ok, won't have '0' into the array and it will be false!

it's not really explicit that is working for '1' and '0', But I got it!


should be good create a test with `0`

jordisala1991: Unless someone has a snapshot with 1|0 as integer, I prefer to not add things for "what if".

eerison: I mean create a test with `'0'` as string

src/Model/TransformerInterface.php
```diff
@@ -21,7 +21,7 @@
  * @phpstan-type BlockContent array{
  *   id: int|string|null,
  *   name?: string|null,
- *   enabled: boolean,
+ *   enabled: boolean|'1'|'0',
```
VincentLanglet, line 28: What is the reason for having `'1'` or `'0'` ?
Can't we always having a bool ?

jordisala1991: Seems not, or at least not for previous versions of this bundle: https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1236801143

cc @eerison

eerison: Yeah, for the previous version many values are string :/

I have the felling that we should fix this snapshot data in 3.x and the users should regenerate all snapshots, this way we can assume that all data is correct, otherwise we will add too much workaround to make snapshot from 3.x compatible with 4.x code

VincentLanglet: Yeah, if I understand correctly the issue is that the data were not perfect in 3.x ; so we have to add BC-layer now in 4.x.
But we will be able to remove those BC-layer in 5.x.

Should we already add NEXT_MAJOR comment to remove the bc layer and update the phpdoc then ?

jordisala1991: I will see if I can add some NEXT_MAJOR comment.

jordisala1991: I tried with a NEXT_MAJOR comment @VincentLanglet

src/Entity/Transformer.php
jordisala1991: Added '0' test case.

VincentLanglet, line 183: Next major can be added here too no ?

jordisala1991: Does it matter tho? phpstan and psalm will comply when you change the type, also in test could be added too, but its the same. Do you prefer adding NEXT_MAJOR on all places?

VincentLanglet: They won't complain here: https://phpstan.org/r/a02bf6cd-33d1-407a-a39c-b0a06b9e8023

jordisala1991, line 183:
```suggestion
        // NEXT_MAJOR: Simplify this code by removing the in_array function and assign directly.
        $block->setEnabled(\in_array($content['enabled'], ['1', true], true));
```

jordisala1991: Alright, added!.

VincentLanglet (approved):

PR #1598, opened 2022-09-07
Fix snapshots created through Snapshot Admin

jordisala1991:
## Subject

I am targeting this branch, because this bug only happens on 4.x.

### DIFF
--- a/src/Admin/SnapshotAdmin.php
+++ b/src/Admin/SnapshotAdmin.php
@@ -23,6 +23,7 @@ use Sonata\Form\Type\DateTimePickerType;
 use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Model\PageManagerInterface;
 use Sonata\PageBundle\Model\SnapshotInterface;
+use Sonata\PageBundle\Model\SnapshotManagerInterface;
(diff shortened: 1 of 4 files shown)


src/Admin/SnapshotAdmin.php
```diff
@@ -73,9 +78,23 @@ protected function alterNewInstance(object $object): void
...
+    protected function postPersist(object $object): void
+    {
+        $page = $object->getPage();
+
+        if (null !== $page) {
+            $page->setEdited(false);
+            $this->pageManager->save($page);
+        }
+
+        $this->snapshotManager->enableSnapshots([$object]);
```
jordisala1991, line 97: Basically here we are splitting the https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Service/CreateSnapshotService.php#L63-L74

in two pieces. First we create the snapshot in pre persist, and once is persisted, we set page edited to false and enable snapshots (to enable snapshot we need the id to be set, that's why we separate in 2 steps).

VincentLanglet (dismissed):
Can we add a test ?

jordisala1991:
> Can we add a test ?

Of coure, added.

VincentLanglet (approved):

PR #1605, opened 2022-09-09
Ensure each admin can access edit page

jordisala1991:
## Subject

I am targeting this branch, because this only happens on 4.x.


src/Admin/PageAdmin.php
```diff
@@ -175,7 +175,13 @@ protected function configureListFields(ListMapper $list): void
             ])
             ->add('decorate', null, ['editable' => true])
             ->add('enabled', null, ['editable' => true])
-            ->add('edited', null, ['editable' => true]);
+            ->add('edited', null, ['editable' => true])
+            ->add(ListMapper::NAME_ACTIONS, ListMapper::TYPE_ACTIONS, [
+                'translation_domain' => false,
```
jordisala1991: I am not sure how to let sonata know I do not want SonataPageBundle as translation domain for this specific field, otherwise this happens:

image

VincentLanglet: Translation false wont translate. Shouldnt it be SonataAdminBundle ?

jordisala1991: just tried, it doesn't matter.

jordisala1991: It does get picked up as SonataAdminBundle on the translation domain, but there isn't any translation for it on the xliff files of SonataAdminBundle ....

eerison: add an edit button a good approach for this links issues 😃 !

VincentLanglet (dismissed):

src/Admin/PageAdmin.php
```diff
@@ -175,7 +175,13 @@ protected function configureListFields(ListMapper $list): void
             ])
             ->add('decorate', null, ['editable' => true])
             ->add('enabled', null, ['editable' => true])
-            ->add('edited', null, ['editable' => true]);
+            ->add('edited', null, ['editable' => true])
+            ->add(ListMapper::NAME_ACTIONS, ListMapper::TYPE_ACTIONS, [
+                'translation_domain' => 'SonataAdminBundle',
```
VincentLanglet, line 180: If I understand correctly, even with this, it displays `list.label__actions` ?

jordisala1991: https://github.com/sonata-project/SonataAdminBundle/pull/7926

Yes... there is no translation for this.

VincentLanglet: I think it's understandable, because
- The label depends on the LabelTranslatorStrategy
- The translation domain is messages by default, but depends on the admin configuration

So I would say that we could remove `'translation_domain' => 'SonataAdminBundle'` and we should add translations for `list.label__actions` in PageBundle.

But I wonder how we didn't need this in any other bundle so far.
The other solution would be to add the translations in SonataAdminBundle, in order to use it in any other bundle. (But technically we should support all the TranslatorStrategy in https://github.com/sonata-project/SonataAdminBundle/tree/4.x/src/Translator ...)

VincentLanglet (changes requested):

src/Admin/PageAdmin.php
jordisala1991: SonataUserBundle needs similar solution, because is relying on addIdentifier.

VincentLanglet (approved):

jordisala1991:
Since the labels are already fixed on AdminBundle (just waiting for a release) I am merging this.

PR #1606, opened 2022-09-12
Fix request matcher to ensure we don't try to match / twice

### LINKED ISSUE #1474: host_with_path with dynamic page.
I want create sites for diferent locales (example.com/page, example.com/en/page) so I am using host_with_path multisite option. But dynamic page (example.com/project/{slug}) stopped working and when visiting example.com/project/example it redirects back to example.com. When I added dynamic page to ignore_uri_patterns, it works, but blocks in global page doesn´t work.

jordisala1991:
## Subject

I am targeting this branch, because this fixes a bug.

Closes #1474 

## Changelog

```markdown
### Fixed
- Fix `host_with_path` with site on a relative path of '/'.
- Fix `View Page` button on Page edit, when site has relative path of '/'.
```

### DIFF
--- a/src/Admin/PageAdmin.php
+++ b/src/Admin/PageAdmin.php
@@ -368,10 +368,10 @@ final class PageAdmin extends AbstractAdmin
                 $site = $page->getSite();
 
                 if (null !== $site) {
-                    $siteRelativePath = $site->getRelativePath();
+                    $relativePath = $site->getRelativePath();
 
-                    if (null !== $siteRelativePath) {
-                        $path = $siteRelativePath.$path;
+                    if (!\in_array($relativePath, [null, '/'], true)) {
+                        $path = $relativePath.$path;
                     }
                 }
 
--- a/src/Site/BaseSiteSelector.php
+++ b/src/Site/BaseSiteSelector.php
@@ -116,7 +116,12 @@ abstract class BaseSiteSelector implements SiteSelectorInterface
         // we read the value from the attribute to handle fragment support
         $requestPathInfo = $request->get('pathInfo', $request->getPathInfo());
 
-        if (0 === preg_match(sprintf('@^(%s)(/.*|$)@', $site->getRelativePath() ?? ''), $requestPathInfo, $results)) {
+        $relativePath = $site->getRelativePath();
+        $regex = !\in_array($relativePath, [null, '/'], true) ?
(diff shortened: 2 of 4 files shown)


jordisala1991:
Since its related, I fixed: https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1241038799

I am sure this bug was present also on 3.x, but whatever. those should be the last bugs on 4.x (BESIDES the big one: https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1235626107)

VincentLanglet (commented):
Does a test should/need to be added ?

jordisala1991:
I think it should be possible in this case.

src/Admin/PageAdmin.php
```diff
@@ -368,10 +368,10 @@ protected function configureTabMenu(ItemInterface $menu, string $action, ?AdminI
                 $site = $page->getSite();
 
                 if (null !== $site) {
-                    $siteRelativePath = $site->getRelativePath();
+                    $relativePath = $site->getRelativePath();
 
-                    if (null !== $siteRelativePath) {
-                        $path = $siteRelativePath.$path;
+                    if (!\in_array($relativePath, [null, '/'], true)) {
+                        $path = $relativePath.$path;
```
jordisala1991, line 374: This one is now tested by the first commit, there is still the other src change missing a test.

src/Site/BaseSiteSelector.php
```diff
@@ -116,7 +116,12 @@ protected function matchRequest(SiteInterface $site, Request $request)
         // we read the value from the attribute to handle fragment support
         $requestPathInfo = $request->get('pathInfo', $request->getPathInfo());
 
-        if (0 === preg_match(sprintf('@^(%s)(/.*|$)@', $site->getRelativePath() ?? ''), $requestPathInfo, $results)) {
+        $relativePath = $site->getRelativePath();
+        $regex = !\in_array($relativePath, [null, '/'], true) ?
+            sprintf('@^(%s)(/.*|$)@', $relativePath) :
+            '@^()(/.*|$)@';
+
+        if (0 === preg_match($regex, $requestPathInfo, $results)) {
```
jordisala1991, line 124: this is now tested with a previous commit.

VincentLanglet (approved):

PR #1607, opened 2022-09-12
Fix custom url

### LINKED ISSUE #502: Custom URL (not fully implemented ?)

jordisala1991:
## Subject

I am targeting this branch, because this is a bug fix.

Closes #502 

## Changelog

```markdown
### Fixed
- Fix `customUrl` implementation.
```

### DIFF
--- a/src/Entity/PageManager.php
+++ b/src/Entity/PageManager.php
@@ -100,8 +100,12 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
             $parent = $page->getParent();
 
             if (null !== $parent) {
-                if (null === $page->getSlug()) {
-                    $page->setSlug($this->slugify->slugify($page->getName() ?? ''));
+                $slug = $page->getSlug();
+
+                if (null === $slug) {
+                    $slug = $this->slugify->slugify($page->getName() ?? '');
+
+                    $page->setSlug($slug);
                 }
 
                 $parentUrl = $parent->getUrl();
@@ -114,11 +118,13 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
                     $base = $parentUrl;
                 }
 
-                $page->setUrl($base.$page->getSlug());
+                $url = $page->getCustomUrl() ?? $slug;
+                $page->setUrl('/'.ltrim($base.$url, '/'));
             } else {
-                // a parent page does not have any slug - can have a custom url ...
                 $page->setSlug(null);
-                $page->setUrl('/'.$page->getSlug());
+
+                $url = $page->getCustomUrl() ?? '';
+                $page->setUrl('/'.ltrim($url, '/'));
             }
         }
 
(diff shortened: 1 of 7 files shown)


VincentLanglet:
Does a test should be added ?

You generally add a test which is failing yhe ci then yiu fix the bug.

Here you changed the code (failing the ci) then you fixed the test. Does it mean test was wrong ?

jordisala1991:
code changed behavior, that's why test was wrong.

I will try to see if a test can be added, but not sure it is easy.

jordisala1991:
Test added, both for pages and snapshots, @VincentLanglet .

The thing is: the URL is not editable by the user. The user can only add pages being child of other pages, with a hierarchy like:

* Home (url: '/')
* Inner: (url: '/test') (because on slug we put "test")
* Another... (url: '/test/test2') (because child of inner and we put "test2" on slug)

The other field you can change is the customURL, that way the url will be whatever you place on that field, like: '/test1/test2/test3', without the need of the full hierarchy. Also this field is only usable on CMS pages, NOT hybrid or dynamic (those pages fixes the url through symfony routing).

IMO it makes sense to keep using everywhere the url field. there was another PR proposing to do some match between url and custom url, but that complicates things IMO.

The only thing that was taken into account is the relativePath of the SITE

VincentLanglet (approved):

Issue #1609, opened 2022-09-12
Keep the blocks into the page rendering even if some blocks does not exist


eerison:
the current code in 3.x and 4.x is crashing the whole page when the block doesn't exist, I faced this issue few months ago, and It should be fixed here.

My idea is, check if the block class still exist, if not log into the logs and jump for the next block to be rendered.

### ACs

- [x] If the block doesn't exist, the page that contain this block must be rendering without this block
- [x] when try to render a block that doesn't exist, you must log the exception and jump for the next block
- [ ] I should be able to remove any block even if it does't exist.

Note: It's just an example, it can happen with any block that doesn't exist for example, if you rename some block that already exist!

<details><summary>Exception</summary>
<p>

```
Twig\Error\RuntimeError:
An exception has been thrown during the rendering of a template ("The block service `sonata.seo.block.email.share_button` does not exist").

  at vendor/sonata-project/page-bundle/src/Resources/views/Block/block_container.html.twig:15
[...]

PR #1614, opened 2022-09-19
Not render block that does not exist

eerison:
## Not render block that does not exist

I am targeting this branch, because it requires a fix in sonata block bundle 4.x and it's supported only for page bundle 4.

Issue #1609.

## Changelog

```markdown
### Fixed
- Skipping blocks that doesn't exist from the page.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "doctrine/doctrine-bundle": "^2.5",
         "doctrine/persistence": "^2.1 || ^3.0",
         "sonata-project/admin-bundle": "^4.15",
-        "sonata-project/block-bundle": "^4.16.2",
+        "sonata-project/block-bundle": "^4.17",
         "sonata-project/doctrine-extensions": "^1.18 || ^2.1",
         "sonata-project/doctrine-orm-admin-bundle": "^4.0",
         "sonata-project/form-extensions": "^1.4",
--- a/src/Twig/Extension/PageExtension.php
+++ b/src/Twig/Extension/PageExtension.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Twig\Extension;
 
+use Sonata\BlockBundle\Exception\BlockNotFoundException;
 use Sonata\BlockBundle\Templating\Helper\BlockHelper;
 use Sonata\PageBundle\CmsManager\CmsManagerSelectorInterface;
 use Sonata\PageBundle\Exception\PageNotFoundException;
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
         "doctrine/doctrine-bundle": "^2.5",
         "doctrine/persistence": "^2.1 || ^3.0",
         "sonata-project/admin-bundle": "^4.15",
-        "sonata-project/block-bundle": "^4.16.2",
```
eerison, line 29: It'll be fixed after this PR https://github.com/sonata-project/SonataBlockBundle/pull/1115 pass

eerison:
I'll expect what gonna be decided in this PR: https://github.com/sonata-project/SonataBlockBundle/pull/1115

eerison:
Hey @VincentLanglet @jordisala1991 
do you think that we could handle this issue with this approach, or do we must use exception strategy from block bundle?

src/Twig/Extension/PageExtension.php
```diff
@@ -183,15 +189,21 @@ public function renderContainer(string $name, $page = null, array $options = [])
...
+        try {
+            if (
+                false === $block->getEnabled()
+                && false === $this->cmsManagerSelector->isEditor()
+                && $this->hideDisabledBlocks
+            ) {
+                return '';
+            }
+
+            return $this->blockHelper->render($block, $options);
```
VincentLanglet, line 201: You can try/catch only the render call.

src/Twig/Extension/PageExtension.php
```diff
@@ -183,15 +189,21 @@ public function renderContainer(string $name, $page = null, array $options = [])
...
+        } catch (BlockNotFoundException $exception) {
```
VincentLanglet, line 202: Why BlockNotFoundException ?

I thought we introduced a BlockServiceNotFoundException ; you need to catch something else ?

VincentLanglet (changes requested):

src/Twig/Extension/PageExtension.php
eerison: Yeah I can do that!

src/Twig/Extension/PageExtension.php
eerison: Yeah it's the old exception I need to change!

composer.json
eerison: But I need this PR :/

VincentLanglet: See https://github.com/sonata-project/SonataBlockBundle/pull/1115#issuecomment-1254704463

PR #1618, opened 2022-09-26
Fix shared block service

jordisala1991:
## Subject

I am targeting this branch, because shared blocks are broken on 4.x.

Fixes: https://github.com/sonata-project/SonataPageBundle/issues/1495#issuecomment-1252021888

### DIFF
--- a/src/Block/SharedBlockBlockService.php
+++ b/src/Block/SharedBlockBlockService.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Block;
 
+use Sonata\AdminBundle\Admin\AdminInterface;
 use Sonata\AdminBundle\Form\FormMapper as AdminFormMapper;
 use Sonata\AdminBundle\Form\Type\ModelListType;
 use Sonata\BlockBundle\Block\BlockContextInterface;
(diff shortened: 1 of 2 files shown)


jordisala1991:
This issue is hard to test because it involves the need of a panther test.

jordisala1991:
@eerison on my tests, with these changes, the shared block works fine, can you try this?

src/Block/SharedBlockBlockService.php
```diff
@@ -130,20 +139,17 @@ public function load(BlockInterface $block): void
      */
     private function getBlockBuilder(AdminFormMapper $form): FormBuilderInterface
     {
-        $admin = $form->getAdmin();
```
VincentLanglet, line 133: it doesn't work with the admin coming from the form ?

VincentLanglet (approved):

src/Block/SharedBlockBlockService.php
jordisala1991: When you use the admin from the form, you are using BlockAdmin, but this field is to related with SharedBlockAdmin. What is more interesting is that this is happening inside PageAdmin, so for some reason, not only the admin linked is not the correct one which shows Blocks, not shared blocks. BUT also Sonata is smart enough to know Block is child of Page and automatically builds child urls, showing only blocks related to that page. And the final funnier thing is that for child urls , the select button does not work, I can't understand why tho.

So to answer you:
1. No, the admin is not BlockAdmin but SharedBlockAdmin
2. Previously only blocks belonging to the page were shown (instead of the expected behavior which is showing ALL shared blocks)
3. Sonata has a bug somewhere (I didn't fully understand why) that make the previous implementation fail because it tries to open a modal with a child, but without providing the correct linkParameters (it is hard to explain because I do not fully know what is happening there)
4. But the bug pointed on 3. is not really related to this issue, which is: We were relating to block instead of shared block.

Maybe we want to investigate more about this 3. in the future.

jordisala1991: Or explaining the other way around: If the point 3 didn't exists, this block should work, but it will let you only blocks that are already present on the page, not shared block intended to be reused across pages.

eerison:
Hey @jordisala1991 I tested your branch and I got the same behaviour, 


I create the block this way 👇🏼 

https://user-images.githubusercontent.com/6358755/192464859-b1f29d60-0553-45b5-8a07-746a14118383.mov

jordisala1991:
I think you are testing another thing, this fixes the shared block service (adding a shared block inside a page), not the batch delete.

eerison:
ahhh ok! I got it :) 

let me check again

eerison:
Hey @jordisala1991 

now I can add a share block in my page ❤️

eerison (approved):

PR #1623, opened 2022-09-30
Fix errors in composer container page

jordisala1991:
## Subject

I am targeting this branch, because this is a problem for 4.0, since some blocks were removed.

Closes #1611 

## Changelog

```markdown
### Added
- Add ability to display and remove blocks that belongs to types that no longer exist on compose page.
```

PR #1624, opened 2022-10-01
Verify that blocks implements EditableBlockService to add/edit them.

### LINKED ISSUE #1621: If Block does not implement EditableBlockService don't show it in composer
## Feature Request
If Block does not implement EditableBlockService don't show it in page composer as it only triggers errors.

jordisala1991:
## Subject

I am targeting this branch, because this becomes a problem on 4.x since block bundle splitted interfaces.

Closes #1621 

## Changelog

```markdown
### Fixed
- Fix to not show blocks that are not meant to be editable.
```

### DIFF
--- a/src/Controller/PageAdminController.php
+++ b/src/Controller/PageAdminController.php
@@ -16,6 +16,7 @@ namespace Sonata\PageBundle\Controller;
 use Sonata\AdminBundle\Controller\CRUDController;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
+use Sonata\BlockBundle\Block\Service\EditableBlockService;
 use Sonata\PageBundle\Admin\BlockAdmin;
 use Sonata\PageBundle\Admin\SnapshotAdmin;
 use Sonata\PageBundle\Model\BlockInteractorInterface;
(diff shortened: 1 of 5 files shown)


jordisala1991:
Needs #1623 first.

src/Controller/PageAdminController.php
```diff
@@ -273,6 +274,14 @@ public function composeContainerShowAction(Request $request): Response
         \assert($blockManager instanceof BlockServiceManagerInterface);
         $blockServices = $blockManager->getServicesByContext('sonata_page_bundle', false);
 
+        foreach ($blockServices as $code => $blockService) {
+            if ($blockService instanceof EditableBlockService) {
+                continue;
+            }
+
+            unset($blockServices[$code]);
+        }
```
VincentLanglet, line 283: We're doing this logic at least twice.

Would it make sens to add/modify a method in BlockServiceManager then ?
If most of the time/always we need to work on EditableBlockService only in PageBundle, it might make sens.

WDYT ?

jordisala1991: BlockServiceManager is from block-bundle, how could we handle this change?

VincentLanglet: I was wondering if we should introduce an EditableBlockServiceManager in BlockBundle then.
Or introduce getEditableService and getEditableServiceByContext...

But this solution is currently the only one we have

VincentLanglet (approved):

jordisala1991:
Should we merge this or wait for another solution? @VincentLanglet ?

VincentLanglet:
> Should we merge this or wait for another solution? @VincentLanglet ?

I'm ok merging this:
- This is working
- This can be easily removed as soon as we have a better solution

The only question would be, would it be useful to provide something in the BlockBundle to avoid this extra code ; if yes we might create a feature request.

haivala:
Check this comment about my thoughts about the implementation: https://github.com/sonata-project/SonataPageBundle/issues/1621#issuecomment-1275626062

PR #1633, opened 2022-10-24
Change default entity class values

eerison:
## Subject

## Changelog

```markdown
### Added

### Changed

### Deprecated

### Removed

### Fixed

### Security
```


eerison:
Hey @jordisala1991 do you have any idea why do we need this : https://github.com/sonata-project/SonataPageBundle/blame/4.x/src/DependencyInjection/SonataPageExtension.php#L266

in theory sonata page bundle works without it

VincentLanglet:
Should we just remove the default values since we don't provide entities ?

jordisala1991:
I think we should do the same like the other bundles (we didnt remove the default values)

VincentLanglet:
> I think we should do the same like the other bundles (we didnt remove the default values)

Like this then https://github.com/sonata-project/SonataPageBundle/pull/1635 ?
Seems like tests are failing

PR #1641, opened 2022-11-02
Overwrite PageAdmin

### LINKED ISSUE #1639: [Doc][Bug] It's throwing an exception when we are decorating PageAdmin
in version 4.x was removed those parameters

https://github.com/sonata-project/SonataPageBundle/blob/3.x/src/Resources/config/admin.xml#L7-L25

I didn't test it yet, But I suspect the only way to overwrite the classes like `PageAdmin` is decorating right?

I don't know! I'll do some tests and add some doc/UPGRADE notes for this!

eerison:
## Subject

Closes #1639.


src/Controller/PageAdminController.php
```diff
@@ -119,7 +119,7 @@ public function treeAction(Request $request): Response
         $theme = $this->admin->getFilterTheme();
         $this->setFormTheme($formView, $theme);
 
-        return $this->renderWithExtraParams($this->admin->getTemplateRegistry()->getTemplate('tree'), [
+        return $this->renderWithExtraParams('@SonataPage/PageAdmin/tree.html.twig', [
```
VincentLanglet: Kinda weird. Doing this won't allow people to override the template by using the template registry.

eerison: hmmmm But can't you overwrite like this

`templates/bundles/SonataPage/PageAdmin/tree.html.twig` ?

VincentLanglet: What if you want to override the template for only one admin ?
See https://github.com/sonata-project/SonataPageBundle/issues/1639#issuecomment-1300000091, there is something to look at in the configuration, about decorated services.

jordisala1991: I don't see the benefit of this change. Even with the old line of code, if you overwrite the template adding a new one with the same name on the `bundles/SonataPage/PageAdmin` it should work.

eerison: Hey @jordisala1991, it was just a try to fix the page admin when I'm decorating the PageAdmin service!

it's the exception: https://github.com/sonata-project/SonataPageBundle/issues/1639#issuecomment-1299912766

eerison:
@VincentLanglet  just to make sure

for now the only way to customize the admin like `PageAdmin` is decorate them, is that correct?

VincentLanglet:
> @VincentLanglet just to make sure
> 
> for now the only way to customize the admin like `PageAdmin` is decorate them, is that correct?

- You can use adminExtensions which should provide all you need.
- You may be able to decorate indeed.

I dunno if we should provide a BasePageAdmin like the BaseMediaAdmin in
https://github.com/sonata-project/SonataMediaBundle/tree/4.x/src/Admin or the Model/UserAdmin in https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/Admin/Model/UserAdmin.php

eerison:
> > @VincentLanglet just to make sure
> > for now the only way to customize the admin like `PageAdmin` is decorate them, is that correct?
> 
> * You can use adminExtensions which should provide all you need.
> * You may be able to decorate indeed.
> 
> I dunno if we should provide a BasePageAdmin like the BaseMediaAdmin in https://github.com/sonata-project/SonataMediaBundle/tree/4.x/src/Admin or the Model/UserAdmin in https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/Admin/Model/UserAdmin.php

> You can use adminExtensions which should provide all you need.

Do you have some example?

> I dunno if we should provide a BasePageAdmin like the BaseMediaAdmin

IMO it's not really interesting because it's almost the same to have PageAdmin without `final` and if I do this, I need to do the same thing for all admins in this bundle because I need to customiza almost all of them :D

VincentLanglet:
>> You can use adminExtensions which should provide all you need.
>
> Do you have some example?

https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/reference/extensions/
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Admin/Extension/LockExtension.php
https://github.com/sonata-project/SonataMediaBundle/pull/2348

docs/reference/customize_admin.rst
```diff
@@ -0,0 +1,65 @@
+Customize admin
+================
+
+Sonata page bundle there are few admins that can be customize using `Admin Extension`_
+As example if you want to add a new field in ``configureListFields``, You can do like this
+
+.. code-block:: php
```
eerison: Hey @VincentLanglet  any ida why lint is complaining about this line?

it's the same of : https://raw.githubusercontent.com/sonata-project/SonataAdminBundle/4.x/docs/reference/form_types.rst

VincentLanglet: Dont know a lot about this linter, but the message seems explicit.

docs/reference/customize_admin.rst
```diff
@@ -0,0 +1,65 @@
...
+
+    namespace App\Admin\Extension;
+
+    use Sonata\AdminBundle\Admin\AbstractAdminExtension;
+    use Sonata\AdminBundle\Datagrid\ListMapper;
+    use Symfony\Component\DependencyInjection\Attribute\AutoconfigureTag;
+
+    #[AutoconfigureTag(name: 'sonata.admin.extension', attributes: ['target' => 'sonata.page.admin.page'])]
+    final class PageAdminExtension extends AbstractAdminExtension
```
VincentLanglet: Isn't this doc enough https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/reference/extensions/ ?

docs/reference/customize_admin.rst
eerison: I don't know why, But it works with `::`!

docs/reference/customize_admin.rst
```diff
@@ -0,0 +1,65 @@
...
+        {
+            $list->add('id');
+        }
+    }
+
+And you will see the field in the end of the table
+
+.. image:: ../images/customize_admin__add_id.png
+
+And for order the field to the begin of your table can use the method ``reorder``
```
VincentLanglet: There is no occurence of `reorder` in the SonataAdmin doc
https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/search/?q=reorder&check_keywords=yes&area=default

I think it would be better to document this in this doc.

Not sure it's better in the list section https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/reference/action_list/#
 in the Extension section  https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/reference/extensions/
or in both

eerison:
ScreenShot Tool -20221107135702

docs/reference/customize_admin.rst
eerison: I guess it will be used more for ExtensionAdmin!

VincentLanglet: I think too (for the record it works for also others methods, like configureshowFields for instance)

VincentLanglet: I think it would be better to
- Improve SonataAdmin doc first
- Eventually add a doc like yours in SonataPage with link to SonataAdmin then :)

eerison: Ok I'll Open a PR to Admin bundle :)

VincentLanglet (commented):
This seems a little "short". I dunno if we need this doc like this...

Maybe we should present the doc in another way.
Instead of a `customize admin` titles, it should be `admin services`, with some text:
"There are some already implemented admin (...):"

Admin name - service name - purpose of the admin

And then a paragraph: "If you need to customize the admin, we recommend you to use admin extension"

src/Resources/config/admin.php
```diff
@@ -126,14 +126,22 @@
...
+        ->set(CreateSnapshotAdminExtension::class)
             ->tag('sonata.admin.extension', ['target' => 'sonata.page.admin.page'])
             ->tag('sonata.admin.extension', ['target' => 'sonata.page.admin.block'])
             ->args([
                 new ReferenceConfigurator('sonata.page.service.create_snapshot'),
             ])
 
+        ->alias(PageAdmin::class, 'sonata.page.admin.page')
+
         ->alias(BlockAdmin::class, 'sonata.page.admin.block')
```
eerison: Hey @VincentLanglet 

is it fine we use the `class` as a service and `the string` as an alias like this?

eerison:
Screenshot 2022-11-18 at 10 40 16

src/Resources/config/admin.php
VincentLanglet: This is not done this way for other bundles https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Resources/config/doctrine_orm_admin.php

And as far I remember there is a reason. I think this is not recommended for libraries.
So I would let this like it was before.
Maybe @jordisala1991 knows more about it.

eerison:
Ok, I'm using the service name as before ;) 

Screenshot 2022-11-18 at 11 02 12

Note: Sorry for the description I'm without creativity, we can improve this later, what do you think?

VincentLanglet (commented):
I'm looking at the phpstan failure.
Already opened https://github.com/doctrine/persistence/pull/315

eerison:
> I'm looking at the phpstan failure. Already opened [doctrine/persistence#315](https://github.com/doctrine/persistence/pull/315)

@VincentLanglet I bumped to ^3.1 :)

VincentLanglet:
> > I'm looking at the phpstan failure. Already opened [doctrine/persistence#315](https://github.com/doctrine/persistence/pull/315)
> 
> @VincentLanglet I bumped to ^3.1 :)

This is not needed, phpstan run on the latest version.
We can still support 3.0.

Revert the bump and the pr is perfect ;)

tests/Functional/Admin/BrowserTest.php
```diff
@@ -121,6 +121,7 @@ private function extractUniqId(Form $form): string
         \assert(\is_string($queryString));
 
         parse_str($queryString, $queryParameters);
+        \assert(\is_string($queryParameters['uniqid']));
```
eerison, line 124: @VincentLanglet I just let this change, because phpstan was complain about it!

VincentLanglet: Yes sure

VincentLanglet (approved):

VincentLanglet:
image

eerison:
> image

I didn't get any conflict 👀

VincentLanglet:
Squash was working, but couldn't rebase...

PR #1718, opened 2023-10-11
Feature/1713 - Switch slugify to symfony string

### LINKED ISSUE #1713: Switch cocur/slugify to symfony/string

eerison:
## Replace  slugify to symfony string

Closes #1713.

## Changelog

```markdown
### Added
- Handle slugs using `Symfony string`

### Deprecated
- Deprecated `slugify` in favor of symfony string
```

### DIFF
--- a/src/Entity/PageManager.php
+++ b/src/Entity/PageManager.php
@@ -13,12 +13,14 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Entity;
 
+use Cocur\Slugify\Slugify;
 use Cocur\Slugify\SlugifyInterface;
 use Doctrine\Persistence\ManagerRegistry;
 use Sonata\Doctrine\Entity\BaseEntityManager;
 use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Model\PageManagerInterface;
 use Sonata\PageBundle\Model\SiteInterface;
+use Symfony\Component\String\Slugger\SluggerInterface;
 
 /**
  * @extends BaseEntityManager<PageInterface>
@@ -35,7 +37,7 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|SluggerInterface $slugger,
         private array $defaults = [],
         private array $pageDefaults = []
     ) {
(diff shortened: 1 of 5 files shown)


src/Entity/PageManager.php
```diff
@@ -73,44 +74,7 @@ public function createWithDefaults(array $defaults = []): PageInterface
 
     public function fixUrl(PageInterface $page): void
     {
-        if ($page->isInternal()) {
-            $page->setUrl(null); // internal routes do not have any url ...
-
-            return;
```
VincentLanglet, line 79: if you put this in the handlePageurl, you lose the early return

eerison: Ahhhhh I got it now,  then I will keep internal check here and I will move just hybrid condition

eerison: Done

src/Entity/PageManager.php
```diff
@@ -35,7 +37,7 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|SluggerInterface $slugger,
```
VincentLanglet, line 40: You need to add a deprecation if the slugger passed is not instance of SluggerInterface.

And a next major comment to reduce the param to sluggerInterface.

VincentLanglet (changes requested):
With this last change we should be good. Thanks

src/Entity/PageManager.php
```diff
@@ -168,4 +139,55 @@ public function getHybridPages(SiteInterface $site): array
...
+            $url = $page->getCustomUrl() ?? '';
+            $page->setUrl('/'.ltrim($url, '/'));
+
+            return;
+        }
+
+        $slug = $page->getSlug();
+
+        // NEXT_MAJOR: Remove Slugify support
+        if (null === $slug && $this->slugger instanceof Slugify) {
```
VincentLanglet, line 163:
```suggestion
        if (null === $slug && $this->slugger instanceof SlugifyInterface) {
```
no?

and might be better to write
```suggestion
        if (null === $slug && !$this->slugger instanceof SluggerInterface) {
```

src/Entity/PageManager.php
```diff
@@ -168,4 +139,55 @@ public function getHybridPages(SiteInterface $site): array
...
+            @trigger_error(sprintf(
+                '%s is deprecated since version 4.8.0 and will be removed in 5.0, use %s instead of.',
+                Slugify::class,
```
VincentLanglet, line 166:
```suggestion
                SlugifyInterface::class,
```

PR #1727, opened 2023-11-22
Add a runtime for the symfony/runtime component.

### LINKED ISSUE #1604: Support the new Symfony 5.3 Runtime with multisite host_with_path
## Feature Request for 4.x

Runtime: https://symfony.com/doc/current/components/runtime.html
Multisite: https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/

> The purpose of the Runtime component is to abstract most of the application bootstrapping logic into so-called runtimes, allowing you to write generic front-controllers. This will make Symfony applications easier to maintain, because the front-controller code can be moved to a [Symfony Flex recipe](https://symfony.com/doc/3.4/setup/flex.html) managed automatically by Symfony.

from https://symfony.com/blog/new-in-symfony-5-3-runtime-component

Current implementation needs changes to index.php to be able to use host_with_path multisite and I'm guessing we would need page runtime now

GeraudBourdin:
## Symfony runtime for SonataPage

Add a  SonataPage runtime to avoid custom configuration in the public/index.php

use the configuration above  for testing this.

`composer require symfony/runtime`

**public/index.php :** 
```

    use App\Kernel;
    
    require_once dirname(__DIR__).'/vendor/autoload_runtime.php';

    return function (array $context) {
        return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
    };

```
**composer.json :** 
```
{
    // ... other composer configs ...
    "extra": {
        // ... other composer configs ...
        "runtime": {
            "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
            "multisite": "host_with_path_by_locale"
        }
    }
}
```
**multisite** is not manatory, by default **host_with_path_by_locale** is selected

Needs to refresh the autoload file: 
`composer dump-autoload`

**Documentation used** : [https://symfony.com/doc/current/components/runtime.html](https://symfony.com/doc/current/components/runtime.html)

Closes #1604 

## Changelog

```markdown
### Added
- Support for the symfony runtime component
```

### DIFF
--- a/docs/reference/advanced_configuration.rst
+++ b/docs/reference/advanced_configuration.rst
@@ -26,7 +26,7 @@ Full configuration options:
     sonata_page:
         skip_redirection: false # Skip asking Editor to redirect
         use_streamed_response: false
-        multisite: ~
+        multisite: ~ # Required
         ignore_route_patterns:
             # Defaults:
             - /(.*)admin(.*)/
--- a/docs/reference/advanced_configuration.rst
+++ b/docs/reference/advanced_configuration.rst
@@ -26,7 +26,7 @@ Full configuration options:
     sonata_page:
         skip_redirection: false # Skip asking Editor to redirect
         use_streamed_response: false
-        multisite: ~ # Required
+        multisite: ~
         ignore_route_patterns:
             # Defaults:
             - /(.*)admin(.*)/
--- a/docs/reference/multisite.rst
+++ b/docs/reference/multisite.rst
@@ -56,15 +56,14 @@ With this strategy it is possible to handle sites like :
 * http://sonata-project.net
 
 This strategy requires a dedicated runtime. So you need to modify your
-composer.json and declare ``SonataPageRuntime`` runtime and the multisite type you want to use.
+composer.json and declare ``SonataPageRuntime`` runtime.
 
 .. code-block:: json
 
 
         "extra": {
             "runtime": {
-                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
-                "multisite": "host_with_path"
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
             }
 
 
@@ -82,15 +81,13 @@ Host and Path By Locale Strategy
 --------------------------------
 
 This strategy requires a dedicated runtime. So you need to modify your
-composer.json and declare ``SonataPageRuntime`` runtime and the multisite type you want to use.
+composer.json and declare ``SonataPageRuntime`` runtime.
 
 .. code-block:: json
 
-
         "extra": {
             "runtime": {
-                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
-                "multisite": "host_with_path_by_locale"
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
             }
 
 
--- a/docs/reference/multisite.rst
+++ b/docs/reference/multisite.rst
@@ -55,38 +55,19 @@ With this strategy it is possible to handle sites like :
 * http://sonata-project.com/fr
 * http://sonata-project.net
 
-This strategy requires a dedicated ``RequestFactory`` object. So you need to alter the
-front controller to use the one provided by the ``PageBundle``.
+This strategy requires a dedicated runtime. So you need to modify your
+composer.json and declare ``SonataPageRuntime`` runtime and the multisite type you want to use.
 
-To do so, open ``index.php`` file and change the following parts::
+.. code-block:: json
 
-    // public/index.php
 
-    use App\Kernel;
-    use Symfony\Component\ErrorHandler\Debug;
-    use Sonata\PageBundle\Request\RequestFactory; # before: use Symfony\Component\HttpFoundation\Request;
+        "extra": {
+            "runtime": {
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
+                "multisite": "host_with_path"
+            }
 
-    require dirname(__DIR__).'/config/bootstrap.php';
 
-    if ($_SERVER['APP_DEBUG']) {
-        umask(0000);
-
-        Debug::enable();
-    }
-
-    if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
-        Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
-    }
-
-    if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? $_ENV['TRUSTED_HOSTS'] ?? false) {
-        Request::setTrustedHosts([$trustedHosts]);
-    }
-
-    $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
-    $request = RequestFactory::createFromGlobals('host_with_path'); // before: $request = Request::createFromGlobals();
-    $response = $kernel->handle($request);
-    $response->send();
-    $kernel->terminate($request, $response);
 
 The last action is to configure the ``sonata_page`` section as:
 
@@ -100,38 +81,18 @@ The last action is to configure the ``sonata_page`` section as:
 Host and Path By Locale Strategy
 --------------------------------
 
-This strategy requires a dedicated ``RequestFactory`` object. So you need to alter the
-front controller to use the one provided by the ``PageBundle``.
+This strategy requires a dedicated runtime. So you need to modify your
+composer.json and declare ``SonataPageRuntime`` runtime and the multisite type you want to use.
 
-To do so, open ``index.php`` file and change the following parts::
+.. code-block:: json
 
-    // public/index.php
 
-    use App\Kernel;
-    use Symfony\Component\ErrorHandler\Debug;
-    use Sonata\PageBundle\Request\RequestFactory; # before: use Symfony\Component\HttpFoundation\Request;
+        "extra": {
+            "runtime": {
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
+                "multisite": "host_with_path_by_locale"
+            }
 
-    require dirname(__DIR__).'/config/bootstrap.php';
-
-    if ($_SERVER['APP_DEBUG']) {
-        umask(0000);
-
-        Debug::enable();
-    }
-
-    if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
-        Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
-    }
-
-    if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? $_ENV['TRUSTED_HOSTS'] ?? false) {
-        Request::setTrustedHosts([$trustedHosts]);
-    }
-
-    $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
-    $request = RequestFactory::createFromGlobals('host_with_path_by_locale'); // before: $request = Request::createFromGlobals();
-    $response = $kernel->handle($request);
-    $response->send();
-    $kernel->terminate($request, $response);
 
 The last action is to configure the ``sonata_page`` section as:
 
--- a/docs/reference/multisite.rst
+++ b/docs/reference/multisite.rst
@@ -63,7 +63,7 @@ composer.json and declare ``SonataPagePathRuntime`` runtime.
 
         "extra": {
             "runtime": {
-                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPagePathRuntime"
             }
 
 Update your autoloader file with ``composer dump-autoload``.
@@ -87,7 +87,7 @@ composer.json and declare ``SonataPagePathRuntime`` runtime.
 
         "extra": {
             "runtime": {
-                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPagePathRuntime"
             }
 
 Update your autoloader file with ``composer dump-autoload``.
--- a/docs/reference/multisite.rst
+++ b/docs/reference/multisite.rst
@@ -56,7 +56,7 @@ With this strategy it is possible to handle sites like :
 * http://sonata-project.net
 
 This strategy requires a dedicated runtime. So you need to modify your
-composer.json and declare ``SonataPageRuntime`` runtime.
+composer.json and declare ``SonataPagePathRuntime`` runtime.
 
 .. code-block:: json
 
@@ -66,7 +66,7 @@ composer.json and declare ``SonataPageRuntime`` runtime.
                 "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
             }
 
-
+Update your autoloader file with ``composer dump-autoload``.
 
 The last action is to configure the ``sonata_page`` section as:
 
@@ -81,7 +81,7 @@ Host and Path By Locale Strategy
 --------------------------------
 
 This strategy requires a dedicated runtime. So you need to modify your
-composer.json and declare ``SonataPageRuntime`` runtime.
+composer.json and declare ``SonataPagePathRuntime`` runtime.
 
 .. code-block:: json
 
@@ -90,6 +90,7 @@ composer.json and declare ``SonataPageRuntime`` runtime.
                 "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
             }
 
+Update your autoloader file with ``composer dump-autoload``.
 
 The last action is to configure the ``sonata_page`` section as:
 
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -85,7 +85,7 @@ final class Configuration implements ConfigurationInterface
                 ->end()
                 ->scalarNode('multisite')
                     ->info('For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
-                    ->defaultValue('host')
+                    ->isRequired()
                     ->validate()
                         ->ifNotInArray(['host', 'host_by_locale', 'host_with_path', 'host_with_path_by_locale'])
                         ->thenInvalid('Invalid multisite configuration %s. For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -85,7 +85,7 @@ final class Configuration implements ConfigurationInterface
                 ->end()
                 ->scalarNode('multisite')
                     ->info('For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
-                    ->isRequired()
+                    ->defaultValue('host')
                     ->validate()
                         ->ifNotInArray(['host', 'host_by_locale', 'host_with_path', 'host_with_path_by_locale'])
                         ->thenInvalid('Invalid multisite configuration %s. For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -150,7 +150,6 @@ final class Transformer implements TransformerInterface
             /**
              * @var PageContent $content
              */
-            // @phpstan-ignore-next-line
             $content = array_filter($data, static fn ($v) => null !== $v);
         }
 
--- a/src/Entity/Transformer.php
+++ b/src/Entity/Transformer.php
@@ -150,6 +150,7 @@ final class Transformer implements TransformerInterface
             /**
              * @var PageContent $content
              */
+            // @phpstan-ignore-next-line
             $content = array_filter($data, static fn ($v) => null !== $v);
         }
 
--- a/src/Request/RequestFactory.php
+++ b/src/Request/RequestFactory.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Request;
 
+use Sonata\PageBundle\Runtime\SonataPagePathRuntime;
 use Symfony\Component\HttpFoundation\Request;
 
 final class RequestFactory
@@ -27,6 +28,7 @@ final class RequestFactory
     ];
 
     /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
      * @param array<string, mixed> $parameters
      * @param array<string, mixed> $cookies
      * @param array<string, mixed> $files
@@ -43,20 +45,50 @@ final class RequestFactory
         array $server = [],
         $content = null
     ): Request {
+        @trigger_error(sprintf(
+            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
+            .'  Use "%s::%s()" instead.',
+            __METHOD__,
+            SonataPagePathRuntime::class,
+            __FUNCTION__
+        ), \E_USER_DEPRECATED);
+
         self::configureFactory($type);
 
         return \call_user_func_array([self::getClass($type), 'create'], [$uri, $method, $parameters, $cookies, $files, $server, $content]);
     }
 
+    /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
+     */
     public static function createFromGlobals(string $type): Request
     {
+        @trigger_error(sprintf(
+            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
+            .'  Use "%s::%s()" instead.',
+            __METHOD__,
+            SonataPagePathRuntime::class,
+            __FUNCTION__
+        ), \E_USER_DEPRECATED);
+
         self::configureFactory($type);
 
         return \call_user_func_array([self::getClass($type), 'createFromGlobals'], []);
     }
 
+    /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
+     */
     private static function configureFactory(string $type): void
     {
+        @trigger_error(sprintf(
+            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
+            .'  Use "%s::%s()" instead.',
+            __METHOD__,
+            SonataPagePathRuntime::class,
+            __FUNCTION__
+        ), \E_USER_DEPRECATED);
+
         if (!\in_array($type, ['host_with_path', 'host_with_path_by_locale'], true)) {
             return;
         }
@@ -78,10 +110,19 @@ final class RequestFactory
     }
 
     /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
      * @return class-string<Request>
      */
     private static function getClass(string $type): string
     {
+        @trigger_error(sprintf(
+            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
+            .'  Use "%s::%s()" instead.',
+            __METHOD__,
+            SonataPagePathRuntime::class,
+            __FUNCTION__
+        ), \E_USER_DEPRECATED);
+
         if (!\array_key_exists($type, self::$types)) {
             throw new \RuntimeException('invalid type');
         }
--- a/src/Request/RequestFactory.php
+++ b/src/Request/RequestFactory.php
@@ -16,6 +16,11 @@ namespace Sonata\PageBundle\Request;
 use Sonata\PageBundle\Runtime\SonataPagePathRuntime;
 use Symfony\Component\HttpFoundation\Request;
 
+/**
+ * NEXT_MAJOR: Remove this class.
+ *
+ * @deprecated since sonata-project/page-bundle 4.6.2
+ */
 final class RequestFactory
 {
     /**
@@ -29,6 +34,7 @@ final class RequestFactory
 
     /**
      * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
+     *
      * @param array<string, mixed> $parameters
      * @param array<string, mixed> $cookies
      * @param array<string, mixed> $files
@@ -76,19 +82,8 @@ final class RequestFactory
         return \call_user_func_array([self::getClass($type), 'createFromGlobals'], []);
     }
 
-    /**
-     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
-     */
     private static function configureFactory(string $type): void
     {
-        @trigger_error(sprintf(
-            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
-            .'  Use "%s::%s()" instead.',
-            __METHOD__,
-            SonataPagePathRuntime::class,
-            __FUNCTION__
-        ), \E_USER_DEPRECATED);
-
         if (!\in_array($type, ['host_with_path', 'host_with_path_by_locale'], true)) {
             return;
         }
@@ -110,19 +105,10 @@ final class RequestFactory
     }
 
     /**
-     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
      * @return class-string<Request>
      */
     private static function getClass(string $type): string
     {
-        @trigger_error(sprintf(
-            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
-            .'  Use "%s::%s()" instead.',
-            __METHOD__,
-            SonataPagePathRuntime::class,
-            __FUNCTION__
-        ), \E_USER_DEPRECATED);
-
         if (!\array_key_exists($type, self::$types)) {
             throw new \RuntimeException('invalid type');
         }
--- a/src/Request/RequestFactory.php
+++ b/src/Request/RequestFactory.php
@@ -19,7 +19,7 @@ use Symfony\Component\HttpFoundation\Request;
 /**
  * NEXT_MAJOR: Remove this class.
  *
- * @deprecated since sonata-project/page-bundle 4.6.2
+ * @deprecated since sonata-project/page-bundle 4.7.0
  */
 final class RequestFactory
 {
@@ -33,7 +33,7 @@ final class RequestFactory
     ];
 
     /**
-     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
+     * @deprecated since sonata-project/page-bundle 4.7.0, to be removed in 5.0.
      *
      * @param array<string, mixed> $parameters
      * @param array<string, mixed> $cookies
@@ -52,7 +52,7 @@ final class RequestFactory
         $content = null
     ): Request {
         @trigger_error(sprintf(
-            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
+            'The method "%s()" is deprecated since sonata-project/page-bundle 4.7.0 and will be removed in 5.0.'
             .'  Use "%s::%s()" instead.',
             __METHOD__,
             SonataPagePathRuntime::class,
@@ -65,12 +65,12 @@ final class RequestFactory
     }
 
     /**
-     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
+     * @deprecated since sonata-project/page-bundle 4.7.0, to be removed in 5.0.
      */
     public static function createFromGlobals(string $type): Request
     {
         @trigger_error(sprintf(
-            'The method "%s()" is deprecated since sonata-project/page-bundle 4.6.2 and will be removed in 5.0.'
+            'The method "%s()" is deprecated since sonata-project/page-bundle 4.7.0 and will be removed in 5.0.'
             .'  Use "%s::%s()" instead.',
             __METHOD__,
             SonataPagePathRuntime::class,
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -28,7 +28,7 @@ class SonataPageRuntime extends SymfonyRuntime
      */
     public function __construct(array $options = [])
     {
-        $this->multisite = $options['multisite'] ?? 'host_with_path_by_locale';
+        $this->multisite = $options['multisite'] ?? 'host';
         parent::__construct($options);
     }
 
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -28,7 +28,11 @@ class SonataPageRuntime extends SymfonyRuntime
      */
     public function __construct(array $options = [])
     {
-        $this->multisite = $options['multisite'] ?? 'host';
+        if (!isset($options['multisite'])) {
+            throw new \InvalidArgumentException('The option multisite is required');
+        }
+
+        $this->multisite = $options['multisite'];
         parent::__construct($options);
     }
 
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -48,6 +48,7 @@ class SonataPageRuntime extends SymfonyRuntime
         }
 
         $this->multisite = $options['multisite'];
+        unset($options['multisite']);
         parent::__construct($options);
     }
 
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -24,7 +24,22 @@ class SonataPageRuntime extends SymfonyRuntime
     private string $multisite;
 
     /**
-     * @param array<mixed> $options
+     * @param  array{
+     *      debug?: bool|null,
+     *      env?: string|null,
+     *      disable_dotenv?: bool|null,
+     *      project_dir?: string|null,
+     *      prod_envs?: string[]|null,
+     *      dotenv_path?: string|null,
+     *      test_envs?: string[]|null,
+     *      use_putenv?: bool|null,
+     *      runtimes?: array<mixed>|null,
+     *      error_handler?: string|false,
+     *      env_var_name?: string,
+     *      debug_var_name?: string,
+     *      dotenv_overload?: bool|null,
+     *      multisite: string,
+     *    }  $options  $options
      */
     public function __construct(array $options = [])
     {
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -38,7 +38,7 @@ class SonataPageRuntime extends SymfonyRuntime
      *      env_var_name?: string,
      *      debug_var_name?: string,
      *      dotenv_overload?: bool|null,
-     *      multisite: string,
+     *      multisite?: string,
      *    }  $options  $options
      */
     public function __construct(array $options = [])
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -13,51 +13,20 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Runtime;
 
-use Sonata\PageBundle\Request\RequestFactory;
+use Sonata\PageBundle\Request\SiteRequest;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
 use Symfony\Component\Runtime\RunnerInterface;
 use Symfony\Component\Runtime\SymfonyRuntime;
 
-class SonataPageRuntime extends SymfonyRuntime
+final class SonataPageRuntime extends SymfonyRuntime
 {
-    private string $multisite;
-
-    /**
-     * @param  array{
-     *      debug?: bool|null,
-     *      env?: string|null,
-     *      disable_dotenv?: bool|null,
-     *      project_dir?: string|null,
-     *      prod_envs?: string[]|null,
-     *      dotenv_path?: string|null,
-     *      test_envs?: string[]|null,
-     *      use_putenv?: bool|null,
-     *      runtimes?: array<mixed>|null,
-     *      error_handler?: string|false,
-     *      env_var_name?: string,
-     *      debug_var_name?: string,
-     *      dotenv_overload?: bool|null,
-     *      multisite?: string,
-     *    }  $options  $options
-     */
-    public function __construct(array $options = [])
-    {
-        if (!isset($options['multisite'])) {
-            throw new \InvalidArgumentException('The option multisite is required');
-        }
-
-        $this->multisite = $options['multisite'];
-        unset($options['multisite']);
-        parent::__construct($options);
-    }
-
     public function getRunner(?object $application): RunnerInterface
     {
-        if ($application instanceof HttpKernelInterface) {
-            return new HttpKernelRunner($application, RequestFactory::createFromGlobals($this->multisite));
+        if (!$application instanceof HttpKernelInterface) {
+            return parent::getRunner($application);
         }
 
-        return parent::getRunner($application);
+        return new HttpKernelRunner($application, SiteRequest::createFromGlobals());
     }
 }
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\PageBundle\Runtime;
 
+use http\Exception\InvalidArgumentException;
 use Sonata\PageBundle\Request\RequestFactory;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
@@ -28,7 +29,11 @@ class SonataPageRuntime extends SymfonyRuntime
      */
     public function __construct(array $options = [])
     {
-        $this->multisite = $options['multisite'] ?? 'host';
+        if (!isset($options['multisite'])) {
+            throw new InvalidArgumentException('The option multisite is required');
+        }
+
+        $this->multisite = $options['multisite'];
         parent::__construct($options);
     }
 
--- a/src/Runtime/SonataPageRuntime.php
+++ b/src/Runtime/SonataPageRuntime.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Runtime;
+
+use Sonata\PageBundle\Request\RequestFactory;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
+use Symfony\Component\Runtime\RunnerInterface;
+use Symfony\Component\Runtime\SymfonyRuntime;
+
+class SonataPageRuntime extends SymfonyRuntime
+{
+    private string $multisite;
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function __construct(array $options = [])
+    {
+        $this->multisite = $options['multisite'] ?? 'host_with_path_by_locale';
+        parent::__construct($options);
+    }
+
+    public function getRunner(?object $application): RunnerInterface
+    {
+        if ($application instanceof HttpKernelInterface) {
+            return new HttpKernelRunner($application, RequestFactory::createFromGlobals($this->multisite));
+        }
+
+        return parent::getRunner($application);
+    }
+}
--- a/tests/Runtime/SonataPageRuntimeTest.php
+++ b/tests/Runtime/SonataPageRuntimeTest.php
@@ -6,10 +6,10 @@ namespace Sonata\PageBundle\Tests\Runtime;
 
 use PHPUnit\Framework\TestCase;
 use Sonata\PageBundle\Runtime\SonataPageRuntime;
+use Symfony\Component\Console\Command\Command;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\Runtime\Runner\Symfony\ConsoleApplicationRunner;
 use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
-use Symfony\Component\Console\Command\Command;
 
 class SonataPageRuntimeTest extends TestCase
 {
@@ -18,12 +18,6 @@ class SonataPageRuntimeTest extends TestCase
     {
         $this->expectException(\InvalidArgumentException::class);
         new SonataPageRuntime();
-    }
-
-    public function testConstructorWithMultisite(): void
-    {
-        $runtime = new SonataPageRuntime(['multisite' => 'host_with_path_by_locale']);
-        static::assertInstanceOf(SonataPageRuntime::class, $runtime);
     }
 
     public function testGetRunnerWithHttpKernel(): void
--- a/tests/Runtime/SonataPageRuntimeTest.php
+++ b/tests/Runtime/SonataPageRuntimeTest.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sonata\PageBundle\Tests\Runtime;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\PageBundle\Runtime\SonataPageRuntime;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Runtime\Runner\Symfony\ConsoleApplicationRunner;
+use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
+use Symfony\Component\Console\Command\Command;
+
+class SonataPageRuntimeTest extends TestCase
+{
+
+    public function testConstructorThrowsExceptionWithoutMultisite(): void
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        new SonataPageRuntime();
+    }
+
+    public function testConstructorWithMultisite(): void
+    {
+        $runtime = new SonataPageRuntime(['multisite' => 'host_with_path_by_locale']);
+        $this->assertInstanceOf(SonataPageRuntime::class, $runtime);
+    }
+
+    public function testGetRunnerWithHttpKernel(): void
+    {
+        $kernelMock = $this->createMock(HttpKernelInterface::class);
+        $runtime = new SonataPageRuntime(['multisite' => 'host_with_path_by_locale']);
+
+        $runner = $runtime->getRunner($kernelMock);
+        $this->assertInstanceOf(HttpKernelRunner::class, $runner);
+    }
+
+    public function testGetRunnerWithNonHttpKernelApplication(): void
+    {
+        $command = new Command('app:test-cmd');
+        $runtime = new SonataPageRuntime(['multisite' => 'host_with_path_by_locale']);
+
+        $runner = $runtime->getRunner($command);
+        $this->assertInstanceOf(ConsoleApplicationRunner::class, $runner);
+    }
+}
--- a/tests/Runtime/SonataPageRuntimeTest.php
+++ b/tests/Runtime/SonataPageRuntimeTest.php
@@ -22,16 +22,10 @@ use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
 
 class SonataPageRuntimeTest extends TestCase
 {
-    public function testConstructorThrowsExceptionWithoutMultisite(): void
-    {
-        $this->expectException(\InvalidArgumentException::class);
-        new SonataPageRuntime();
-    }
-
     public function testGetRunnerWithHttpKernel(): void
     {
         $kernelMock = $this->createMock(HttpKernelInterface::class);
-        $runtime = new SonataPageRuntime(['multisite' => 'host_with_path_by_locale']);
+        $runtime = new SonataPageRuntime();
 
         $runner = $runtime->getRunner($kernelMock);
         static::assertInstanceOf(HttpKernelRunner::class, $runner);
@@ -40,7 +34,7 @@ class SonataPageRuntimeTest extends TestCase
     public function testGetRunnerWithNonHttpKernelApplication(): void
     {
         $command = new Command('app:test-cmd');
-        $runtime = new SonataPageRuntime(['multisite' => 'host_with_path_by_locale']);
+        $runtime = new SonataPageRuntime();
 
         $runner = $runtime->getRunner($command);
         static::assertInstanceOf(ConsoleApplicationRunner::class, $runner);
--- a/tests/Runtime/SonataPageRuntimeTest.php
+++ b/tests/Runtime/SonataPageRuntimeTest.php
@@ -2,6 +2,15 @@
 
 declare(strict_types=1);
 
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace Sonata\PageBundle\Tests\Runtime;
 
 use PHPUnit\Framework\TestCase;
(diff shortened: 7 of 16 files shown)


VincentLanglet:
ci is failing

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,43 @@
...
+class SonataPageRuntime extends SymfonyRuntime
+{
+    private string $multisite;
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function __construct(array $options = [])
+    {
+        $this->multisite = $options['multisite'] ?? 'host_with_path_by_locale';
```
VincentLanglet: In the configuration, we have 
```
->scalarNode('multisite')
                    ->info('For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
                    ->isRequired()
                    ->validate()
                        ->ifNotInArray(['host', 'host_by_locale', 'host_with_path', 'host_with_path_by_locale'])
                        ->thenInvalid('Invalid multisite configuration %s. For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
                    ->end()
                ->end()
```

So No default value for multisite => Why should we provide one in the runtime and not in the config ? Should we be consistent and not having one here.
Also, in the doc we have `multisite: host` so it seems like we promote this value rather than others as default.
We might
- Add the host as default value in the config AND in the runtime (?)

VincentLanglet (changes requested):
The doc https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/ need to be updated too.

src/Runtime/SonataPageRuntime.php
GeraudBourdin: So we can remove the default value in SonataPageRuntime ? and the multisite param should be mandatory. 

For the configuration of ->scalarNode('multisite') it remains unchanged as this is the actual use of this bundle. 

We have to configure the request object in index.php, and the multisite option in the config is in charge of the services setup ? The both configurations are mandatory, so i do not know what the good strategy should be on this subject.

Before going further in the documentation part, it would be great to have more comment on this subject as i consider this PR only as a first step and i don't really know if it suits the needs.

VincentLanglet: > So we can remove the default value in SonataPageRuntime ? and the multisite param should be mandatory.
> 
> For the configuration of ->scalarNode('multisite') it remain unchanged as this is the actual use of this bundle.

Imho, either we use `?? 'host'` and we had `host` as the default value value of the `->scalarNode('multisite')`
either we don't use default value and `multisite` is mandatory (and we untouch the scalarNode).

This will give consistency.
 
> Before going further in the documentation part, it would be great to have more comment on this subject as i consider this PR only as a first step and i don't really know if it suits the needs.

Ping @haivala then which open the issue https://github.com/sonata-project/SonataPageBundle/issues/1604
Also ping @eerison

eerison: well if it gonna be added then we should deprecate the current implementation and update the doc

then it will be easier to reproduce the steps to use symfony runtime.

Btw, are you using symfony runtime in your project instead of update front controller?

GeraudBourdin: @eerison thanks to be part of this discussion. 

here is my actual index.php that i would like to avoid  :

```php

use App\Kernel;
use Sonata\PageBundle\Request\RequestFactory;

require_once dirname(__DIR__).'/vendor/autoload_runtime.php';

return function (array $context) {
    $request = RequestFactory::createFromGlobals('host_with_path_by_locale');
    $kernel = new Kernel($context['APP_ENV'], $context['APP_DEBUG']);
    $response = $kernel->handle($request);
    $response->send();
    $kernel->terminate($request, $response);
};
```

eerison: Nice <3

But isn't possible you add this `$request = RequestFactory::createFromGlobals('host_with_path_by_locale');`

into the Sonata runtime? because as far as I know the idea is not update index.php file, you should just create your runtime and  enable it and it will work, or am I missing something?

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,43 @@
...
+class SonataPageRuntime extends SymfonyRuntime
+{
+    private string $multisite;
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function __construct(array $options = [])
+    {
+        $this->multisite = $options['multisite'] ?? 'host';
```
eerison: Should be interesting you create different runtimes then the dev can load the runtime that he wants, does it make sense?

eerison (changes requested):

src/Runtime/SonataPageRuntime.php
GeraudBourdin: this is the goal of this MR. 
check the first post  : https://github.com/sonata-project/SonataPageBundle/pull/1727#issue-2005839773
my last post was what i am doing right now and this is what i would like to avoid. Keep the default symfony index.php
so the request part is done in the SonataPageRuntime, the multisite part is taken from the composer file

src/Runtime/SonataPageRuntime.php
GeraudBourdin: this is an alternative. So one Runtime by multisite or one Runtime with the multisite taken by env (with the composer extra param) wich was the direction i choose. 
```json

{
    // ... other composer configs ...
    "extra": {
        // ... other composer configs ...
        "runtime": {
            "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
            "multisite": "host_with_path_by_locale"
        }
    }
}

```
In my point of view this is the same. The only goal is to keep the index.php clean. :) 

That s why any comments/directions are greatly apreciated.

VincentLanglet: We could have more option later. I think it's better to have one runtime with config like this rather than multiple class :)

VincentLanglet (commented):
Can you change `isRequired` to `->defaultValue('host')` in the Config too ?

src/Runtime/SonataPageRuntime.php
eerison: ahhh I see I got it now ;)

GeraudBourdin:
> Can you change `isRequired` to `->defaultValue('host')` in the Config too ?

done

VincentLanglet:
> > Can you change `isRequired` to `->defaultValue('host')` in the Config too ?
> 
> done

Thanks.

To me, code is ready. But now we need to update the doc.

In https://github.com/sonata-project/SonataPageBundle/blob/c0b2c445d423b8880ca53be8ce20442a6d29c2e8/docs/reference/advanced_configuration.rst#L29
the part `multisite: ~ # Required` need to be removed (and maybe in some other files).

This page need to be update too
https://github.com/sonata-project/SonataPageBundle/blob/c0b2c445d423b8880ca53be8ce20442a6d29c2e8/docs/reference/multisite.rst#L2

GeraudBourdin:
> > > Can you change `isRequired` to `->defaultValue('host')` in the Config too ?
> > 
> > 
> > done
> 
> Thanks.
> 
> To me, code is ready. But now we need to update the doc.
> 
> In
> 
> https://github.com/sonata-project/SonataPageBundle/blob/c0b2c445d423b8880ca53be8ce20442a6d29c2e8/docs/reference/advanced_configuration.rst#L29
> 
> 
> the part `multisite: ~ # Required` need to be removed (and maybe in some other files).
> This page need to be update too
> 
> https://github.com/sonata-project/SonataPageBundle/blob/c0b2c445d423b8880ca53be8ce20442a6d29c2e8/docs/reference/multisite.rst#L2

Doc updated. Hope that's ok.

VincentLanglet (dismissed):

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,43 @@
...
+        parent::__construct($options);
+    }
+
+    public function getRunner(?object $application): RunnerInterface
+    {
+        if ($application instanceof HttpKernelInterface) {
+            return new HttpKernelRunner($application, RequestFactory::createFromGlobals($this->multisite));
```
eerison: Hey @GeraudBourdin do we still this request factory?

eerison: I guess we could use symfony request directly hire. 


And deprecate this RequestFactory class, this way we can avoid people use the Old way. It will be easier for maintaining 


If you look [here](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Request/RequestFactory.php#L64) it is just wrapping symfony request and use SiteRequest

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace Sonata\PageBundle\Runtime;
+
+use Sonata\PageBundle\Request\RequestFactory;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
+use Symfony\Component\Runtime\RunnerInterface;
+use Symfony\Component\Runtime\SymfonyRuntime;
+
```
eerison: It would  be awesome if you could write test for this class.

src/Runtime/SonataPageRuntime.php
GeraudBourdin: if you take a look at SiteRequestInterface you'll see that there is pathInfo and baseUrl is used. I think this is because Symfony\Component\HttpFoundation\Request is protected ? Si my guess is that you still need it.

src/Runtime/SonataPageRuntime.php
GeraudBourdin: I don't really know how and what to test on a runtime context. as HttpKernelRunner is allready tested in symfony and createFromGlobals too.

src/Runtime/SonataPageRuntime.php
eerison: to make this change simple you could just deprecate `createFromGlobals()` and move the logic for here

it will be like this

```php
        Request::setFactory(
            /**
             * @param string|resource|null $content
             */
            static fn (
                array $query = [],
                array $request = [],
                array $attributes = [],
                array $cookies = [],
                array $files = [],
                array $server = [],
                $content = null
            ) => new SiteRequest($query, $request, $attributes, $cookies, $files, $server, $content)
        );
        
        return SiteRequest::createFromGlobals();
```

and this runtime just need to be set if the user want to use `host_with_path` or `host_with_path_by_locale`

because when you pass `host`, it uses the default Request from symfony then this runtime is not needed!

Then we could rename this runtime for something like `SonataPagePathRuntime`. it will be nice because for simple SonataPage usage you don't need to do anything, just for `path` or `by_locale`.

Note: in this case we don't deprecate the whole `RequestFactory`, just `createFromGlobals`, because there is a `create` method and I saw it is used in others places, then to not complicate we can keep this method there!

make sense?

eerison: Since host won't be used for runtime we don't need this line

```php
$this->multisite = $options['multisite'] ?? 'host';
```

GeraudBourdin: > Since host won't be used for runtime we don't need this line
> 
> ```
> $this->multisite = $options['multisite'] ?? 'host';
> ```

in that case you don't have to configure runtime in the  composer.json. as usual ? i only use host_with_path_by_locale so it s a guess  :)

VincentLanglet: > Since host won't be used for runtime we don't need this line
> 
> ```
> $this->multisite = $options['multisite'] ?? 'host';
> ```

I proposed this, because:
- I don't know a lot this bundle
- This is the value we seem to recommend by default in the doc https://github.com/sonata-project/SonataPageBundle/blob/4.x/docs/reference/installation.rst#sonatapagebundle-configuration
- This is the first value in the doc https://github.com/sonata-project/SonataPageBundle/blob/4.x/docs/reference/multisite.rst

The proposal was 
```
$this->multisite = $options['multisite'] ?? 'host_with_path_by_locale';
```
but this had lack of consistency with my previous points.

src/Runtime/SonataPageRuntime.php
VincentLanglet: No idea either, that's why I won't enforce a test, (it would just be awesome ^^)

src/Runtime/SonataPageRuntime.php
GeraudBourdin: $options['multisite'] is needed for `host_with_path` and `host_with_path_by_locale` that's why my first PR included `host_with_path_by_locale` by default as you have to pick one default, so i picked mine  ^^

So now, i do not know what to do :)

VincentLanglet: > $options['multisite'] is needed for `host_with_path` and `host_with_path_by_locale` that's why my first PR included `host_with_path_by_locale` by default as you have to pick one default, so i picked mine ^^

Oh ! I see...
My bad, I thought this will be needed for the 4 strategy. The explanation help me a lot.

Would it make sens to not have a default value then ?
```
if (!isset('multisite')) {
    throws new InvalidArgument('The option multisite is required');
}
$this->multisite = $options['mutisite'];
```

Since no option is promoted in the sonata configuration between host_with_path and host_with_path_by_locale, this will be consistent in the runtime.

src/Runtime/SonataPageRuntime.php
GeraudBourdin: it s a challange and it's cool,  so may be it can be another ticket just for this task  ?

src/Runtime/SonataPageRuntime.php
GeraudBourdin: yes that make sense. In my understanding, we need the request, and the request can only be set at the kernel init because it's protected you canno't alter it after. So i think runtime will be allways needed for path and path locals.  Some changes can be done to avoid this mechanism but i think it's a bigger work.

GeraudBourdin: by the way i like the default value too, one of the too will not have to set the multisite option.

eerison: > if (!isset('multisite')) {
    throws new InvalidArgument('The option multisite is required');
}
$this->multisite = $options['mutisite'];

if you check [Requestfactory](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Request/RequestFactory.php#L51) this multsite option is used just to call `SiteRequest` or `Symfony Request`,  then we don't need `host` (because it is the default behaviour, calling Symfony Request), just the others `host_with_path` and `host_with_path_by_locale`, and [both options](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Request/RequestFactory.php#L23) use SiteRequest, then for runtime it is unnecessary.

then I would suggest to test

1. check if it works without any runtime configured
2. Using SonataPagePathRuntime just for `host_with_path` and `host_with_path_by_locale`


> So now, i do not know what to do :)

```php
    public function getRunner(?object $application): RunnerInterface
    {           
        if (!$application instanceof HttpKernelInterface) {
            return parent::getRunner($application);
        }
        
        //NOTE: I just pick up this from the RequestFactory, But I don't know if this this is needed.
        // Because you are returning the correct runtime, then "maybe" it was necessary for old versions only
        // But well just know testing
        Request::setFactory(
            /**
             * @param string|resource|null $content
             */
            static fn (
                array $query = [],
                array $request = [],
                array $attributes = [],
                array $cookies = [],
                array $files = [],
                array $server = [],
                $content = null
            ) => new SiteRequest($query, $request, $attributes, $cookies, $files, $server, $content)
        );
        
        return new HttpKernelRunner($application, SiteRequest::createFromGlobals());
    }
```

Edit: I suspect that this Request:setFactory is not needed.
Because into the symfony code it is setting here: https://github.com/symfony/http-foundation/blob/6.3/Request.php#L302

Note: SiteRequest extends SymfonyRequest ;)

Edit 2: I really believe it is not needed 😄 , because it is set here: https://github.com/symfony/http-foundation/blob/6.3/Request.php#L1989, could you test without this ?

then in the second test could be like this 🤯 

```php
  public function getRunner(?object $application): RunnerInterface
    {           
        if (!$application instanceof HttpKernelInterface) {
            return parent::getRunner($application);
        }
        
        return new HttpKernelRunner($application, SiteRequest::createFromGlobals());
    }
```

src/Runtime/SonataPageRuntime.php
eerison: I just ask for test because it give us confidence to merge future fix or features regarding this feature.
I switched the company and my current job is not using this bundle. then I do not have a real project to test it :/ .

eerison: But in other hand I do not want to push you down because of this, Because I know it takes time to be familiar with the code, then I would say it is just a wish for now :)

src/Runtime/SonataPageRuntime.php
GeraudBourdin: We need setPathInfo Symfony\Component\HttpFoundation\Request pathInfo is protected so SonataBundle cannot alter it after HttpKernelRunner it have to be changed before. So Runner is needed in thos two cases. (have a look at src/Site/HostPathSiteSelector.php )

May be i'am wrong. You can make a fork whith what you have in mind  i can test it on my side with my host_with_path_by_locale  app. This could help validate what you says. In fact this would be really cool to get rid of this setup and get the same for all. By the way, this mr break nothing anywhere as it is just a new way of configuration and is only used in case you want thoses two multisites config.
plus, \Sonata\PageBundle\Request\SiteRequestInterface::29 will have to be modifyed that was an easy way to ensure config is ok.

eerison: > We need setPathInfo Symfony\Component\HttpFoundation\Request pathInfo is protected so SonataBundle cannot alter it after HttpKernelRunner it have to be changed before. So Runner is needed in thos two cases. (have a look at src/Site/HostPathSiteSelector.php )


But that is correct!

[HostPathSiteSelector](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Site/HostPathSiteSelector.php) and HostPathByLocaleSiteSelector uses [SiteRequest](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Request/SiteRequest.php) and it contains setPathInfo and setBaseUrl, for this reason when use  `host_with_path` and `host_with_path_by_locale` it calls SiteRequest.

in the other hand if you look into [HostSiteSelector](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Site/HostSiteSelector.php), there isn't any setPathInfo because in theory it uses Symfony request.


> May be i'am wrong. You can make a fork whith what you have in mind i can test it on my side with my host_with_path_by_locale app. This could help validate what you says. In fact this would be really cool to get rid of this setup and get the same for all. By the way, this mr break nothing anywhere as it is just a new way of configuration and is only used in case you want thoses two multisites config.
plus, \Sonata\PageBundle\Request\SiteRequestInterface::29 will have to be modifyed that was an easy way to ensure config is ok.


I will give a try on Monday ;)

did you try my example and it doesn't work?

GeraudBourdin:
```
        if (!$request instanceof SiteRequestInterface) {
            throw new \RuntimeException('You must change the main Request object in the front controller (app.php) in order to use the `host_with_path_by_locale` strategy');
        }
```
will block

eerison: > ```
>         if (!$request instanceof SiteRequestInterface) {
>             throw new \RuntimeException('You must change the main Request object in the front controller (app.php) in order to use the `host_with_path_by_locale` strategy');
>         }
> ```
> 
> will block

could you give a context please, where is it located?

eerison: Ok I found it https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Site/HostPathSiteSelector.php
why will it block? block what?

VincentLanglet: > by the way i like the default value too, one of the too will not have to set the multisite option.

I understand this. But on the "open source perspective" I currently don't see a reason for promoting one default value rather the other. And it doesn't seems to hard for both users to pass the value they want (cause they still need an entry in the composer json to use the Runtime)

I'll let you try the @eerison recommendation next week if you want to ; and then I think we could merge this without a default value.

Also an argument against the default value is the fact that
- Adding a default value is BC
- Removing a default value is not BC

GeraudBourdin: @eerison i am not sure to understand what you want. 

to resume my point of view : 
SonataPagePathRuntime is not needed for `host` and `host_by_locale`. user need to use regular symfony runtime. No need of this PR. No need to change the composer.json. `createFromGlobals` can be done later in the process. 

`host_with_path` and `host_with_path_by_locale`  need to change `Request` object before setting up the  `HttpKernelRunner`. Why ? because `pathInfo` need to be accessible and written in the Request object from the `SiteSelector`. See \Sonata\PageBundle\Site\HostPathByLocaleSiteSelector:58  
If` $request->setPathInfo` is not done, (from /fr/some-route  to /some-route) you break the feature, no route found.

Why override the Request object  ?  Because setPathInfo is not in writable in Request object. You cannot alter Request object after the runtime. So we need to add this method. 

Please explain more what you have in mind. My understanding of what you said is to get rid of the createFromGlobals before the HttpKernelRunner. Wich is not an option for this feature.

Once again, this is my understanding of this mechanism. I am probably wrong, so feel free to adapt this MR or may be close it as you consider it useless and change internal process to get the routes.

docs/reference/advanced_configuration.rst
```diff
@@ -26,7 +26,7 @@ Full configuration options:
     sonata_page:
         skip_redirection: false # Skip asking Editor to redirect
         use_streamed_response: false
-        multisite: ~ # Required
+        multisite: ~
```
VincentLanglet: Can you revert this change now the multisite is required for the runtime ^^'

src/DependencyInjection/Configuration.php
```diff
@@ -85,7 +85,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->end()
                 ->scalarNode('multisite')
                     ->info('For more information, see https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/')
-                    ->isRequired()
+                    ->defaultValue('host')
```
VincentLanglet: Can you revert this change

VincentLanglet (dismissed):

src/Runtime/SonataPageRuntime.php
eerison: Hello Hello @GeraudBourdin  :)

Well I guess we are talking about the samething But with other code perspective, maybe I am also wrong :D, But please let me give a try, I am setup this locally and I will use your Runtime and see if my changes works if it works then I will open a PR for your branch just as a POC, I guess it will be easier to Visualize the whole idea.

> Once again, this is my understanding of this mechanism. I am probably wrong, so feel free to adapt this MR or may be close it as you consider it useless and change internal process to get the routes.

Nope I do not want close your PR, we just need to be patience and adapt the code for the best solution, sometimes it takes time, But it is really common in new features, maintainers have different point overview and it just find something in the middle :)

I am trying to setup this in one private project that I have, I just need to update some dependencies 😄 , I am doing this in the mean time of my current tasks.

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,47 @@
...
+class SonataPageRuntime extends SymfonyRuntime
+{
+    private string $multisite;
+
+    /**
+     * @param array<mixed> $options
```
VincentLanglet:
```suggestion
     * @param array{
     *     debug?: bool|null,
     *     env?: string|null,
     *     disable_dotenv?: bool|null,
     *     project_dir?: string|null,
     *     prod_envs?: string[]|null,
     *     dotenv_path?: string|null,
     *     test_envs?: string[]|null,
     *     use_putenv?: bool|null,
     *     runtimes?: array<mixed>|null,
     *     error_handler?: string|false,
     *     env_var_name?: string,
     *     debug_var_name?: string,
     *     dotenv_overload?: bool|null,
     *     multisite: string,
     *   } $options $options
```
should solve the psalm issue

VincentLanglet:
```
multisite: string,
```
need to be changed in
```
multisite?: string,
```

and Ci need to be fixed https://github.com/sonata-project/SonataPageBundle/actions/runs/7003266717/job/19048641209?pr=1727

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,62 @@
...
+     *    }  $options  $options
+     */
+    public function __construct(array $options = [])
+    {
+        if (!isset($options['multisite'])) {
+            throw new \InvalidArgumentException('The option multisite is required');
+        }
+
+        $this->multisite = $options['multisite'];
+        parent::__construct($options);
```
VincentLanglet, line 51:
```suggestion
        $this->multisite = $options['multisite'];
        unset($options['multisite']);
        parent::__construct($options);
```
will also help to solve the psalm issue.

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,62 @@
...
```
VincentLanglet:
```suggestion
        $this->multisite = $options['multisite'];
        unset($options['multisite']);
        parent::__construct($options);
```
will solve the psalm issue

src/Entity/Transformer.php
```diff
@@ -150,6 +150,7 @@ public function create(PageInterface $page, ?SnapshotInterface $snapshot = null)
             /**
              * @var PageContent $content
              */
+            // @phpstan-ignore-next-line
```
VincentLanglet: Please revert this, I'll merge your PR with the phpstan error.
We have an issue for this https://github.com/sonata-project/SonataPageBundle/issues/1728

src/Runtime/SonataPageRuntime.php
GeraudBourdin: see this message now. should be allready solved

VincentLanglet: No, you need the unset to solve
```
src/Runtime/SonataPageRuntime.php:51:29: InvalidArgument: Argument 1 of Symfony\Component\Runtime\SymfonyRuntime::__construct expects array{debug?: bool|null, debug_var_name?: string, disable_dotenv?: bool|null, dotenv_overload?: bool|null, dotenv_path?: null|string, env?: null|string, env_var_name?: string, error_handler?: false|string, prod_envs?: array<array-key, string>|null, project_dir?: null|string, runtimes?: array<array-key, mixed>|null, test_envs?: array<array-key, string>|null, use_putenv?: bool|null}, but array{debug?: bool|null, debug_var_name?: string, disable_dotenv?: bool|null, dotenv_overload?: bool|null, dotenv_path?: null|string, env?: null|string, env_var_name?: string, error_handler?: false|string, multisite: string, prod_envs?: array<array-key, string>|null, project_dir?: null|string, runtimes?: array<array-key, mixed>|null, test_envs?: array<array-key, string>|null, use_putenv?: bool|null} with additional array shape fields (multisite) was provided
```

the "parent" doesn't support the multisite option, and psalm ask the array to remove the key

GeraudBourdin: i thought i had already pushed this :s sorry about that.

src/Runtime/SonataPageRuntime.php
eerison: Hey @GeraudBourdin 

How did you test this runtime?

did you use your branch as alias in composer?

for example

```json
           "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/GeraudBourdin/SonataPageBundle"
        }
    ]
.....
        "sonata-project/page-bundle": "dev-1604-runtime as 4.0.0",
        
....

    "extra": {
        "symfony": {
            "allow-contrib": false,
            "require": "^6.2"
        },
            "runtime" : {
                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
                "multisite": "host_with_path_by_locale"
            }
    },


```

Because using your runtime and, it does not work with

localhost:8001/page1 ✅ 
localhost:8001/de/page1 ❌

eerison: did you change anything in your front controller 🤔 ?

eerison (changes requested):

VincentLanglet (dismissed):

VincentLanglet:
Thanks for your time and your work.

I'll wait one more day if someone has something to say and I'll merge it ^^

eerison:
> Thanks for your time and your work.
> 
> I'll wait one more day if someone has something to say and I'll merge it ^^

should be good confirm that, it is working ;)

src/Runtime/SonataPageRuntime.php
GeraudBourdin:
```
Regular symfony  index.php 
<?php

error_reporting(E_ALL & ~E_NOTICE & ~E_DEPRECATED);
use App\Kernel;

require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
    
    return function (array $context) {
        return new Kernel($context['APP_ENV'], (bool) $context['APP_DEBUG']);
    };
```



Composer : 
```
.....
    "extra": {
....
        "runtime": {
            "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime",
            "multisite": "host_with_path_by_locale"
        }
    },
    
    
    "repositories": {
    ......
    {
            "type": "path",
            "url": "../SonataPageBundle/",
            "options": {
                "symlink": true
            }
        }
    }
```

linked directly to my cloned project on the 1604-runtime branch. Need a dump autoload to get this loaded.

Hope this helps

eerison: > linked directly to my cloned project on the 1604-runtime branch. Need a dump autoload to get this loaded.

Yep I did it

Note: should be good also add this into the docs ;)

well I do not know why it is not working, do you have sites path, don't you? like `/en/page-a` and `/fr/page-a` or `/de/page-a`

and all of them worked with you runtime, right?

other thing could you take this from the url? then we can reproduce exactly the same code.
I mean: `https://github.com/GeraudBourdin/SonataPageBundle`

tests/Runtime/SonataPageRuntimeTest.php
```diff
@@ -0,0 +1,48 @@
...
+namespace Sonata\PageBundle\Tests\Runtime;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\PageBundle\Runtime\SonataPageRuntime;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Runtime\Runner\Symfony\ConsoleApplicationRunner;
+use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
+
+class SonataPageRuntimeTest extends TestCase
```
eerison: Thank you <3

src/Runtime/SonataPageRuntime.php
eerison: the runtime is called, Because when I add an invalid value for mutisite, I see that it is called.

But when I use like `host_with_path_by_locale`, it is not finding the route 🤔

eerison: Please run `composer info symfony/* sonata-project/*`

and let's see what we have different, post here your output!


just to save, it is my output

```
symfony/asset                      v6.3.8  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image...
symfony/cache                      v6.3.8  Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/cache-contracts            v3.4.0  Generic abstractions related to caching
symfony/clock                      v6.3.4  Decouples applications from the system clock
symfony/config                     v6.3.8  Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v6.3.8  Eases the creation of beautiful and testable command line interfaces
symfony/debug-bundle               v6.3.2  Provides a tight integration of the Symfony VarDumper component and the ServerLogCommand from MonologBr...
symfony/dependency-injection       v6.3.8  Allows you to standardize and centralize the way objects are constructed in your application
symfony/deprecation-contracts      v3.4.0  A generic function and convention to trigger deprecation notices
symfony/doctrine-bridge            v6.3.8  Provides integration for Doctrine with various Symfony components
symfony/doctrine-messenger         v6.3.8  Symfony Doctrine Messenger Bridge
symfony/dotenv                     v6.3.7  Registers environment variables from a .env file
symfony/error-handler              v6.3.5  Provides tools to manage errors and ease debugging PHP code
symfony/event-dispatcher           v6.3.2  Provides tools that allow your application components to communicate with each other by dispatching eve...
symfony/event-dispatcher-contracts v3.4.0  Generic abstractions related to dispatching event
symfony/expression-language        v6.3.0  Provides an engine that can compile and evaluate expressions
symfony/filesystem                 v6.3.1  Provides basic utilities for the filesystem
symfony/finder                     v6.3.5  Finds files and directories via an intuitive fluent interface
symfony/flex                       v2.4.1  Composer plugin for Symfony
symfony/form                       v6.3.8  Allows to easily create, process and reuse HTML forms
symfony/framework-bundle           v6.3.8  Provides a tight integration between Symfony components and the Symfony full-stack framework
symfony/http-client                v6.3.8  Provides powerful methods to fetch HTTP resources synchronously or asynchronously
symfony/http-client-contracts      v3.4.0  Generic abstractions related to HTTP clients
symfony/http-foundation            v6.3.8  Defines an object-oriented layer for the HTTP specification
symfony/http-kernel                v6.3.8  Provides a structured process for converting a Request into a Response
symfony/intl                       v6.3.7  Provides access to the localization data of the ICU library
symfony/mailer                     v6.3.5  Helps sending emails
symfony/maker-bundle               v1.51.1 Symfony Maker helps you create empty commands, controllers, form classes, tests and more so you can for...
symfony/messenger                  v6.3.7  Helps applications send and receive messages to/from other applications or via message queues
symfony/mime                       v6.3.5  Allows manipulating MIME messages
symfony/monolog-bridge             v6.3.8  Provides integration for Monolog with various Symfony components
symfony/monolog-bundle             v3.10.0 Symfony MonologBundle
symfony/notifier                   v6.3.6  Sends notifications via one or more channels (email, SMS, ...)
symfony/options-resolver           v6.3.0  Provides an improved replacement for the array_replace PHP function
symfony/password-hasher            v6.3.8  Provides password hashing utilities
symfony/polyfill-intl-grapheme     v1.28.0 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-icu          v1.28.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-intl-idn          v1.28.0 Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-intl-normalizer   v1.28.0 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring          v1.28.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php83             v1.28.0 Symfony polyfill backporting some PHP 8.3+ features to lower PHP versions
symfony/process                    v6.3.4  Executes commands in sub-processes
symfony/property-access            v6.3.2  Provides functions to read and write from/to an object or array using a simple string notation
symfony/property-info              v6.3.0  Extracts information about PHP class' properties using metadata of popular sources
symfony/proxy-manager-bridge       v6.3.0  Provides integration for ProxyManager with various Symfony components
symfony/routing                    v6.3.5  Maps an HTTP request to a set of configuration variables
symfony/runtime                    v6.3.2  Enables decoupling PHP applications from global state
symfony/security-acl               v3.3.2  Symfony Security Component - ACL (Access Control List)
symfony/security-bundle            v6.3.8  Provides a tight integration of the Security component into the Symfony full-stack framework
symfony/security-core              v6.3.7  Symfony Security Component - Core Library
symfony/security-csrf              v6.3.2  Symfony Security Component - CSRF Library
symfony/security-http              v6.3.8  Symfony Security Component - HTTP Integration
symfony/serializer                 v6.3.8  Handles serializing and deserializing data structures, including object graphs, into array structures o...
symfony/service-contracts          v3.4.0  Generic abstractions related to writing services
symfony/stopwatch                  v6.3.0  Provides a way to profile code
symfony/string                     v6.3.8  Provides an object-oriented API to strings and deals with bytes, UTF-8 code points and grapheme cluster...
symfony/translation                v6.3.7  Provides tools to internationalize your application
symfony/translation-contracts      v3.4.0  Generic abstractions related to translation
symfony/twig-bridge                v6.3.8  Provides integration for Twig with various Symfony components
symfony/twig-bundle                v6.3.8  Provides a tight integration of Twig into the Symfony full-stack framework
symfony/validator                  v6.3.8  Provides tools to validate values
symfony/var-dumper                 v6.3.8  Provides mechanisms for walking through any arbitrary PHP variable
symfony/var-exporter               v6.3.6  Allows exporting any serializable PHP data structure to plain PHP code
symfony/web-link                   v6.3.0  Manages links between resources
symfony/web-profiler-bundle        v6.3.8  Provides a development tool that gives detailed information about the execution of any request
symfony/webapp-meta                v1.0.0  A meta package for the web app pack
symfony/webpack-encore-bundle      v1.17.2 Integration with your Symfony app & Webpack Encore!
symfony/yaml                       v6.3.8  Loads and dumps YAML files

```

GeraudBourdin:
```
symfony/acl-bundle                 v2.3.0  Symfony AclBundle
symfony/apache-pack                v1.0.1  A pack for Apache support in Symfony
symfony/asset                      v6.3.8  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and i...
symfony/cache                      v6.3.8  Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/cache-contracts            v3.4.0  Generic abstractions related to caching
symfony/clock                      v6.3.4  Decouples applications from the system clock
symfony/config                     v6.3.8  Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v6.2.13 Eases the creation of beautiful and testable command line interfaces
symfony/debug-bundle               v6.3.2  Provides a tight integration of the Symfony VarDumper component and the ServerLogCommand from Monol...
symfony/dependency-injection       v6.3.8  Allows you to standardize and centralize the way objects are constructed in your application
symfony/deprecation-contracts      v3.4.0  A generic function and convention to trigger deprecation notices
symfony/doctrine-bridge            v6.3.8  Provides integration for Doctrine with various Symfony components
symfony/dotenv                     v6.2.8  Registers environment variables from a .env file
symfony/error-handler              v6.3.5  Provides tools to manage errors and ease debugging PHP code
symfony/event-dispatcher           v6.3.2  Provides tools that allow your application components to communicate with each other by dispatching...
symfony/event-dispatcher-contracts v3.4.0  Generic abstractions related to dispatching event
symfony/expression-language        v6.3.0  Provides an engine that can compile and evaluate expressions
symfony/filesystem                 v6.3.1  Provides basic utilities for the filesystem
symfony/finder                     v6.3.5  Finds files and directories via an intuitive fluent interface
symfony/flex                       v2.4.1  Composer plugin for Symfony
symfony/form                       v6.3.8  Allows to easily create, process and reuse HTML forms
symfony/framework-bundle           v6.3.8  Provides a tight integration between Symfony components and the Symfony full-stack framework
symfony/google-mailer              v6.3.0  Symfony Google Mailer Bridge
symfony/http-client                v6.3.8  Provides powerful methods to fetch HTTP resources synchronously or asynchronously
symfony/http-client-contracts      v3.4.0  Generic abstractions related to HTTP clients
symfony/http-foundation            v6.3.8  Defines an object-oriented layer for the HTTP specification
symfony/http-kernel                v6.3.8  Provides a structured process for converting a Request into a Response
symfony/intl                       v6.3.7  Provides access to the localization data of the ICU library
symfony/mailer                     v6.3.5  Helps sending emails
symfony/maker-bundle               v1.50.0 Symfony Maker helps you create empty commands, controllers, form classes, tests and more so you can...
symfony/mime                       v6.3.5  Allows manipulating MIME messages
symfony/monolog-bridge             v6.3.8  Provides integration for Monolog with various Symfony components
symfony/monolog-bundle             v3.10.0 Symfony MonologBundle
symfony/options-resolver           v6.3.0  Provides an improved replacement for the array_replace PHP function
symfony/password-hasher            v6.3.8  Provides password hashing utilities
symfony/polyfill-intl-grapheme     v1.28.0 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-icu          v1.28.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-intl-idn          v1.28.0 Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-intl-normalizer   v1.28.0 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring          v1.28.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php83             v1.28.0 Symfony polyfill backporting some PHP 8.3+ features to lower PHP versions
symfony/process                    v6.3.4  Executes commands in sub-processes
symfony/property-access            v6.3.2  Provides functions to read and write from/to an object or array using a simple string notation
symfony/property-info              v6.3.0  Extracts information about PHP class' properties using metadata of popular sources
symfony/psr-http-message-bridge    v2.3.1  PSR HTTP message bridge
symfony/routing                    v6.3.5  Maps an HTTP request to a set of configuration variables
symfony/runtime                    v6.3.2  Enables decoupling PHP applications from global state
symfony/security-acl               v3.3.2  Symfony Security Component - ACL (Access Control List)
symfony/security-bundle            v6.3.8  Provides a tight integration of the Security component into the Symfony full-stack framework
symfony/security-core              v6.3.7  Symfony Security Component - Core Library
symfony/security-csrf              v6.3.2  Symfony Security Component - CSRF Library
symfony/security-http              v6.3.8  Symfony Security Component - HTTP Integration
symfony/serializer                 v6.3.8  Handles serializing and deserializing data structures, including object graphs, into array structur...
symfony/service-contracts          v3.4.0  Generic abstractions related to writing services
symfony/stopwatch                  v6.3.0  Provides a way to profile code
symfony/string                     v6.3.8  Provides an object-oriented API to strings and deals with bytes, UTF-8 code points and grapheme clu...
symfony/templating                 v6.3.0  Provides all the tools needed to build any kind of template system
symfony/translation                v6.3.7  Provides tools to internationalize your application
symfony/translation-contracts      v3.4.0  Generic abstractions related to translation
symfony/twig-bridge                v6.3.8  Provides integration for Twig with various Symfony components
symfony/twig-bundle                v6.3.8  Provides a tight integration of Twig into the Symfony full-stack framework
symfony/validator                  v6.3.8  Provides tools to validate values
symfony/var-dumper                 v6.3.8  Provides mechanisms for walking through any arbitrary PHP variable
symfony/var-exporter               v6.3.6  Allows exporting any serializable PHP data structure to plain PHP code
symfony/web-profiler-bundle        v6.3.8  Provides a development tool that gives detailed information about the execution of any request
symfony/webpack-encore-bundle      v1.17.2 Integration with your Symfony app & Webpack Encore!
symfony/yaml                       v6.3.8  Loads and dumps YAML files
www-data@3f5846d80f08:~$ composer info symfony/*                 
symfony/acl-bundle                 v2.3.0  Symfony AclBundle
symfony/apache-pack                v1.0.1  A pack for Apache support in Symfony
symfony/asset                      v6.3.8  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and i...
symfony/cache                      v6.3.8  Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/cache-contracts            v3.4.0  Generic abstractions related to caching
symfony/clock                      v6.3.4  Decouples applications from the system clock
symfony/config                     v6.3.8  Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v6.2.13 Eases the creation of beautiful and testable command line interfaces
symfony/debug-bundle               v6.3.2  Provides a tight integration of the Symfony VarDumper component and the ServerLogCommand from Monol...
symfony/dependency-injection       v6.3.8  Allows you to standardize and centralize the way objects are constructed in your application
symfony/deprecation-contracts      v3.4.0  A generic function and convention to trigger deprecation notices
symfony/doctrine-bridge            v6.3.8  Provides integration for Doctrine with various Symfony components
symfony/dotenv                     v6.2.8  Registers environment variables from a .env file
symfony/error-handler              v6.3.5  Provides tools to manage errors and ease debugging PHP code
symfony/event-dispatcher           v6.3.2  Provides tools that allow your application components to communicate with each other by dispatching...
symfony/event-dispatcher-contracts v3.4.0  Generic abstractions related to dispatching event
symfony/expression-language        v6.3.0  Provides an engine that can compile and evaluate expressions
symfony/filesystem                 v6.3.1  Provides basic utilities for the filesystem
symfony/finder                     v6.3.5  Finds files and directories via an intuitive fluent interface
symfony/flex                       v2.4.1  Composer plugin for Symfony
symfony/form                       v6.3.8  Allows to easily create, process and reuse HTML forms
symfony/framework-bundle           v6.3.8  Provides a tight integration between Symfony components and the Symfony full-stack framework
symfony/google-mailer              v6.3.0  Symfony Google Mailer Bridge
symfony/http-client                v6.3.8  Provides powerful methods to fetch HTTP resources synchronously or asynchronously
symfony/http-client-contracts      v3.4.0  Generic abstractions related to HTTP clients
symfony/http-foundation            v6.3.8  Defines an object-oriented layer for the HTTP specification
symfony/http-kernel                v6.3.8  Provides a structured process for converting a Request into a Response
symfony/intl                       v6.3.7  Provides access to the localization data of the ICU library
symfony/mailer                     v6.3.5  Helps sending emails
symfony/maker-bundle               v1.50.0 Symfony Maker helps you create empty commands, controllers, form classes, tests and more so you can...
symfony/mime                       v6.3.5  Allows manipulating MIME messages
symfony/monolog-bridge             v6.3.8  Provides integration for Monolog with various Symfony components
symfony/monolog-bundle             v3.10.0 Symfony MonologBundle
symfony/options-resolver           v6.3.0  Provides an improved replacement for the array_replace PHP function
symfony/password-hasher            v6.3.8  Provides password hashing utilities
symfony/polyfill-intl-grapheme     v1.28.0 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-icu          v1.28.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-intl-idn          v1.28.0 Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-intl-normalizer   v1.28.0 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring          v1.28.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php83             v1.28.0 Symfony polyfill backporting some PHP 8.3+ features to lower PHP versions
symfony/process                    v6.3.4  Executes commands in sub-processes
symfony/property-access            v6.3.2  Provides functions to read and write from/to an object or array using a simple string notation
symfony/property-info              v6.3.0  Extracts information about PHP class' properties using metadata of popular sources
symfony/psr-http-message-bridge    v2.3.1  PSR HTTP message bridge
symfony/routing                    v6.3.5  Maps an HTTP request to a set of configuration variables
symfony/runtime                    v6.3.2  Enables decoupling PHP applications from global state
symfony/security-acl               v3.3.2  Symfony Security Component - ACL (Access Control List)
symfony/security-bundle            v6.3.8  Provides a tight integration of the Security component into the Symfony full-stack framework
symfony/security-core              v6.3.7  Symfony Security Component - Core Library
symfony/security-csrf              v6.3.2  Symfony Security Component - CSRF Library
symfony/security-http              v6.3.8  Symfony Security Component - HTTP Integration
symfony/serializer                 v6.3.8  Handles serializing and deserializing data structures, including object graphs, into array structur...
symfony/service-contracts          v3.4.0  Generic abstractions related to writing services
symfony/stopwatch                  v6.3.0  Provides a way to profile code
symfony/string                     v6.3.8  Provides an object-oriented API to strings and deals with bytes, UTF-8 code points and grapheme clu...
symfony/templating                 v6.3.0  Provides all the tools needed to build any kind of template system
symfony/translation                v6.3.7  Provides tools to internationalize your application
symfony/translation-contracts      v3.4.0  Generic abstractions related to translation
symfony/twig-bridge                v6.3.8  Provides integration for Twig with various Symfony components
symfony/twig-bundle                v6.3.8  Provides a tight integration of Twig into the Symfony full-stack framework
symfony/validator                  v6.3.8  Provides tools to validate values
symfony/var-dumper                 v6.3.8  Provides mechanisms for walking through any arbitrary PHP variable
symfony/var-exporter               v6.3.6  Allows exporting any serializable PHP data structure to plain PHP code
symfony/web-profiler-bundle        v6.3.8  Provides a development tool that gives detailed information about the execution of any request
symfony/webpack-encore-bundle      v1.17.2 Integration with your Symfony app & Webpack Encore!
symfony/yaml                       v6.3.8  Loads and dumps YAML files



sonata-project/admin-bundle              4.28.1  The missing Symfony Admin Generator
sonata-project/block-bundle              5.1.0   Symfony SonataBlockBundle
sonata-project/classification-bundle     4.7.1   Symfony SonataClassificationBundle
sonata-project/doctrine-extensions       2.3.0   Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 4.14.1  Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  3.3.0   Lightweight Exporter library
sonata-project/form-extensions           2.3.0   Symfony form extensions
sonata-project/formatter-bundle          5.3.0   Symfony SonataFormatterBundle
sonata-project/intl-bundle               3.1.0   Symfony SonataIntlBundle
sonata-project/media-bundle              4.10.5  Symfony SonataMediaBundle
sonata-project/page-bundle               5.x-dev This bundle provides a Site and Page management through container and block services
sonata-project/seo-bundle                3.7.0   Symfony SonataSeoBundle
sonata-project/translation-bundle        3.3.0   SonataTranslationBundle
sonata-project/twig-extensions           2.4.0   Sonata twig extensions
sonata-project/user-bundle               5.10.0  Symfony SonataUserBundle
```

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,63 @@
...
+        if (!isset($options['multisite'])) {
+            throw new \InvalidArgumentException('The option multisite is required');
+        }
+
+        $this->multisite = $options['multisite'];
+        unset($options['multisite']);
+        parent::__construct($options);
+    }
+
+    public function getRunner(?object $application): RunnerInterface
```
eerison: Ok as I supected run time works with this code ;)

```php
<?php

declare(strict_types=1);

/*
 * This file is part of the Sonata Project package.
 *
 * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Sonata\PageBundle\Runtime;

use Sonata\PageBundle\Request\SiteRequest;
use Symfony\Component\HttpKernel\HttpKernelInterface;
use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
use Symfony\Component\Runtime\RunnerInterface;
use Symfony\Component\Runtime\SymfonyRuntime;

class SonataPageRuntime extends SymfonyRuntime
{
    public function getRunner(?object $application): RunnerInterface
    {
        if (!$application instanceof HttpKernelInterface) {
            return parent::getRunner($application);
        }

        return new HttpKernelRunner($application, SiteRequest::createFromGlobals());

    }
}
```

and you don't need to pass `multisite` on composer.json

eerison: and I just search and `RequestFactory` is not used anywhere, then we can deprecate the whole class!

eerison: well if you want you can keep the logical like this


```php
        if ($application instanceof HttpKernelInterface) {
            return new HttpKernelRunner($application, SiteRequest::createFromGlobals());
        }
        
        return parent::getRunner($application);
```

it is up to you ;)

Note: As I said before I would rename this for `SonataPagePathRuntime` since it is used for path route only.

GeraudBourdin: `and you don't need to pass multisite on composer.json`

You are right, you could get rid of this, but as the Sonata\PageBundle\Request\RequestFactory line 60 it prevents the creation of the Request::setFactory twice as the first one is already done. So in my point of view, no, you should keep it. 

`well if you want you can keep the local like this`
What do you mean ?

GeraudBourdin: Do you have a branch i could use to try if i's ok in my set up ?

eerison: > You are right, you could get rid of this, but as the Sonata\PageBundle\Request\RequestFactory line 60 it prevents the creation of the Request::setFactory twice as the first one is already done. So in my point of view, no, you should keep it.

It was there because it was used in front controller and we have only one, as it was there we need to to check what was passing, because it could be `host` or `host_path`, but as we are using runtime, it will use host_path... only.

then this if gonna be always true! it is not needed.

> What do you mean ?

Sorry it was a typo, I edited the comment.

> Do you have a branch i could use to try if i's ok in my set up ?

nope, I just used your branch and update directly on vendor folder, if you copy and past the code above into your runtime, you gonna see the same behaviour then me ;)

eerison: > Do you have a branch i could use to try if i's ok in my set up ?

or do you think, it's necessary I create a branch with this?

GeraudBourdin: That's right, you can go directly to `SiteRequest::createFromGlobals()`  it could simplify the all process. kudo!

As `RequestFactory::create` and   `RequestFactory::createFromGlobals` seams useless in the case that only 
`host_with_path` ad `host_with_path_by_locale` should be used with the Runtime. 

But, how to prevent user to use multisite:host params with the runtime ? Would it be any impact ?

eerison: > But, how to prevent user to use multisite:host params with the runtime ? Would it be any impact ?

it is used in others class via param, I saw that it was used as a event listener, and we are also duplicating some logical there, I believe we can also simplify this, But I didn't start this topic to not complicate this PR.

But I guess it will be easier to handle it later since it is a internal logical.

btw, do not forgot to add final class for your runtime.

eerison: > But, how to prevent user to use multisite:host params with the runtime

Note1: we still need to keep this param because it is used in some event listeners/handlers, But for runtime, it is not important ;)

eerison: @GeraudBourdin when you update let me know then I can review again 😃

GeraudBourdin: Sorry for the delay, i had work to do. 
So PR updated with the new SiteRequest call, updated tests and doc. 
Let me know what' could be improved.

docs/reference/multisite.rst
```diff
@@ -100,38 +80,16 @@ The last action is to configure the ``sonata_page`` section as:
...
-    if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? $_ENV['TRUSTED_HOSTS'] ?? false) {
-        Request::setTrustedHosts([$trustedHosts]);
-    }
+        "extra": {
+            "runtime": {
+                "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
+            }
 
-    $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
-    $request = RequestFactory::createFromGlobals('host_with_path_by_locale'); // before: $request = Request::createFromGlobals();
```
VincentLanglet, line 131: Does `RequestFactory::createFromGlobals` still useful or can it be deprecated @eerison ?

eerison: The whole RequestFactory class can be deprecated.

src/Runtime/SonataPageRuntime.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Sonata\PageBundle\Runtime;
+
+use Sonata\PageBundle\Request\SiteRequest;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner;
+use Symfony\Component\Runtime\RunnerInterface;
+use Symfony\Component\Runtime\SymfonyRuntime;
+
+final class SonataPageRuntime extends SymfonyRuntime
```
eerison: Hey @GeraudBourdin 

as it will be used just for path routes, don't you think could be better we be most specific with this class' name?

because just `SonataPageRuntime`, it is like used for all use cases, But as it gonna be used just for path, IMO we could use something like `SonataPagePathRuntime`.

what do you think?

docs/reference/multisite.rst
eerison: @GeraudBourdin 

Please do not forget to deprecate `RequestFactory` and all methods inside of there.
You can search for examples into the code ;)

docs/reference/multisite.rst
```diff
@@ -100,38 +80,16 @@ The last action is to configure the ``sonata_page`` section as:
...
-    use Sonata\PageBundle\Request\RequestFactory; # before: use Symfony\Component\HttpFoundation\Request;
-
-    require dirname(__DIR__).'/config/bootstrap.php';
-
-    if ($_SERVER['APP_DEBUG']) {
-        umask(0000);
-
-        Debug::enable();
-    }
+This strategy requires a dedicated runtime. So you need to modify your
```
eerison, line 83: ~I guess we could simplify this doc.~

~In case the user wants to use `Host path` or `Host path by locale`, then add the runtime and change the mutisite config.~

Let's make this simple, Just add runtime example bellow of multisite config.
and also add a note saying that is important run `composer dump-autoload`.

docs/reference/multisite.rst
```diff
@@ -100,38 +80,16 @@ The last action is to configure the ``sonata_page`` section as:
...
+composer.json and declare ``SonataPageRuntime`` runtime.
```
eerison: on the line 78

```yaml
 sonata_page:
        multisite: host_with_path
```

please do not remove this form the doc, it is really important to make this work.
It was the issue that I have, if I just use runtime it won't work. there are part of the code doing some logical with this parameter.

I guess it can be avoided in the future, but for  the current solution this config is needed ;)

eerison (changes requested):

docs/reference/multisite.rst
GeraudBourdin: i never intended to do this as we need this to setup all services.

src/Runtime/SonataPageRuntime.php
GeraudBourdin: Done. You can review changes. Please pay specific attention to the deprecation parts :), thanks. Hope we will end one day with this pr ^^

docs/reference/multisite.rst
```diff
@@ -90,6 +90,7 @@ composer.json and declare ``SonataPageRuntime`` runtime.
                 "class": "Sonata\\PageBundle\\Runtime\\SonataPageRuntime"
```
VincentLanglet: SonataPagePathRuntime

docs/reference/multisite.rst
```diff
@@ -66,7 +66,7 @@ composer.json and declare ``SonataPageRuntime`` runtime.
...
```
VincentLanglet: SonataPagePathRuntime

src/Request/RequestFactory.php
```diff
@@ -78,10 +110,19 @@ private static function configureFactory(string $type): void
     }
 
     /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
      * @return class-string<Request>
      */
     private static function getClass(string $type): string
     {
+        @trigger_error(sprintf(
```
VincentLanglet: The method is private so you don't need to deprecate it, it will naturally be remove.

src/Request/RequestFactory.php
```diff
@@ -78,10 +110,19 @@ private static function configureFactory(string $type): void
     }
 
     /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
```
VincentLanglet:
```suggestion
```

src/Request/RequestFactory.php
```diff
@@ -43,20 +45,50 @@ public static function create(
...
 
         return \call_user_func_array([self::getClass($type), 'createFromGlobals'], []);
     }
 
+    /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
+     */
     private static function configureFactory(string $type): void
     {
+        @trigger_error(sprintf(
```
VincentLanglet: The method is private so you dont need to deprecate this one.

src/Request/RequestFactory.php
```diff
@@ -27,6 +28,7 @@ final class RequestFactory
     ];
 
     /**
+     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
```
VincentLanglet:
```suggestion
     * @deprecated since sonata-project/page-bundle 4.6.2, to be removed in 5.0.
     *
```

src/Request/RequestFactory.php
```diff
@@ -13,6 +13,7 @@
 
 namespace Sonata\PageBundle\Request;
 
+use Sonata\PageBundle\Runtime\SonataPagePathRuntime;
 use Symfony\Component\HttpFoundation\Request;
 
 final class RequestFactory
```
VincentLanglet, line 24:
```suggestion
/**
 * NEXT_MAJOR: Remove this class
 *
 * @deprecated since ...
 */
final class RequestFactory
```

VincentLanglet (commented):
For the deprecation in the tests, just add
```
* @group legacy
*
* NEXT_MAJOR: Remove this class
```
at the top of the RequestFactoryTest class

src/Request/RequestFactory.php
eerison: hmmm true 👀

eerison:
> For the deprecation in the tests, just add
> 
> ```
> * @group legacy
> *
> * NEXT_MAJOR: Remove this class
> ```
> 
> at the top of the RequestFactoryTest class

@GeraudBourdin ☝🏼

GeraudBourdin:
ok i missed one line between annotation and NEXT_major.

VincentLanglet:
> ok i missed one line between annotation and NEXT_major.

```
/**
 * @group legacy
 *
 * NEXT_MAJOR: Remove this class
```
was for the RequestFactoryTest and not the runtime one ;)

eerison:
> ok i missed one line between annotation and NEXT_major.

you are also missing this comment 😄 

https://github.com/sonata-project/SonataPageBundle/pull/1727#pullrequestreview-1754787850

eerison:
> > ok i missed one line between annotation and NEXT_major.
> 
> ```
> /**
>  * @group legacy
>  *
>  * NEXT_MAJOR: Remove this class
> ```
> 
> was for the RequestFactoryTest and not the runtime one ;)

I guess we need to do something else to psalm didn't complain about deprecated method.
Do you have any idea what should we do in this case @VincentLanglet ?

eerison (approved):

eerison:
Looks good for me 💪

Thank you @GeraudBourdin ,it is a really nice feature ♥️

Good job 👏👏

tdumalin:
Hi everyone,

How can I use the `multisite: host_with_path` before the next release ?
For now if i tried to use this option I got the following error:

```console
You must change the main Request object in the front controller (app.php) in order to use the `host_with_path` strategy
```


I guess there is some workaround to do in the `public/index.php` ? 


Thanks for the help

VincentLanglet:
I just made a new release

tdumalin:
@VincentLanglet, perfect thank you very much !
Let's use this new feature :+1:

haivala:
I installed this but now I cannot access web_profiler in dev anymore. Anyone experiencing the same? going to _profiler/{hash} url is giving me `An exception has been thrown during the rendering of a template ("You must change the main Request object in the front controller (app.php) in order to use the `host_with_path_by_locale` strategy").`

eerison:
> I installed this but now I cannot access web_profiler in dev anymore. Anyone experiencing the same? going to _profiler/{hash} url is giving me `An exception has been thrown during the rendering of a template ("You must change the main Request object in the front controller (app.php) in order to use the `host_with_path_by_locale` strategy").`

did you add the runtime?

eerison:
@haivala 

did you do this: https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/#host-and-path-strategy ?

eerison:
> @haivala
> 
> did you do this: https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/#host-and-path-strategy ?

in theory you didn't add the runtime: https://github.com/sonata-project/SonataPageBundle/blob/b009becd106bab26d572bf81141b7d398d3fa81b/src/Site/HostPathByLocaleSiteSelector.php#L29 and run `composer dump-autoload`.

Note: I guess we could update this message as well :D

VincentLanglet:
> > @haivala
> > did you do this: [docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/#host-and-path-strategy](https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/#host-and-path-strategy) ?
> 
> in theory you didn't add the runtime:
> 
> https://github.com/sonata-project/SonataPageBundle/blob/b009becd106bab26d572bf81141b7d398d3fa81b/src/Site/HostPathByLocaleSiteSelector.php#L29
> 
> and run `composer dump-autoload`.
> Note: I guess we could update this message as well :D

PR welcomed for the message if it's the reason of the issue ;)

eerison:
> > > @haivala
> > > did you do this: [docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/#host-and-path-strategy](https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/multisite/#host-and-path-strategy) ?
> > 
> > 
> > in theory you didn't add the runtime:
> > https://github.com/sonata-project/SonataPageBundle/blob/b009becd106bab26d572bf81141b7d398d3fa81b/src/Site/HostPathByLocaleSiteSelector.php#L29
> > 
> > and run `composer dump-autoload`.
> > Note: I guess we could update this message as well :D
> 
> PR welcomed for the message if it's the reason of the issue ;)

https://github.com/sonata-project/SonataPageBundle/pull/1737

haivala:
Yes. I did. Runtime was added automatically with the pagebundle update. Everything else is working as before. Just cannot access the web profiler anymore.

haivala:
Feels like `ignore_route_patterns` config option is not working anymore.

eerison:
> Feels like ignore_route_patterns config option is not working anymore.

Please open an issue with

- your composer.json extra config (and what appear when you run composer dump-autoload)
- index.php file
- sonata_page.yaml config

Issue #1728, opened 2023-11-25
Fix pipeline TransformerInterface


eerison:
Hello Hello

I was trying to fix the pipelines But I saw that it is complain about `id`,`created_at` and `updated_at`.

Here it is defined that those fields are not optional: https://github.com/sonata-project/SonataPageBundle/blame/4.x/src/Model/TransformerInterface.php#L48

But in this line it is removing in case those fields are null
https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/Transformer.php#L153

Just to be sure those fields must exist even if it's null , right @Hanmac ?

Hanmac:
Ugh I need to test again how the datetime normalizer handles the code

I will check this out soon


For me, I think the property can be marked as missing `created_at?`

eerison:
I guess those fields are need, there are others places expecting that it exists.

Hanmac:
> I guess those fields are need, there are others places expecting that it exists.

I need to test it if that's really the case

Can't do it right now

I would try to make an PR with the changes in the Interface, and then try to write test cases for what happens if the value are null

Or have you found places where it explicit looks for the values?

eerison:
> https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/Transformer.php#L153

I guess it is in this line: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/Transformer.php#L153

VincentLanglet:
Hi. I already take a look.

This can be done: https://github.com/sonata-project/SonataPageBundle/pull/1729/commits/6854a19299bf05712b0737d9e9324a417ed39169

But we still get an error with phpstan because of 
https://phpstan.org/r/7c91cd9d-2db3-4fb1-bdfb-a35685bd8921

On the other side, removing the var annotation solve the issue because phpstan is good enought to compute the type, but psalm doesn't.

Static analysis Ci can be solved by removing the filter of null values https://github.com/sonata-project/SonataPageBundle/pull/1729/commits/e97d3f1ac6c18f810091b7046855f2d821d22b78 thougth. I'm not sure why we filter them. (But test need to be fixed)

Anyway I will end ignoring psalm issue I think, but I opened https://github.com/phpstan/phpstan/discussions/10185

eerison:
Hey @VincentLanglet  I guess there is an issue there,

I created a test for this and the current transformers are not handling `id`, `created_at` and `updated_at`

[...]

PR #1730, opened 2023-11-25
Bug/1728 Required filds for transformer

eerison:
Issue #1728


eerison:
I added the reason into the ticket ;)

PR #1809, opened 2025-05-27
Add symfony recipes in installation doc

eerison:
## Clean up installation doc

I am targeting this branch, because it does not have any break change and can be introduce on version 4.x.

## Changelog

```markdown
### Changed
- Add symfony recipes in installation doc.
```

I was playing with sonata pagebundle in a new symfony project and I was strugulling to make this work 😄 ,
after a while I realised the issue was, because new symfony projects are using php attributes as default, for this reason I am changing the recipes: https://github.com/symfony/recipes-contrib/pull/1809

and running `composer require sonata-project/page-bundle -W` it works as expected.

here there is some screenshots

Screenshot from 2025-05-27 20-18-32

Screenshot from 2025-05-27 20-24-40

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -5,8 +5,22 @@
 Installation
 ============
 
-Prerequisites
+Automatically
 -------------
+
+Using Symfony recipes, you can easily set up Sonata Page.
+
+Add the ``SonataPageBundle`` via Composer::
+
+    composer require sonata-project/page-bundle
+
+The command will ask you to execute some Symfony recipes. Just type ``y`` for all of them.
+
+Manually
+-----------------
+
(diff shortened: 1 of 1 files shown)


VincentLanglet (changes requested):
We cannot assume that people are using flex (and recipes).
That's why the doc is descriptive.

So you shouldn't remove part that are automatically added by flex/recipe ; and keep consistency with other packages doc

eerison:
screenshot-2025-10-22_18-15-43

Hey @VincentLanglet 

Could you check if it looks better?

VincentLanglet (approved):

VincentLanglet:
thanks

PR #1817, opened 2025-09-17
fix(manager): fix implementation bugs with CmsManager classes

rande:
Co-authored-by: Vincent Langlet <vincentlanglet@hotmail.fr>

## fix implementation bugs with CmsManager classes

This PR fixes 2 issues:
 - the CmsManagerSelector constructor does not use interface for the Cms(Page|Snapshot)Manager, this make impossible to inject another implementation and defeat the purpose on IoC.
- the cache keys for the page in the `getPageBy` does not use the provide $site object. So if the code load 2 similar page (same fieldName and same value). then the former is always returned. 

I am targeting this branch, because we need to fix the bug on the active branch.

## Changelog

```markdown
### Changed
- Update `CmsManagerSelector::__construct()` to do use `CmsManagerInterface` and not the final class. 

### Fixed
- Update `CmsSnapshotManager::getPageBy()` and `CmsPageManager::getPageBy` to properly used the `$site` variable within the cache key.
```

### DIFF
--- a/src/CmsManager/CmsManagerSelector.php
+++ b/src/CmsManager/CmsManagerSelector.php
@@ -37,8 +37,8 @@ final class CmsManagerSelector implements CmsManagerSelectorInterface
      * @param AdminInterface<PageInterface> $pageAdmin
      */
     public function __construct(
-        private CmsPageManager $cmsPageManager,
-        private CmsSnapshotManager $cmsSnapshotManager,
+        private CmsManagerInterface $cmsPageManager,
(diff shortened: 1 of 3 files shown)


qsomazzi (approved):

rande:
@VincentLanglet hello, any idea on the composer error ?

VincentLanglet:
No idea maybe re running the ci might work.

Could you add a test ?

VincentLanglet (approved):

PR #1821, opened 2025-10-05
Remove usage of request::get

VincentLanglet:
## Subject

## Changelog

```markdown
### Fixed
- Stop using internal/deprecated method `Request::get()`
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -22,6 +22,3 @@ parameters:
             message: "#^Parameter \\#2 \\.\\.\\.\\$arrays of function array_merge expects array, array\\<string, bool\\|string\\>\\|Sonata\\\\PageBundle\\\\Model\\\\PageInterface\\|Sonata\\\\PageBundle\\\\Model\\\\SiteInterface\\|null given\\.$#"
             count: 1
             path: src/Form/Type/PageSelectorType.php
-        -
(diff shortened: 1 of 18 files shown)


dmaicher (approved):

eerison:
Hey @VincentLanglet 

Did you think in any other solution besides this BCHelper? 

And why it was needed? 

Edit: I know it's because of deprecated method,  but isn't it possible just replace for something else?  Or Symfony didn't provide any easy replacement?

VincentLanglet:
> Hey @VincentLanglet

Hi, @eerison.

I'm sorry but I'm getting pinged (by you) lot of time in the same week.
I need to remember you that:
- I'm not suppose to be the only one maintainer
- I'm maintaining around 10 sonata project FOR FREE even if I don't use them AT ALL.
- I have already lot of work on other subject (like my mental health)

> Did you think in any other solution besides this BCHelper?
> 
> And why it was needed?
> 
> Edit: I know it's because of deprecated method, but isn't it possible just replace for something else? Or Symfony didn't provide any easy replacement?

The right question is:
- Does the BCHelper is a quick solution => Yes
- Does the BCHelper is an issue for someone => No
- Does Symfony provides an easy replacement ? => You can look by yourself and you'll that no, except calling manually on query/attributes/request ; which would be a BC break.

eerison:
I'm sorry to ping you those amount of time.

As I got some free time, then I was trying to support, but it leads effort of some one to review and merge.

But take your time. and in case you need help to review/merge let me know. I will try to give some support.

PR #1827, opened 2025-10-24
Feature/1713 - Replace slugify to symfony string

### LINKED ISSUE #1713: Switch cocur/slugify to symfony/string

eerison:
## Replace slugify to symfony string

Closes #1713.
Related PR: #1718.

## Changelog

```markdown
### Added
- `FixPageUrlService` contains the logic from `$pageManager->fixUrl($page)`
- `FixPageUrlInterface` was added.
 
### Deprecated
- `SlugifyInterface` has been deprecated in favour of the`SluggerInterface` from `symfony/string`
```

### DIFF
--- a/src/Entity/PageManager.php
+++ b/src/Entity/PageManager.php
@@ -19,8 +19,8 @@ use Sonata\Doctrine\Entity\BaseEntityManager;
 use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Model\PageManagerInterface;
 use Sonata\PageBundle\Model\SiteInterface;
-use Sonata\PageBundle\Service\Contract\FixPageUrlInterface;
-use Sonata\PageBundle\Service\FixPageUrlService;
+use Sonata\PageBundle\Service\Contract\PageFixerInterface;
+use Sonata\PageBundle\Service\PageFixerService;
 use Symfony\Component\String\Slugger\AsciiSlugger;
 
 /**
@@ -38,17 +38,17 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface|FixPageUrlInterface $fixPageUrl,
+        private SlugifyInterface|PageFixerInterface $pageFixer,
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
         // NEXT_MAJOR: Remove the if block bellow and "cocur/slugify" dependecy.
-        if ($this->fixPageUrl instanceof SlugifyInterface) {
+        if ($this->pageFixer instanceof SlugifyInterface) {
             @trigger_error(\sprintf(
                 'Inject %s in %s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
                 SlugifyInterface::class,
                 self::class,
-                FixPageUrlInterface::class,
+                PageFixerInterface::class,
             ), \E_USER_DEPRECATED);
         }
         parent::__construct($class, $registry);
@@ -84,23 +84,23 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     }
 
     /**
-     * NEXT_MAJOR: keep only $this->fixPageUrl->fix($page) in this method.
+     * NEXT_MAJOR: keep only $this->fixPageUrl->fixUrl$page) in this method.
      */
     public function fixUrl(PageInterface $page): void
     {
-        $fixPageUrl = $this->fixPageUrl;
+        $pageFixer = $this->pageFixer;
 
-        if ($fixPageUrl instanceof SlugifyInterface) {
+        if ($pageFixer instanceof SlugifyInterface) {
             @trigger_error(\sprintf(
                 'Inject %s in %s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
                 SlugifyInterface::class,
                 self::class,
-                FixPageUrlInterface::class,
+                PageFixerInterface::class,
             ), \E_USER_DEPRECATED);
-            $fixPageUrl = new FixPageUrlService(new AsciiSlugger());
+            $pageFixer = new PageFixerService(new AsciiSlugger());
         }
 
-        $fixPageUrl->fix($page);
+        $pageFixer->fixUrl($page);
     }
 
     /**
--- a/src/Entity/PageManager.php
+++ b/src/Entity/PageManager.php
@@ -19,6 +19,9 @@ use Sonata\Doctrine\Entity\BaseEntityManager;
 use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Model\PageManagerInterface;
 use Sonata\PageBundle\Model\SiteInterface;
+use Sonata\PageBundle\Service\Contract\FixPageUrlInterface;
+use Sonata\PageBundle\Service\FixPageUrlService;
+use Symfony\Component\String\Slugger\AsciiSlugger;
 
 /**
  * @extends BaseEntityManager<PageInterface>
@@ -35,10 +38,19 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|FixPageUrlInterface $fixPageUrl,
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
+        // NEXT_MAJOR: Remove the if block bellow and "cocur/slugify" dependecy.
+        if ($this->fixPageUrl instanceof SlugifyInterface) {
+            @trigger_error(\sprintf(
+                'Inject %s in %s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
+                SlugifyInterface::class,
+                self::class,
+                FixPageUrlInterface::class,
+            ), \E_USER_DEPRECATED);
+        }
         parent::__construct($class, $registry);
     }
 
@@ -71,50 +83,24 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
         return $page;
     }
 
+    /**
+     * NEXT_MAJOR: keep only $this->fixPageUrl->fix($page) in this method.
+     */
     public function fixUrl(PageInterface $page): void
     {
-        if ($page->isInternal()) {
-            $page->setUrl(null); // internal routes do not have any url ...
+        $fixPageUrl = $this->fixPageUrl;
 
-            return;
+        if ($fixPageUrl instanceof SlugifyInterface) {
+            @trigger_error(\sprintf(
+                'Inject %s in %s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
+                SlugifyInterface::class,
+                self::class,
+                FixPageUrlInterface::class,
+            ), \E_USER_DEPRECATED);
+            $fixPageUrl = new FixPageUrlService(new AsciiSlugger());
         }
 
-        // hybrid page cannot be altered
-        if (!$page->isHybrid()) {
-            $parent = $page->getParent();
-
-            if (null !== $parent) {
-                $slug = $page->getSlug();
-
-                if (null === $slug) {
-                    $slug = $this->slugify->slugify($page->getName() ?? '');
-
-                    $page->setSlug($slug);
-                }
-
-                $parentUrl = $parent->getUrl();
-
-                if ('/' === $parentUrl) {
-                    $base = '/';
-                } elseif (!str_ends_with($parentUrl ?? '', '/')) {
-                    $base = $parentUrl.'/';
-                } else {
-                    $base = $parentUrl;
-                }
-
-                $url = $page->getCustomUrl() ?? $slug;
-                $page->setUrl('/'.ltrim($base.$url, '/'));
-            } else {
-                $page->setSlug(null);
-
-                $url = $page->getCustomUrl() ?? '';
-                $page->setUrl('/'.ltrim($url, '/'));
-            }
-        }
-
-        foreach ($page->getChildren() as $child) {
-            $this->fixUrl($child);
-        }
+        $fixPageUrl->fix($page);
     }
 
     /**
--- a/src/Service/FixPageUrlService.php
+++ b/src/Service/FixPageUrlService.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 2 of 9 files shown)


src/Service/PageFixerService.php
```diff
@@ -0,0 +1,73 @@
...
+            } else {
+                $page->setSlug(null);
+
+                $url = $page->getCustomUrl() ?? '';
+                $page->setUrl('/'.ltrim($url, '/'));
+            }
+        }
+
+        foreach ($page->getChildren() as $child) {
+            $this->fix($child);
```
eerison: renamed from `fixUrl` to `fix`

src/Service/PageFixerService.php
```diff
@@ -0,0 +1,73 @@
...
+
+        // hybrid page cannot be altered
+        if (!$page->isHybrid()) {
+            $parent = $page->getParent();
+
+            if (null !== $parent) {
+                $slug = $page->getSlug();
+
+                if (null === $slug) {
+                    $slug = $this->slugger
```
eerison, line 42: adapted the code inside of this `if block code`

src/Validator/UniqueUrlValidator.php
```diff
@@ -53,7 +53,12 @@ public function validate(mixed $value, Constraint $constraint): void
             return;
         }
 
-        $this->manager->fixUrl($value);
+        /**
+         * TODO: check if it is really necessary.
+         *  Because it is a validator, it shouldn't change the object.
+         *  and as it isn't persisting the changes on DB, it is changing the object before pass on the lines bellow.
+         */
+        // $this->manager->fixUrl($value);
```
eerison: it was added along time ago

https://github.com/sonata-project/SonataPageBundle/commit/50dcc77ff0697cc4f5d81d2ec3e5597c339c981b#diff-a2851bfaa1a6b101fb0fd268d31da8db71b64a9765532be31d255da7c53a796bR50

eerison:
hey @VincentLanglet 

do you know why [doctrine.orm.entity_listener](https://symfony.com/doc/6.4/doctrine/events.html#doctrine-entity-listeners) does not work 😄 ?

Edit: I used Entity/BasePage (it worked)

src/Listener/FixPageUrlLister.php
```diff
@@ -0,0 +1,29 @@
...
+use Sonata\PageBundle\Model\PageInterface;
+use Sonata\PageBundle\Service\Contract\FixPageUrlInterface;
+
+final class FixPageUrlLister
+{
+    public function __construct(private FixPageUrlInterface $fixPageUrl)
+    {
+    }
+
+    public function __invoke(PageInterface $page): void
```
eerison: @VincentLanglet is it ok. create a listener?

I remember that we talked about listener in the past.

if it is ok. then I will keep the url changes in this place.

as I can see some places uses the `save` method in [PageManager](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/PageManager.php#L123), and other no.

VincentLanglet (commented):
I feel like there is to many changes in a single PR.

If the purpose is to replace suglify by symfony string, why not JUST:
- changing SlugifyInterface into `SlugifyInterface|SluggerInterface`
- Updating the code to work with both
- Deprecating SlugifyInterface

?

Currently you are
- Introducing a new interface
- Introducing a new class
- Allowing `SlugifyInterface|FixPageUrlInterface` in PageManager
- Adding a deprecation
- Introducing an Adapter
- Introducing a Factory
- Introducing a Listener
- And removing a `fixUrl` call in the validator

Be aware that sonata-project repository have almost no maintainer, so the smaller are the change, the better it is

eerison:
> I feel like there is to many changes in a single PR.
> 
> If the purpose is to replace suglify by symfony string, why not JUST:
> 
> * changing SlugifyInterface into `SlugifyInterface|SluggerInterface`
> * Updating the code to work with both
> * Deprecating SlugifyInterface
> 
> ?
> 
> Currently you are
> 
> * Introducing a new interface
> * Introducing a new class
> * Allowing `SlugifyInterface|FixPageUrlInterface` in PageManager
> * Adding a deprecation
> * Introducing an Adapter
> * Introducing a Factory
> * Introducing a Listener
> * And removing a `fixUrl` call in the validator
> 
> Be aware that sonata-project repository have almost no maintainer, so the smaller are the change, the better it is

Yeah I know it is a bit big , I was trying to avoid  it.

> `SlugifyInterface|SluggerInterface`

It was my first thought, but I wanted to move this logical for a separated layer.  But extracting the logical lead for a PR with too many things.

I was trying to avoid change the arguments, and later change again.  But I will see how to make changes small. Maybe I back with the first solution accepting both Slug interface

eerison:
@VincentLanglet  could you take a look.

I just extracted the `fixUrl` logic from `PageManager` to `FixPageUrlService`, and yes I keeped 2 classes in BC layer.

src/Entity/PageManager.php
```diff
@@ -71,50 +82,24 @@ public function createWithDefaults(array $defaults = []): PageInterface
...
+        $fixPageUrl = $this->fixPageUrl;
+
+        if ($fixPageUrl instanceof SlugifyInterface) {
+            @trigger_error(\sprintf(
+                'Inject %s in %s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
+                SlugifyInterface::class,
+                self::class,
+                FixPageUrlInterface::class,
+            ), \E_USER_DEPRECATED);
+            $fixPageUrl = FixPageUrlSlugifyFactory::create($fixPageUrl);
```
eerison: Maybe this factory could be removed, since it is used only here and it won't be reused in the code. it will be like this

```php
$fixPageUrl = new FixPageUrlService(new SluggerSlugifyAdapter($slugify));
```

src/Entity/PageManager.php
```diff
@@ -35,10 +37,19 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|FixPageUrlInterface $fixPageUrl,
```
VincentLanglet: why not changing this signature to
```
private SlugifyInterface|SluggerInterface
```
?

Either it make sens to introduce a FixPageUrl service and this should be done in a separate PR in which the name will be challenged, either it is not needed and we could simplify this pr

VincentLanglet: Currently,
- I see no reason to introduce a FixPageUrlInterface because we don't plan having multiple implementation
- PageManager::fixUrl is basically useless since you can call FixPageUrl::fix directly (Which is not a bad thing but then the fixUrl method could be deprecated and the service injection)

src/Entity/PageManager.php
VincentLanglet: Since you're already planning to pass a Symfony slugger to this service, we could already have a small change of behavior here, so I don't think we have to force the use of sluggify here, you could 

(new FixPageUrlService(new AsciiSlugger()))->fix()

eerison: Well it will make the things easier, I do not need to keep support for slugify.

But don't you think we should keep an option to use Slugify in case someone find an issue and want to keep the old way?

or if someone want to use the old version, they should implement an `SlugifyAdapter` in their side?

src/Entity/PageManager.php
eerison: > private SlugifyInterface|SluggerInterface

if you see in my first [try](
https://github.com/sonata-project/SonataPageBundle/pull/1718/files#diff-d6d793e5c496d019efee48b4802b5314fe3474ac6934ac1ae4962b80fab77c61R163), I added some `if statement` to check which interface I am getting, and It makes the code more complex.

edit: those managers are like repositories: https://github.com/sonata-project/sonata-doctrine-extensions/blob/2.x/src/Entity/BaseEntityManager.php. and this logical should be in other layer.

> Either it make sens to introduce a FixPageUrl service and this should be done in a separate PR in which the name will be challenged, either it is not needed and we could simplify this pr

But don't you think this PR is simple enough to review? I basically extracted the logical and I am injecting it. 

> I see no reason to introduce a FixPageUrlInterface because we don't plan having multiple implementation

we do not plan to have multi implementation in our code. But it will help the devs to override the code if they need.

> PageManager::fixUrl is basically useless since you can call FixPageUrl::fix directly (Which is not a bad thing but then the fixUrl method could be deprecated and the service injection)

yep that is true. but I am planning to deprecate this in other PR.

src/Entity/PageManager.php
VincentLanglet: In next major we will remove sluggify anyway so if someone encounter an issue with the change it's better to fix it instead of keeping slugify support

eerison: Changed!

src/Service/FixPageUrlService.php
```diff
@@ -0,0 +1,74 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\PageBundle\Service;
+
+use Sonata\PageBundle\Model\PageInterface;
+use Sonata\PageBundle\Service\Contract\FixPageUrlInterface;
+use Symfony\Component\String\Slugger\SluggerInterface;
+
+final class FixPageUrlService implements FixPageUrlInterface
```
VincentLanglet: I think we should go with
```
final class PageFixer impelments PageFixerInterface
{}
```
and a method `PageFixerInterface::fixUrl`

eerison: as it is a generic name, probably it will contain more methods with more fix. then you prefer have everything centralized in one place, even if it increases for 3 or 5 methods. and if we see it is necessary we refactor.

I do not have problem to change to `PageFixer` I just want to get your idea :). it will help for future PRs 😄

eerison: other question:

without suffix `Service`? if I keep inside of Service it will be different of other classes that already exist.

edit: it is mixed, there are services inside of [src/Page/Service](https://github.com/sonata-project/SonataPageBundle/tree/4.x/src/Page/Service), [src/Bock](https://github.com/sonata-project/SonataPageBundle/tree/4.x/src/Block) and  `src/Service`

edit 2: I will keep with `Service` suffix as others services uses the same way, is it ok?

VincentLanglet: I personally hate the suffix service, didn't know some of the had.
On the opposite, other service like PageManager don't have such suffix.

We could create a folder Fixer ? with:
```
src/Fixer/PageFixerInterface
src/Fixer/PageFixer
```
WDYT ?

eerison: in src/ level we should keep things like

- Adapter
- Bridge
- Serializer
- Http
- Runtime
- Service
- Listener

etc I guess you know what I mean 😄  .. Fixer does not sound good IMO in `src/` level.

I guess inside of Services fits good, But wihout Suffix, maybe we need to decide to keep or not Suffix.

well the problem in not using suffix, is it is in PSR for others things like `MyCodeInterface`.

eerison: Maybe we could keep `PageFixerService` like everywhere, and discuss this in other issue?

eerison: @VincentLanglet 
Or do you think we should decide this before merge?

src/Entity/PageManager.php
```diff
@@ -71,50 +83,24 @@ public function createWithDefaults(array $defaults = []): PageInterface
...
-
-        // hybrid page cannot be altered
-        if (!$page->isHybrid()) {
-            $parent = $page->getParent();
-
-            if (null !== $parent) {
-                $slug = $page->getSlug();
-
-                if (null === $slug) {
-                    $slug = $this->slugify->slugify($page->getName() ?? '');
```
dmaicher, line 90: instead of introducing `PageFixerInterface` and `PageFixerService` I would suggest to just keep this code here and adjust it so it works with either `SlugifyInterface` or Symfony's `SluggerInterface` :thinking:

src/Entity/PageManager.php
```diff
@@ -71,50 +83,24 @@ public function createWithDefaults(array $defaults = []): PageInterface
...
+        $pageFixer = $this->pageFixer;
+
+        if ($pageFixer instanceof SlugifyInterface) {
+            @trigger_error(\sprintf(
+                'Inject %s in %s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
+                SlugifyInterface::class,
+                self::class,
+                PageFixerInterface::class,
+            ), \E_USER_DEPRECATED);
+            $pageFixer = new PageFixerService(new AsciiSlugger());
```
dmaicher, line 100: but this means the `SlugifyInterface` that is passed to the class is not used at all :thinking: seems like a breaking change to me. What if someone customized the slugging behavior?

src/Entity/PageManager.php
eerison: Hi @dmaicher thank you to take a time to review  :)

I was doing in other pr (#1718),  but ...

- The code had too many conditions to check which interface I'm using.
- I also saw, that this should be in other place beside "manager", managers are like repositories. And it shouldn't have any method to fix something.
- I can keep basically the same code with really small adjustment. 
- Extracting this code to other layer will help me with other PR https://github.com/sonata-project/SonataPageBundle/pull/1828

src/Entity/PageManager.php
eerison: Yep I think so too 👀

src/Entity/PageManager.php
dmaicher: yes I would be more comfortable just adjusting the code without extracting it. I agree there is room for improvement on extracting this out into its own class. But thats not really related to replacing `SlugifyInterface` imho.

Also see my other comment regarding the BC break for injecting a custom implementation of `SlugifyInterface`. This case needs to keep working.

eerison: But it's possible to make SlugifyInterface works. 

I created an Adapter to make PageFixService work with both. But It was suggested to just pass Symfony SluggerInterface and ignore SlugifyInterface. 

If I pass `SlugifyInterface | SluggerInterface ` It will keep the code into repository(manager). And it will stay there.

In case I want to refactor/extract this code,  I neeed to accept something like. 

`SlugifyInterface |SluggerInterface |PageFixInterface `, it will be more complex to refactor.

And IMO move this code to PageFixer isn't a big change at all.

dmaicher: > I created an Adapter to make PageFixService work with both. But It was suggested to just pass Symfony SluggerInterface and ignore SlugifyInterface.

that is additional code and complexity where I personally don't see the value for now. As @VincentLanglet also mentioned there aren't many maintainers left so your chances are higher to get this merged if you keep changes to a minimum.

src/Entity/PageManager.php
VincentLanglet: The strategy should be
```
if ($pageFixer instanceof PageFixerInterface) {
     $pageFixer->fixUrl($page);
} else {
     // old implementation
}
```

src/Entity/PageManager.php
eerison: OK, I'll back to the old  PR. 👍

eerison:
Closed in favor #1718

Issue #616, opened 2016-02-18
Unique Url validation not called


Th3Mouk:
I'm investigating on the Url generation what is wrong.

It's necessary to generate 2 snapshots of the page after changing the slug, to get the new url.

It's due to function fixUrl() on `PageManager` that she's only called during the snaping phase, and not during the edit of the page.

fixUrl() is called during the validation of the page edit, in the `Constraints\UniqueUrl` who is a class constraint, but seems not called.

Anyone has a clue how to proceed effectively ?

Th3Mouk:
ping @OskarStark @greg0ire @core23

greg0ire:
Sorry, I don't know that much about the page bundle. Constraints are called during validation, so I think you should have a look at how and whether validation is done when editing the a page.

OskarStark:
to be honest, i don't know something about that.

@core23 you?

core23:
Tested against the latest bundle versions... Everythink looks okay to me. I generated one new snapshot: the old link isn't accessible anymore the new link exist.

Th3Mouk:
@core23 the problem was on 2.3, have a look on the creation of a page. When you're doing the first snapshot the slug stay empty.
I'm no longer supporting 2.3, so i can probably close it.

Issue #60, opened 2012-02-27
page_defaults parameter not working


pulse00:
i'm trying to disable decoration for some pages whose routes are defined in my bundle. however, when i run the `update-core-routes` and `create-snapshot` commands, the `decorate` field is set to 1 in the database for the pages/snapshots.

I've searched through the code where the `page_defaults` configuration parameter is used to turn of decoration for those pages, but i cant't find this anywhere.

am i missing something here?

jorenvh1:
Can anyone tell me how to disable decoration if not from the config? And also what this exactly does?

Issue #50, opened 2012-02-15
Cannot add a new page, does not generate the url


janoist1:
If I try to create a new page the following error occures: The URL '/n-a' is already associated with another page
How to fix the url generation?

houssem86:
I have the same probleme, I want to create a new page but it doesn't generate URL (It didn't create new URL in table phpcr_nodes). Also, I want that the new page will be composed from blocks which I can edit them without associating a template

finchen:
same issue. on the demo site the page is created but has no url.
on my latest update, the page is not create:
`/ url already exists`

i have temporary fixed it by overriding PageAdmin
```
public function getNewInstance()
    {
        $instance = parent::getNewInstance();
        if ($instance->getParent() === null) {
            try {
                $parent = $this->pageManager->getPageByUrl($instance->getSite(), '/');
            } catch (PageNotFoundException $e) {
                throw new InternalErrorException('Unable to find the root url, please create a route with url = /');
            }
            $instance->setParent($parent ?: null);
        }

        return $instance;
    }
```

houssem86:
Hi finchen,
I tried your solution but I had this error message : 
"Catchable Fatal Error: Argument 1 passed to Sonata\PageBundle\Entity\PageManager::getPageByUrl() must implement interface Sonata\PageBundle\Model\SiteInterface, null given"

houssem86:
I fixed the error but this overriding like this : 
"    
public function getNewInstance() {
        $instance = parent::getNewInstance();

        if ($site = $this->getSite()) {
            $instance->setSite($site);
        }

        if ($instance->getParent() === null) {
            try {
                $parent = $this->pageManager->getPageByUrl($instance->getSite(), '/');
            } catch (PageNotFoundException $e) {
                throw new InternalErrorException('Unable to find the root url, please create a route with url = /');
            }
            $instance->setParent($parent ?: null);
        }

        return $instance;
    } 
"
But this did not fix my problem and the URL is not generated when I create a new page !

houssem86:
I fixed the message error by overriding PageAdmin like this : 
    public function getNewInstance()
    {
        $instance = parent::getNewInstance();

        if ($site = $this->getSite()) {
            $instance->setSite($site);
        }

        if ($instance->getParent() === null) {
            try {
                $parent = $this->pageManager->getPageByUrl($instance->getSite(), '/');
            } catch (PageNotFoundException $e) {
[...]

PR #1828, opened 2025-10-25
Remove fixUrl from UniqueUrlValidator

eerison:
## Remove fixUrl call from `UniqueUrlValidator`

I want to avoid inject a dependency that maybe it isn't necessary on `UniqueUrlValidator` on PR #1827.

Probably it was related with those issues: #50 #616.

and as it is a validator, IMO it shouldn't change any thing, just validate if it is valid or invalid. in case it need to be fix, it should be in other place.

I will try to reproduce those errors related on issues (#60, #616).

the goal is remove `fixUrl` from `UniqueUrlValidator`.

## Changelog

```markdown
### Removed
 - Removed `fixUrl()` call from `UniqueUrlValidator`
```

## To do

- [x] Update the tests;
- [x] Manual tests
  - [x] Creating page with a page that already exist
  - [x] Update page with the same name other page
  - [x] Update page with `/page-already-exist` and `page-already-exist`
  - [x] generating snapshot (But the Validator is not called here)

### DIFF
--- a/src/Validator/UniqueUrlValidator.php
+++ b/src/Validator/UniqueUrlValidator.php
@@ -53,8 +53,6 @@ final class UniqueUrlValidator extends ConstraintValidator
             return;
         }
 
-        $this->manager->fixUrl($value);
-
         $similarPages = $this->manager->findBy([
             'site' => $value->getSite(),
             'url' => $value->getUrl(),
--- a/src/Validator/UniqueUrlValidator.php
+++ b/src/Validator/UniqueUrlValidator.php
@@ -53,6 +53,8 @@ final class UniqueUrlValidator extends ConstraintValidator
             return;
         }
 
+        // $this->manager->fixUrl($value);
+
         $similarPages = $this->manager->findBy([
             'site' => $value->getSite(),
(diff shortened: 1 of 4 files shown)


eerison:
hey @VincentLanglet 

do you know how can I force composer install a specific symfony version?

I tried `SYMFONY_REQUIRE=7.1. VERSION=7.1. composer install`, But it didn't work :/

Edit: never mind I can reproduce the test :)

VincentLanglet (commented):
While I agree this call has no reason to be here, removing it might be consider as a BC break for people which was voluntary or involuntary relying on it...
We might call this a bugfix, but it's kinda opinionated...

I would be interested about the behavior changes of the removal of this line (Would be nice to have a test covering this validator and changing value when this line is removed...). This way we would understand better the BC break.

Another option would be to introduce an option to call or not this method ?

eerison:
I didn't thought it could be a BC, but yeah it will change the behavior.

> Another option would be to introduce an option to call or not this method ?

it would be nice, I just do not have any idea 😄 .

PR #1833, opened 2025-11-14
Feature/1713 - Switch slugify for symfony string

### LINKED ISSUE #1713: Switch cocur/slugify to symfony/string

eerison:
## Replace  slugify to symfony string

Closes #1713.

## Changelog

```markdown
### Added
- Handle slugs using `SluggerInterface::class` from `symfony/string`

### Deprecated
- Deprecated passing `SlugifyInterface::class` to `PageManager::__construct`.
- Deprecated `sonata.page.slugify.cocur` service.
```

### DIFF
--- a/src/Entity/PageManager.php
+++ b/src/Entity/PageManager.php
@@ -19,6 +19,7 @@ use Sonata\Doctrine\Entity\BaseEntityManager;
 use Sonata\PageBundle\Model\PageInterface;
 use Sonata\PageBundle\Model\PageManagerInterface;
 use Sonata\PageBundle\Model\SiteInterface;
+use Symfony\Component\String\Slugger\SluggerInterface;
 
 /**
  * @extends BaseEntityManager<PageInterface>
@@ -35,10 +36,21 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|SluggerInterface $slugger, // NEXT_MAJOR: Reduce $slugger to SluggerInterface
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
+        if ($slugger instanceof SlugifyInterface) {
+            trigger_deprecation(
+                'sonata-project/page-bundle',
+                '4.11.0',
+                'Passing an instance of %s to %s is deprecated. Use an instance of %s instead.',
+                SlugifyInterface::class,
+                __METHOD__,
+                SluggerInterface::class
+            );
+        }
+
         parent::__construct($class, $registry);
     }
 
@@ -87,7 +99,7 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
                 $slug = $page->getSlug();
 
                 if (null === $slug) {
-                    $slug = $this->slugify->slugify($page->getName() ?? '');
+                    $slug = $this->fixUrlWithSluggerOrSlugify($page->getName() ?? '');
 
                     $page->setSlug($slug);
                 }
@@ -165,5 +177,21 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
             ->setParameter('site', $site->getId())
             ->getQuery()
             ->execute();
+    }
+
+    /**
+     * NEXT_MAJOR: Remove this method and move $this->slugger->slug()... to fixUrl method.
+     */
+    private function fixUrlWithSluggerOrSlugify(string $value): string
+    {
+        if ($this->slugger instanceof SlugifyInterface) {
+            return $this->slugger->slugify($value);
+        }
+
+        // NEXT_MAJOR: Move this to fixUrl method.
+        return $this->slugger
+                    ->slug($value)
+                    ->lower()
+                    ->toString();
     }
 }
--- a/src/Resources/config/orm.php
+++ b/src/Resources/config/orm.php
@@ -27,6 +27,7 @@ use Sonata\PageBundle\Model\SnapshotPageProxy;
 use Sonata\PageBundle\Model\SnapshotPageProxyFactory;
 use Sonata\PageBundle\Serializer\BlockTypeExtractor;
 use Sonata\PageBundle\Serializer\InterfaceTypeExtractor;
+use Symfony\Component\String\Slugger\SluggerInterface;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
     $containerConfigurator->services()
@@ -42,7 +43,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             ->args([
                 param('sonata.page.page.class'),
                 service('doctrine'),
-                service('sonata.page.slugify.cocur'),
+                service(SluggerInterface::class),
                 abstract_arg('defaults'),
                 abstract_arg('page defaults'),
             ])
--- a/src/Resources/config/slugify.php
+++ b/src/Resources/config/slugify.php
@@ -15,9 +15,15 @@ namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Cocur\Slugify\Slugify;
 
+// NEXT_MAJOR: Remove this file.
 return static function (ContainerConfigurator $containerConfigurator): void {
     $containerConfigurator->services()
 
         ->set('sonata.page.slugify.cocur', Slugify::class)
+            ->deprecate(
+                'sonata-project/page-bundle',
+                '4.11.0',
+                'Service "%service_id%" is deprecated.',
+            )
             ->public();
 };
--- a/tests/Entity/PageManagerTest.php
+++ b/tests/Entity/PageManagerTest.php
@@ -14,19 +14,37 @@ declare(strict_types=1);
 namespace Sonata\PageBundle\Tests\Entity;
 
 use Cocur\Slugify\Slugify;
+use Cocur\Slugify\SlugifyInterface;
 use Doctrine\Persistence\ManagerRegistry;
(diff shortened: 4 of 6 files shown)


eerison:
@dmaicher @VincentLanglet could you take a look.

src/Entity/PageManager.php
```diff
@@ -87,7 +96,7 @@ public function fixUrl(PageInterface $page): void
                 $slug = $page->getSlug();
 
                 if (null === $slug) {
-                    $slug = $this->slugify->slugify($page->getName() ?? '');
+                    $slug = $this->fixUrlWithSluggerOrSlugify($page->getName());
```
dmaicher:
```suggestion
                    $slug = $this->fixUrlWithSluggerOrSlugify($page->getName() ?? '');
```

src/Entity/PageManager.php
```diff
@@ -166,4 +175,20 @@ public function getHybridPages(SiteInterface $site): array
             ->getQuery()
             ->execute();
     }
+
+    /**
+     * NEXT_MAJOR: Remove this method and move $this->slugger->slug()... to fixUrl method.
+     */
+    private function fixUrlWithSluggerOrSlugify(?string $value): string
```
dmaicher:
```suggestion
    private function fixUrlWithSluggerOrSlugify(string $value): string
```

src/Entity/PageManager.php
```diff
@@ -166,4 +175,20 @@ public function getHybridPages(SiteInterface $site): array
...
+    {
+        if ($this->slugger instanceof SlugifyInterface) {
+            return $this->slugger->slugify($value ?? '');
```
dmaicher:
```suggestion
            return $this->slugger->slugify($value);
```

src/Entity/PageManager.php
```diff
@@ -166,4 +175,20 @@ public function getHybridPages(SiteInterface $site): array
...
+        }
+
+        // NEXT_MAJOR: Move this to fixUrl method.
+        return $this->slugger
+                    ->slug($value ?? '')
```
dmaicher:
```suggestion
                    ->slug($value)
```

src/Entity/PageManager.php
```diff
@@ -35,10 +36,18 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
...
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|SluggerInterface $slugger, // NEXT_MAJOR: Reduce $slugger to SluggerInterface
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
+        if ($slugger instanceof SlugifyInterface) {
+            @trigger_error(\sprintf(
+                '%s is deprecated since version 4.10.0 and will be removed in 5.0, use %s instead of.',
```
dmaicher:
```suggestion
                '%s is deprecated since version 4.11.0 and will be removed in 5.0, use %s instead of.',
```

let's use the next minor here which is 4.11

dmaicher (commented):
:+1:

src/Entity/PageManager.php
```diff
@@ -35,10 +36,18 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
...
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|SluggerInterface $slugger, // NEXT_MAJOR: Reduce $slugger to SluggerInterface
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
+        if ($slugger instanceof SlugifyInterface) {
+            @trigger_error(\sprintf(
+                '%s is deprecated since version 4.11.0 and will be removed in 5.0, use %s instead of.',
```
dmaicher: Maybe we can be more specific? Like 

```
Passing an instance of ... to ... is deprecated since ... . Use an instance of ... instead.

File: tests/Entity/PageManagerTest.php [hunk 7]
```diff
@@ -14,19 +14,35 @@
...
+     * @return iterable<array<SlugifyInterface|SluggerInterface>>
+     **/
+    public static function providerSlugs(): iterable
+    {
+        yield [new Slugify()];
+        yield [new AsciiSlugger()];
+    }
+
+    #[DataProvider('providerSlugs')]
+    public function testFixUrl(SlugifyInterface|SluggerInterface $slug): void
```
dmaicher, line 42: do those tests now trigger deprecations? I guess they should. Maybe we need to ignore/suppress those?

dmaicher (commented):
2 more small comments from my side

tests/Entity/PageManagerTest.php
dmaicher: Those I mean 
image

src/Entity/PageManager.php
dmaicher: we could also add and use `symfony/deprecation-contracts` as it simplifies this slightly

tests/Entity/PageManagerTest.php
eerison: I guess we need to use [group legacy](https://github.com/sonata-project/SonataPageBundle/blob/4.x/CONTRIBUTING.md?plain=1#L415), I used it in the past, But I'm not sure if it still run on pipelines.

it should be good we keep those tests running, it will guarantee that code keep working with both implementations.

dmaicher: I think we can use PHPUnit's `#[IgnoreDeprecations]`? :thinking: 

Edit: I think the legacy group was used back then by symfony/phpunit-bridge to suppress deprecations. Now with native PHPUnit its better to use `#[IgnoreDeprecations]`

eerison: ahhh maybe it is the reason we are using it here : https://github.com/sonata-project/SonataPageBundle/blob/4.x/tests/Runtime/SonataPagePathRuntimeTest.php#L27, I got it!

src/Entity/PageManager.php
```diff
@@ -35,10 +36,21 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
...
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
+        if ($slugger instanceof SlugifyInterface) {
+            trigger_deprecation(
+                'sonata-project/page-bundle',
+                '4.11.0',
+                'Passing an instance of %s to %s is deprecated. Use an instance of %s instead.',
+                SlugifyInterface::class,
+                __CLASS__,
```
dmaicher:
```suggestion
                __METHOD__,
```

src/Resources/config/orm.php
```diff
@@ -42,7 +43,7 @@
             ->args([
                 param('sonata.page.page.class'),
                 service('doctrine'),
-                service('sonata.page.slugify.cocur'),
```
dmaicher, line 45: let's also mark this service as deprecated?

src/Entity/PageManager.php
```diff
@@ -35,10 +36,21 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
     public function __construct(
         string $class,
         ManagerRegistry $registry,
-        private SlugifyInterface $slugify,
+        private SlugifyInterface|SluggerInterface $slugger, // NEXT_MAJOR: Reduce $slugger to SluggerInterface
         private array $defaults = [],
         private array $pageDefaults = [],
     ) {
+        if ($slugger instanceof SlugifyInterface) {
+            trigger_deprecation(
```
dmaicher, line 44: let's also add a requirement for `symfony/deprecation-contracts` then to composer.json.

src/Entity/PageManager.php
eerison:
```diff
+ Passing an instance of ... to ... is deprecated.
- since ... . 
+Use an instance of ... instead.
```

I keep the message until `deprecated` because `symfony deprecation` add this into the message.

src/Entity/PageManager.php
```diff
@@ -35,10 +36,21 @@ final class PageManager extends BaseEntityManager implements PageManagerInterfac
...
+                'sonata-project/page-bundle',
+                '4.11.0',
+                'Passing an instance of %s to %s is deprecated. Use an instance of %s instead.',
+                SlugifyInterface::class,
+                self::class,
```
dmaicher: would suggest using `__METHOD__`

src/Entity/PageManager.php
eerison: heheheh true 😄

src/Resources/config/slugify.php
```diff
@@ -14,10 +14,19 @@
...
 
 use Cocur\Slugify\Slugify;
+use Symfony\Component\String\Slugger\SluggerInterface;
 
+// NEXT_MAJOR: Remove this file.
 return static function (ContainerConfigurator $containerConfigurator): void {
     $containerConfigurator->services()
 
         ->set('sonata.page.slugify.cocur', Slugify::class)
+            ->deprecate(
```
eerison, line 23: @dmaicher

do you mean like this?

src/Resources/config/orm.php
eerison: Deprecated!

src/Resources/config/slugify.php
dmaicher: Yes exactly but I guess need to use sprintf there for the 3rd argument

eerison: Yep, I just saw it 😄

dmaicher (approved):

VincentLanglet (approved):

dmaicher:
Thanks @eerison

Issue #1837, opened 2025-11-23
Add Symfony 8 compatibility


dmaicher:
We want to add Symfony 8 compatibility and also

- drop PHP 8.1
- drop Symfony 7.1 and 7.2
- remove Psalm

For the 4.x branch there is a PR with the automated updates for some files: https://github.com/sonata-project/SonataPageBundle/pull/1835

dmaicher:
Blocked by https://github.com/sonata-project/SonataBlockBundle/pull/1245 and https://github.com/sonata-project/SonataAdminBundle/pull/8357

eerison:
I would say it is a big block for page bundle : https://github.com/symfony-cmf/routing-bundle/blob/3.x/composer.json

I don't see any movement to support symfony 8.
I don't know how much coupled is page bundle with this dependency.

Edit: just a quickly look, I guess it isn't used in many places: https://github.com/search?q=repo%3Asonata-project%2FSonataPageBundle%20cmf&type=code

dmaicher:
I don't use SonataPageBundle myself and are not familiar enough with it or `symfony-cmf/routing-bundle` to discuss if we need it or not 🤷‍♂️ 

But a first start could be to look into adding Symfony 8 compatibility to https://github.com/symfony-cmf/routing-bundle with a PR?

eerison:
> I don't use SonataPageBundle myself and are not familiar enough with it or `symfony-cmf/routing-bundle` to discuss if we need it or not 🤷‍♂️
> 
> But a first start could be to look into adding Symfony 8 compatibility to https://github.com/symfony-cmf/routing-bundle with a PR?

Yeah, for symfony 8 we must have symfony-cmf updated. Even if we replace for some symfony alternative, we can't remove this dependency in 4.x branch.

dmaicher:
This commit should be merged as well with the Symfony 8 update then: https://github.com/sonata-project/SonataPageBundle/pull/1840

eerison:
Hey @dmaicher , do you have any email or platform that I could ping you?

PR #1838, opened 2025-11-25
Small Symfony 7.3 deprecations

eerison:
## Fix small Symfony 7.3 deprecation

I am targeting this branch, because we will make 4.x version compatible with symfony 8

Issue #1837

Note:  I can't test symfony 8, because most of the blocks related with this bundle is also blocking admin bundle, then I believe it is possible to fully test this bundle after admin bundle.

the first block was related with doctrine-bundle, But as I can see, Vicente is aware of this https://github.com/doctrine/DoctrineBundle/pull/2113#issuecomment-3508818346


tests/App/config/config.yaml
```diff
@@ -14,6 +14,8 @@ framework:
     http_method_override: false
     session:
         storage_factory_id: session.storage.factory.mock_file
+    property_info:
```
eerison: it is used on test only.
then I am use the default value of symfony 8

eerison: damn it was added on 7.3 :/

eerison:
@dmaicher

dmaicher (approved):
Thanks :+1: I removed the changelog since this change does not affect users of the bundle

dmaicher:
Symfony 8 support is still blocked by some external packages like for example https://github.com/KnpLabs/KnpMenuBundle/pull/500 :confused:

PR #1839, opened 2025-11-26
POC Use MapQueryString and Remove BCHelper

eerison:
I want to use this PR just to see if we could use [MapQueryString](https://symfony.com/blog/new-in-symfony-6-3-query-parameters-mapper) and remove [BCHelper](https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/BCLayer/BCHelper.php)

Will it work everywhere? I don't know yet. But I guess we could use this we it is possible.

yeah it will add a lot of BC because I need to replace the `Request` to `SomeQuery`

I added it inside of  `Query/` But on symfony docs recomends `Dto`, But I also saw in some places as `Input`.
Then I would like to know which one you prefer.

I am fine with Dto as well, then it will be into `Dto/BlockQueryDto.php`.

@dmaicher @VincentLanglet

### DIFF
--- a/src/Controller/BlockAdminController.php
+++ b/src/Controller/BlockAdminController.php
@@ -14,14 +14,14 @@ declare(strict_types=1);
 namespace Sonata\PageBundle\Controller;
 
 use Sonata\AdminBundle\Controller\CRUDController;
-use Sonata\AdminBundle\Exception\BadRequestParamHttpException;
 use Sonata\BlockBundle\Block\BlockServiceManagerInterface;
 use Sonata\BlockBundle\Block\Service\EditableBlockService;
-use Sonata\PageBundle\BCLayer\BCHelper;
 use Sonata\PageBundle\Model\BlockInteractorInterface;
 use Sonata\PageBundle\Model\PageBlockInterface;
+use Sonata\PageBundle\Query\BlockQuery;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Attribute\MapQueryString;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Things is you cannot do this because
- We supports older Symfony version than the one with MapQueryString (I think)
- MapQueryString only works for query, but the BCHelper works for query + request + attributes, so this changes is an too big BC break.

Second point is important to understand.
EVEN IF Sonata only use query param (which is not always the case) if someone in his project pass the value in the request or as attributes, you will break his project with this change.

See https://github.com/sonata-project/SonataAdminBundle/issues/7365#issuecomment-894617526

eerison:
> * We supports older Symfony version than the one with MapQueryString (I think)

The minimum version is 6.4, and it was added on 6.3, then I guess we are fine with this.

> MapQueryString only works for query, but the BCHelper works for query + request + attributes, so this changes is an too big BC break.

well there are
 - MapQueryString
 - MapRequestPayload
 
But for attributes I dot know yet, how we could handle it.

> but the BCHelper works for query + request + attributes, so this changes is an too big BC break.
Yeah I was thinking of it, I guess it will be a Big BC.

But do you think worth the effort to evaluate this approach? or do you think it is a complete NO GO?

> https://github.com/sonata-project/SonataAdminBundle/issues/7365#issuecomment-894617526

Good to know

VincentLanglet:
> well there are
> MapQueryString
> MapRequestPayload

But they dont support both query and request at the same type.

We have route which works for both.

eerison:
> > well there are
> > MapQueryString
> > MapRequestPayload
> 
> But they dont support both query and request at the same type.
> 
> We have route which works for both.

Yep I know, but in this case we need to replace for something like this


```diff
 public functon yourMethod(
-        Request $request
+       #[MapQueryString] YourQueryDto $query,
+       #[MapRequestPayload] YourPayloadDto $payload,
    )
```

then things that comme from query will be mapped to `$query` and things from `payload` to `$payload` (and probably other argument  to `Attributes`)

Edit: Do you have some endpoint in mind that uses all of this? `query`, `payload` and `attibutes` in the same place?

VincentLanglet:
No, you don't understand.

Our endpoint works with query AND request/attributes for the SAME value.

`/path?admin=foo`
`/path` with `{admin: foo}` in the body
`/path` with `[admin => foo]` in the attributes

eerison:
> No, you don't understand.
> 
> Our endpoint works with query AND request/attributes for the SAME value.
> 
> `/path?admin=foo` `/path` with `{admin: foo}` in the body `/path` with `[admin => foo]` in the attributes

